PR #1, opened 2019-09-26
Unlicense default config

gromakovsky:
## Description

Problem: default config is not really smart and seems easier to just
unlicense it, i. e. let everyone do whatever they want with it.
Also it does not have SPDX header (like any other file in this repo,
but that's another story).

Solution: add SPDX header stating that the Unlicense applies.
Make Serokell copyright holder of it, AFAIK @martoon is fine with
that (and I am not sure whether copyright holder matters for
unlicensed files at all).

## Related issue(s)

That's for a private project.

## :white_check_mark: Checklist for your Pull Request

Ideally a PR has all of the checkmarks set.

If something in this list is irrelevant to your PR, you should still set this
checkmark indicating that you are sure it is dealt with (be that by irrelevance).

#### Related changes (conditional)

- Tests
  - [x] If I added new functionality, I added tests covering it.
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [x] I checked whether I should update the docs and did so if necessary:
    - [README](/README.md)
    - Haddock

- Public contracts
  - [x] Any modifications of public contracts comply with the [Evolution
  of Public Contracts](https://www.notion.so/serokell/Evolution-of-Public-Contracts-2a3bf7971abe4806a24f63c84e7076c5) policy.
  - [x] provided a migration guide for breaking changes if possible

#### Stylistic guide (mandatory)

[//]: # (Update link to style guide if necesary or remove if it's not present)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [x] My code complies with the [style guide](docs/code-style.md).

PR #2, opened 2019-11-26
[INT-130] Use newer resolver

gromakovsky:
## Description

Problem: latest available LTS resolver is 14.16 and `crossref-verifier`
doesn't compile with it.

Solution: update resolver to 14.16. Newer versions of some packages
are used from now on. I assume that it shouldn't break anything as
long as tests pass.
`req` stopped using `Default` class, that's the only change in the code
I had to make.

## Related issue(s)

https://issues.serokell.io/issue/INT-130

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts


gromakovsky:
Okay, will take a closer look, created https://issues.serokell.io/issue/INT-130 for this purpose.

Also after creating this issue I realize that we should have one meta issue for `crossref-verifier` and mostly use GitHub for it because it's a public project for everyone.

gromakovsky:
@Martoon-00 I have fixed that (it was not related to this PR, rather a bug in #1). Now waiting for @balsoft to fix CI.

balsoft:
CI fixing is happening in balsoft/haskell.nix branch. I hope to finish it tomorrow, but it's not trivial (as I'm switching to a completely different framework and also this framework contains bugs that need to be fixed). So even if I finish tomorrow, we'll have to wait for upstream to accept the changes.

gromakovsky:
No problem, it can wait. Hopefully we'll have to wait no more than a couple of weeks.

Martoon-00:
Good, generated executable seems working now.

balsoft:
So okay, haskell.nix seems deeply flawed, so we either need to fix it which can take a long time, or revive https://github.com/serokell/nixage which can take a really long time, or continue to use stackage2nix which is generally recognized to be deprecated and bad.

gromakovsky:
Well, all I want to do is to make CI work with lts-14.16 or least some other 14.*. Is it hard to do that with `stackage2nix` (and treat it as a bad solution that should be changed afterwards)?

Also can't we just use `stack` like we do in for instance `tezos-btg`? https://github.com/serokell/tezos-btg/blob/master/.buildkite/pipeline.yml

gromakovsky:
In other words, for this issue let's just do whatever is easier even if it's "bad" but works.

yorickvP:
merging in https://github.com/serokell/crossref-verifier/tree/balsoft/haskell.nix should fix the CI failure for this

gromakovsky:
Now I get
```
fatal: unable to access 'https://github.com/serokell/lootbox.git/': Could not resolve host: github.com
--
```

and retrying doesn't help.

balsoft:
`--no-sandbox` will get you just a bit further

yorickvP:
@balsoft why does that error not happen on the branch itself?

balsoft:
@yorickvP because caching, and for some reason it's not caching on this branch.

stack.yaml
```diff
@@ -1,23 +1,15 @@
...
-- req-1.2.1
-- aeson-options-0.0.0
-- aeson-1.3.1.1
-- o-clock-1.0.0.1
+- aeson-options-0.1.0
+- o-clock-1.1.0
 
 - git: https://github.com/serokell/lootbox.git
-  commit: 34e389808e34f1cfa56f456773682325bed56d17
+  commit: 4700376b8493f6ac164461715cb72a0259148ac2 # master
```
balsoft: Probably because of this :)

balsoft:
I'll change this branch a bit so that fails later

gromakovsky:
But why can't it resolve host? When I get such error on my computer it usually means that I am not connected to the internet. But if jupiter was not connected to the internet, I think we wouldn't get anything from it :confused:

stack.yaml
gromakovsky: What shall we do if we update some revision in `stack.yaml`? How can we regenerate `sources.json`?

yorickvP:
Because the failure mode of not specifying the right package url/name/version in the `cache` argument to `stackProject` is it trying to get it from the internet, which it can't do inside a nix build

stack.yaml
balsoft: Hmm, you gave me a good idea actually

balsoft: No, not a good idea. There's no good way to combine `niv` with `nix-prefetch-git`

balsoft:
I don't understand why it fails now, will repro locally.

stack.yaml
balsoft: Hm, wait, actually a good idea. There is a way to combine them (of course, niv uses nix-prefetch-git)

balsoft:
Should be partially fixed. Also, now lootbox cache can be updated with `niv update lootbox`

balsoft:
Hmm, this now seems to just fail the tests because of a missing dep

stack.yaml
yorickvP: we could use the code from https://github.com/serokell/stack-to-nix to automatically derive the cache file from stack.yaml

yorickvP:
Shouldn't hspec-discover be a dependency specified in `package.yaml`?

balsoft:
Even if I add it as a dependency, the `hspec-discover` package in `haskell.nix` contains only the libraries, not the binaries. I'm not sure how to make it build with binaries.

balsoft:
Hooray! @yorickvP @gromakovsky this passes CI checks now.

result-1
```diff
@@ -0,0 +1 @@
+/nix/store/x4kvk5qc73l2dpn0zjzqdmn47y75f1hg-default-Setup
```
gromakovsky: What is this?

gromakovsky:
Rebased on master and force pushed.

gromakovsky:
>  I'm not sure how to make it build with binaries.

How did you solve that?

balsoft:
@gromakovsky nix-tools are smart enough to build the executable only when the package is in `build-tools` instead of `dependencies`, which makes perfect sense.

gromakovsky:
Ah, right, makes sense indeed.

result-1
gromakovsky: Ah, AFAIU it is a symlink that was committed by mistake. Correct?

balsoft: Yep, correct. We need to remove this and probably add a gitignore with `result*/`

balsoft:
@gromakovsky CI passes (again) now.

gromakovsky:
I don't want to merge it yet because I want to understand each file that I don't understand. I am reading about haskell.nix and niv now, probably will ask some questions.

gromakovsky:
Okay, I have figured everything out, thanks to @balsoft who explained some stuff to me (in DM :kot:).

PR #6, opened 2019-12-05
Improve readme

Martoon-00:
## Description

Make README more attractive to an arbitrary user.

[Rendered](https://github.com/serokell/crossref-verifier/tree/martoon/int128-nice-readme#crossref-verifier).

I adjusted it according to my own taste, which though is known to be strage enough to deserve Lebedev's award, so I would appreciate if someone else takes a look.  

I did everything what we discussed before except for committing release

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ## Reporting issues
 
-Please [open an issue](../../../issues/new/choose) if you find a bug or have a feature request.
+Please [open an issue](https://github.com/serokell/crossref-verifier/issues/new/choose) if you find a bug or have a feature request.
 Note: you need to login (e. g. using your GitHub account) first.
 Before submitting a bug report or feature request, check to make sure it hasn't already been submitted
 
--- a/README.md
+++ b/README.md
@@ -109,9 +109,8 @@ Currently supported options include:
 
 ## Issue tracker [↑](#crossref-verifier)
 
-We use [YouTrack](https://issues.serokell.io/issues/INT) as our issue
-tracker. You can login using your GitHub account to leave a comment or
-create a new issue.
+We use GitHub issues as our issue tracker.
+You can login using your GitHub account to leave a comment or create a new issue.
 
 ## For Contributors [↑](#crossref-verifier)
 
--- a/README.md
+++ b/README.md
@@ -27,6 +27,9 @@ Our CI uses `nix-build crossref-verifier.nix` to build the whole project includi
 It is based on the [`haskell.nix`](https://input-output-hk.github.io/haskell.nix/) project.
 You can do that too if you want.
 
+<details>
+  <summary>Details</summary>
+
 There is a [bug](https://github.com/input-output-hk/haskell.nix/issues/335) which causes us to put some redundancy into Nix files:
 1. [`nix/sources.json`](nix/sources.json) lists all such dependencies that we obtain using `git`.
 It specifies concrete git revisions and SHA256 checksums.
@@ -34,6 +37,8 @@ It specifies concrete git revisions and SHA256 checksums.
 
 As a consequence, you may have to update these files when you update [`stack.yaml`](stack.yaml).
 You can use [`niv update`](https://github.com/nmattia/niv#update) to update [`nix/sources.json`](nix/sources.json).
+
+</details>
 
 ## Usage [↑](#crossref-verifier)
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Crossref-verifier
 
+[![Build status](https://badge.buildkite.com/75461331a6058b334383cdfca1071dc1f908b70cf069d857b7.svg?branch=master)](https://buildkite.com/serokell/crossref-verifier)
+
 Tool for verifying local and external references in repository documentation.
 
 For now, support for markdown files only is provided.
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Please see [CONTRIBUTING.md](/.github/CONTRIBUTING.md) for more information.
 
 ## About Serokell [↑](#crossref-verifier)
 
-Crossref-verifier is maintained by [Serokell](https://serokell.io/).
+Crossref-verifier is maintained and funded with ❤️ by [Serokell](https://serokell.io/).
+The names and logo for Serokell are trademark of Serokell OÜ.
 
-We love open source software.
-See which [services](https://serokell.io/#services) we provide and [drop us a line](mailto:hi@serokell.io) if you are interested.
+We love open source software! See [our other projects](https://serokell.io/community?utm_source=github) or [hire us](https://serokell.io/hire-us?utm_source=github) to design, develop and grow your idea!
--- a/README.md
+++ b/README.md
@@ -6,11 +6,40 @@ Tool for verifying local and external references in repository documentation.
 
 <img src="https://user-images.githubusercontent.com/5394217/70820564-06b06e00-1dea-11ea-9680-27f661ca2a58.png" alt="Output sample" width="600"/>
 
+### Motivation
+
+Comparing to alternative solutions, this tool tries to achieve the following points.
+
+As project evolves, links in documentation have tention to get broken. This is usually induced by
+1. Files movements;
+2. Markdown headers renames;
+3. Outer sites ceasing their existence.
+
+This tool will help you to keep references in order.
+
 ### Aims
 
 Comparing to alternative solutions, this tool tries to achieve the following points.
 
-<img src="https://imgur.com/2w8FVIH.png" alt="Output sample" width="600"/>
+* Quickness - local references are verified instantly even for moderately-sized repositories.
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -1,8 +1,43 @@
...
+[![Build status](https://badge.buildkite.com/75461331a6058b334383cdfca1071dc1f908b70cf069d857b7.svg?branch=master)](https://buildkite.com/serokell/crossref-verifier)
 
-For now, support for markdown files only is provided.
+Tool for verifying local and external references in repository markdown documentation.
+
+<img src="https://imgur.com/2w8FVIH.png" alt="Output sample" width="600"/>
+
+### Motivation
+
+As project evolves, links in documentation has tention to get broken. This is usually induced by
```
gromakovsky:
```suggestion
As project evolves, links in documentation have tention to get broken. This is usually induced by
```

README.md
```diff
@@ -36,19 +82,36 @@ For description of other options
...
+- [ ] Support for non-Unix systems.
+- [ ] Support links detection in different languages, not only Markdown.
+  - [ ] Haskell haddock is first in turn.
+
 ## Issue tracker [↑](#crossref-verifier)
 
-We use [YouTrack](https://issues.serokell.io/issues/INT) as our issue
-tracker. You can login using your GitHub account to leave a comment or
-create a new issue.
+We use [GitHub issues](../../../issues) as our issues tracker.
```
gromakovsky: I think it's always called "issue tracker", not "issues tracker".

src-files/def-config.yaml
```diff
@@ -25,15 +25,24 @@ verification:
...
   notScanned:
-    # Github-specific files
+    # GitHub-specific files
     - .github/pull_request_template.md
     - docs/pull_request_template.md
 
-  # Glob patterns describing the files which do not physically exist in the repository
-  # but should be treated as existing nevertheless.
+    # GitLab-specific files
+    - .gitlab/merge_request_templates/
```
gromakovsky: Also there can be `.gitlab/issues_templates`. But I would check whether local references work fine there first. Also I suppose we don't want to check only local references in MR templates, but we want to check global references, right? How does it work currently?

src-files/def-config.yaml
```diff
@@ -25,15 +25,24 @@ verification:
...
+
+  # Glob patterns describing the files which do not physically exist in the
+  # repository but should be treated as existing nevertheless.
```
gromakovsky: How is it used? Is there an example?

README.md
```diff
@@ -6,6 +6,8 @@ Tool for verifying local and external references in repository documentation.
 
 For now, support for markdown files only is provided.
 
+<img src="https://imgur.com/2w8FVIH.png" alt="Output sample" width="600"/>
```
gromakovsky: Imo, it's more reliable to put the file with this picture into the repo, then it will work even if imgur dies.

gromakovsky:
Could you please also add issue templates? See how it's done in https://github.com/serokell/tezos-client-packaging/tree/master/.github

And "Open an issue" link in `CONTRIBUTING.md` should be https://github.com/serokell/crossref-verifier/issues/new/choose

README.md
Martoon-00: Right.

src-files/def-config.yaml
Martoon-00: Isn't it clear from the examples below?

README.md
Martoon-00: Hm, that picture weights 64k, and we may need to update it often which means that picture will take a considerable amount of space in the repository.
Maybe as compromise upload this picture to gist? Probably we can assume that GitHub working `<=>` gists working.

src-files/def-config.yaml
Martoon-00: Currently, it's only possible to ignore a file entirely. Which seems good in our case because templates often contain incomplete links like `https://issues.serokell.io/issue/INT-`.

> But I would check whether local references work fine there first
Do you know a repository where I could spoil `master` to check that? :peka:

Martoon-00:
> Could you please also add issue templates? See how it's done in https://github.com/serokell/tezos-client-packaging/tree/master/.github
> 
> And "Open an issue" link in `CONTRIBUTING.md` should be https://github.com/serokell/crossref-verifier/issues/new/choose

Resolved here: https://github.com/serokell/crossref-verifier/pull/7

src-files/def-config.yaml
gromakovsky: >  UPD: actually I can try in this repository and then just force-push to master.

Meh, it's better not to do it, because in some repos we use `master` of this repo in CI (not a specific commit). What about this one https://github.com/Martoon-00/Natural-science-X? (I've been using my old personal repo for testing previously.) You can force-push there, probably chances that it will be noticed are smaller than for `crossref-verifier`.

gromakovsky: Or maybe https://github.com/Martoon-00/HW-10, it has less contributors.

src-files/def-config.yaml
gromakovsky: I am talking exactly about those examples. Where do we use `../../../pulls` and such?

gromakovsky: AFAIU it's some GitHub magic/hacks that I am not familiar with (or don't realize that).

README.md
gromakovsky: Okay, let's do gist :man_shrugging:

src-files/def-config.yaml
gromakovsky: Hm, apparently you don't even have to force-push.

Martoon-00: That's true :joy: :joy: :joy: 

Okay, I've found out that if I want to refer to some repository file from issue, it should be of form `../blob/master/<absolute path to file>`. So apparently crossref-verifier cannot handle such for now.

src-files/def-config.yaml
Martoon-00: Ah hm, I was sure that we use such links in metatemplates, but we actually don't.

AFAIU it's the only way to refer to `issues` or `merge requests` page without mentioning the repository name. And Github/Gitlab requires such weird number of '..', no one knows why.

src-files/def-config.yaml
Martoon-00: > Hm, apparently you don't even have to force-push.

gromakovsky: Oh, wait.
> Also there can be .gitlab/issues_templates. But I would check whether local references work fine there first. 

Here I was taking about **GitLab**.

Martoon-00: Oh mda, right.

Well, somehow it happens that GitLab also required `../blob/master/README.md` link if I put it to issue, so I'm adding.

Martoon-00:
Screenshot from 2019-12-06 16-25-10

That was pretty simple.

.github/CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 
 ## Reporting issues
 
-Please [open an issue](../../../issues/new/choose) if you find a bug or have a feature request.
+Please [open an issue](/../../../issues/new/choose) if you find a bug or have a feature request.
```
gromakovsky: For me it does not work [here](https://github.com/serokell/crossref-verifier/blob/2a27f239c6c7786081efbce79605bf87a1722111/.github/CONTRIBUTING.md). Why do you want to make it repository-independent? It seems to be fragile. If it's not documented anywhere (which is the case AFAIU), I think it can be silently broken anytime. So I'd rather use full link (https://github.com/serokell/crossref-verifier/issues/new/choose) instead.

gromakovsky: And btw, if the repo is moved somewhere outside GitHub, it will likely stop working. And if it's moved to another place on GitHub, it will normally keep working because there will be a redirection. And maybe there are ways to open this file so that the link won't work.

Martoon-00: Okay, since it does not work everywhere, let's use total link.

_I'm already anticipating the moment when we rename the repository and I regret it._

gromakovsky: Well, after renaming nothing will break because redirection (it's even if we forget to update it right after the renaming). It will just look a bit ugly (but `../../issues` doesn't look cool either, at least to me) and maybe will break after a long amount of time (most likely we'll change it earlier).

Martoon-00:
Before merging, I'd like someone else to take a bird-eye view on resulting [README](https://github.com/serokell/crossref-verifier/tree/martoon/int128-nice-readme#crossref-verifier).
Does it look right / pleasant / attractive? Is anything missing?

@kirelagin Maybe you could take a look?

gromakovsky:
Or maybe @NaeosPsy

gromakovsky:
@Martoon-00 I have pinged @kirelagin in a call. Since one "technical" person (i. e. me) has already reviewed README, it makes more sense for some non-technical person to take another look, I will ping @NaeosPsy in Slack. Maybe Kirill will take a look as well, but he can't promise that.

PR #10, opened 2019-12-09
default.nix: return a derivation instead of a set

zhenyavinogradov:
`default.nix` was returning a set with a single derivation instead of a derivation. It works with `nix run` and `nix-build`, because they traverse all values in a set, but doesn't allow to use the package in places where a derivation is expected.


gromakovsky:
It would be better if you put the explanation of this change into the commit message (and then GitHub would automatically put it into PR description).

PR #16, opened 2019-12-21
[INT-111] Allow to build crossref-verify statically

balsoft:
## Description

Change `crossref-verifier.nix` and add `static.nix` so that building a fully static executable is possible.

Results look like this:
[crossref-verify.gz](https://github.com/serokell/crossref-verifier/files/3991431/crossref-verify.gz)

It's probably better if someone checks this on multiple distros before we merge.

## Related issue(s)

https://issues.serokell.io/issue/INT-111

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -4,8 +4,16 @@
 
 steps:
   - command: nix-build crossref-verifier.nix
-    label: build everything
+    label: Library and tests
+  - command: nix-build
+    label: Executable
+    artifact_paths:
+      - "result/bin/crossref-verify"
   - command: nix run -f. -c crossref-verify
-    label: crossref-verify itself (import as if done externally)
+    label: Crossref-verify itself
   - command: nix run -f $(nix eval -f nix/sources.nix --raw nixpkgs) reuse -c reuse lint
-    label: reuse lint
+    label: REUSE lint
+  - command:
+      - GITHUB_TOKEN=$(cat ~/niv-bot-token) NIX_PATH=nixpkgs=$(nix eval --raw '(import ./nix/sources.nix).nixpkgs') nix-shell -p curl git gitAndTools.hub --run "curl https://raw.githubusercontent.com/serokell/scratch/release-binary/scripts/release-binary.sh | bash"
+    label: Create a pre-release
+    branches: master
--- a/release/default.nix
+++ b/release/default.nix
@@ -1,0 +1,20 @@
+# SPDX-FileCopyrightText: 2019 Serokell <https://serokell.io>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+{ pkgs ? import (import ../nix/sources.nix).nixpkgs { } }:
+let
+  executable = (import ../crossref-verifier.nix {
+    static = true;
+  }).components.exes.crossref-verify;
+  joinByBasename = name: paths:
+    pkgs.runCommandNoCC name { } (''
+      mkdir $out
+    '' + pkgs.lib.concatMapStrings (path: ''
+      ln -s ${path} $out/${builtins.baseNameOf path}
+    '') paths);
+in joinByBasename "crossref-verifier-release" [
+  "${executable}/bin"
+  ../LICENSES
+  ../README.md
+]
--- a/crossref-verifier.nix
+++ b/crossref-verifier.nix
@@ -2,10 +2,11 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
+{ static ? true }:
 let
   sources = import ./nix/sources.nix;
   nixpkgs = import sources.nixpkgs (import sources."haskell.nix");
-  hn = nixpkgs.haskell-nix;
+  hn = if static then nixpkgs.pkgsCross.musl64.haskell-nix else nixpkgs.haskell-nix;
   project = hn.stackProject {
     src = hn.haskellLib.cleanGit { src = ./.; };
     modules = [{
@@ -14,6 +15,21 @@ let
         postHaddock = ''
           [[ -z "$(ls -A dist/doc/html)" ]] && exit 1 || echo "haddock successfully generated documentation"'';
         package.ghcOptions = "-Werror";
+        components.exes.crossref-verify.configureFlags =
+          with nixpkgs.pkgsStatic;
+          lib.optionals static [
+            "--disable-executable-dynamic"
+            "--disable-shared"
+            "--ghc-option=-optl=-pthread"
+            "--ghc-option=-optl=-static"
+            "--ghc-option=-optl=-lc"
+            "--ghc-option=-optl=-lz"
+            "--ghc-option=-optl=-lgmp"
+            "--ghc-option=-optl=-lffi"
+            "--ghc-option=-optl=-L${gmp6}/lib"
+            "--ghc-option=-optl=-L${zlib.static}/lib"
+            "--ghc-option=-optl=-L${libffi}/lib"
+          ];
       };
     }];
     cache = with sources; [{
--- a/crossref-verifier.nix
+++ b/crossref-verifier.nix
@@ -6,7 +6,10 @@
 let
   sources = import ./nix/sources.nix;
   nixpkgs = import sources.nixpkgs (import sources."haskell.nix");
-  hn = if static then nixpkgs.pkgsCross.musl64.haskell-nix else nixpkgs.haskell-nix;
+  hn = if static then
+    nixpkgs.pkgsCross.musl64.haskell-nix
+  else
+    nixpkgs.haskell-nix;
   project = hn.stackProject {
     src = hn.haskellLib.cleanGit { src = ./.; };
     modules = [{
--- a/default.nix
+++ b/default.nix
@@ -2,4 +2,4 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-(import ./crossref-verifier.nix).components.exes.crossref-verify
+(import ./crossref-verifier.nix {}).components.exes.crossref-verify


Martoon-00:
Thanks, it works for me on Ubuntu 17.10. Though file from distro was not executable, not sure whether should it be so or not.

balsoft:
@Martoon-00 hmm, I just manually compressed a single file with gzip, I don't think it keeps permissions.

Of course IRL we will have a more complete distribution, with LICENSE and README and everything. This is why this PR is WIP at the moment.

balsoft:
A script that builds a nice tarball and automatically creates a "nightly" pre-release draft on github is [here](https://github.com/serokell/scratch/pull/7) if anybody wants to speed this up and review it. The example of the script working can be found in "releases" of this repo.

gromakovsky:
> PS: why is there no "releases" link in github interface now?

What do you mean? It is here: https://i.imgur.com/Ns1s9eo.png

gromakovsky:
Imo, it would be more convenient if tarball contained one folder with all files. Currently it contains 3 files that are put into current directory. Though I don't know what is the standard approach.

gromakovsky:
Can you add build result to CI artifacts please?

gromakovsky:
As for testing, the binary works fine for me on archlinux. And I didn't have to `chmod`, it was executable out of the box.

balsoft:
@gromakovsky changed both the script and this to reflect your suggestions.

gromakovsky:
Still waiting for this thing:
> Can you add build result to CI artifacts please?

AFAIU all you have to do is to add `artifact_paths:`. I think all we need there is `crossref-verifier` executable. This way when we make a PR we'll be able to manually test executable that will go to a prerelease when PR is merged to `master`.

gromakovsky:
Executable from the latest prerelease works fine for me and the structure looks almost fine, but a bit weird. There is `share/doc` with `LICENSE` file and `LICENSES` folder. Also there is `README.md` which essentially contains documentation, but is not in `share/doc`. It seems better to either remove `share/doc` completely (now it is redundant, because it contains only one `LICENSE` file which is MPL and is present in `LICENSES` as well) or to make sure that all licenses go there and `README.md` as well. I don't know what piece of code creates `share/doc` and decides what should go there, but I suppose it is possible to tell that piece of code "hey, please take all the licenses and `README.md` as well".

balsoft:
@gromakovsky restrucutred the release a bit

release/default.nix
```diff
@@ -0,0 +1,20 @@
...
+    static = true;
+  }).components.exes.crossref-verify;
+  joinByBasename = name: paths:
+    pkgs.runCommandNoCC name { } (''
+      mkdir $out
+    '' + pkgs.lib.concatMapStrings (path: ''
+      ln -s ${path} $out/${builtins.baseNameOf path}
+    '') paths);
+in joinByBasename "crossref-verifier-release" [
+  "${executable}/bin"
```
zhenyavinogradov: As I understand, the static binary is needed to make it easy to run crossref-verifier from another CI, so how do I do it? Also, `./default.nix` is not needed anymore as well as support for building a dynamic executable.

release/default.nix
```diff
@@ -0,0 +1,20 @@
...
+  ../LICENSES
```
zhenyavinogradov: Why do you include the licences folder in the release?

release/default.nix
```diff
@@ -0,0 +1,20 @@
...
+  ../README.md
```
zhenyavinogradov: README is definitely redundant in the archive, everyone will use the github page to learn how to download and use crossref-verifier

zhenyavinogradov:
I think if you use github releases for publishing, you should also spend time on putting a list of changes there, so people will know what will change when they update.

zhenyavinogradov:
Newer haskell.nix allows to change `dontStrip` parameter to strip debug information from the binary, we should use it

release/default.nix
balsoft: To run a static version,
```
wget https://github.com/serokell/crossref-verifier/releases/download/auto-release/release.tar.gz
tar -xf release.tar.gz
./crossref-verifier/bin/crossref-verify
```

And I believe it's better to build a dynamic executable for regular CI because it should generally be faster due to better binary cache support for it.

release/default.nix
balsoft: Because licensing information is not included in the binary I don't think and it's better to include it in the distribution for legal reasons.

release/default.nix
balsoft: Maybe. Can @gromakovsky and @Martoon-00 express their opinion about this?

release/default.nix
zhenyavinogradov: I'm bad with legal stuff, can you explain the reasons?

release/default.nix
zhenyavinogradov: What do you mean by binary cache support?

release/default.nix
balsoft:
```
   3. Responsibilities

      3.1. Distribution of Source Form

All distribution of Covered Software in Source Code Form, including any Modifications
that You create or to which You contribute, must be under the terms of this
License. You must inform recipients that the Source Code Form of the Covered
Software is governed by the terms of this License, and how they can obtain
a copy of this License. You may not attempt to alter or restrict the recipients'
rights in the Source Code Form.

      3.2. Distribution of Executable Form

      If You distribute Covered Software in Executable Form then:

(a) such Covered Software must also be made available in Source Code Form,
as described in Section 3.1, and You must inform recipients of the Executable
Form how they can obtain a copy of such Source Code Form by reasonable means
in a timely manner, at a charge no more than the cost of distribution to the
recipient; and

(b) You may distribute such Executable Form under the terms of this License,
or sublicense it under different terms, provided that the license for the
Executable Form does not attempt to limit or alter the recipients' rights
in the Source Code Form under this License.
```

I'm definitely no lawyer either, but it appears to me that distributing the license with the executable is a good idea even if not required.

zhenyavinogradov: Ok

release/default.nix
Martoon-00: I don't know how it is usually done, though assuming that users have read README at GitHub upon getting the archive seems like a sane assumption to me.

gromakovsky: > I don't know how it is usually done

Same here. I think I've seen several repos where releases contain documentation files (README and others present in the repo), but I don't know the reasoning behind that. E. g. `stack` releases seem to contain `README.md` and some other docs. It seems to me that putting `README.md` is useful only if we want to put a link to this archive to some website.

So, on the one hand `README.md` seems to be useless currently, but on the other hand it makes the archive more self-contained and suitable for other ways of distribution. Also it seems to be quite harmless: if we have to put licenses (which is being discussed in another conversation here), it should be an archive anyway and one `README.md` file is negligible. Though if we conclude that including licenses is not necessary, we'll be able to put only the binary into releases and then people won't have to extract it.

release/default.nix
gromakovsky: GitHub seems to put source code into each release by default (there is `Source code` in assets [here](https://github.com/serokell/crossref-verifier/releases/tag/auto-release)). And AFAIU for distribution in Executable Form we only have to make it easy to obtain Source Code.

So:
1. Licenses are not necessary currently because if a person goes to GitHub releases to download this software in Executable Form the will clearly see how to download this software in Source Form.
2. If we ever want to distribute `crossref-verifier` somewhere else (outside of GitHub releases), including only the license is not sufficient, we should also include Source Code or a link to it (or somehow else tell people how to obtain Source Code).

At the same time I see that we use a thing with MIT license and it says that we can do whatever we want with that software, but:
> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

So apparently that license requires us to include at least the MIT license into distribution. I don't know whether `release.tar.gz` is considered as its own "distribution" or only the whole GitHub release can be considered as a distribution. In the latter case including the license seems to be unnecessary because it is part of "Source code".

Without further investigation I agree that including all licenses is safer anyway, but I'd like to get answers, will try asking in #legal.

gromakovsky:
> I think if you use github releases for publishing, you should also spend time on putting a list of changes there, so people will know what will change when they update.

Yes, sure, we should add a changelog file and maintain it. In other projects this year my strategy was to add a changelog before the first release and write something like "MVP release" and maybe briefly mention features in this release. And then either require to update the changelog on each externally visible change (by explicitly asking people and adding a checkmark to PR template) or just update it prior to each upcoming release.

For the very first version changelog is not very useful. It is mostly needed to compare two existing versions rather than 1 existing version and nothing.

.buildkite/pipeline.yml
```diff
@@ -9,3 +9,12 @@ steps:
...
     label: reuse lint
+  - command: nix-build release -o result-release
+    label: "Build the release"
+    artifact_paths:
+      - "result-release/bin/*"
+  - command:
+      - GITHUB_TOKEN=$(cat ~/niv-bot-token) NIX_PATH=nixpkgs=channel:nixos-unstable nix-shell -p git gitAndTools.hub --run "curl https://raw.githubusercontent.com/serokell/scratch/587aded2cd1bb624f85b49727272616fdc49f4fc/scripts/release-binary.sh | bash"
+    label: Create a release
+    # if:
+    #  build.branch == "master"
```
gromakovsky: I think you can specify `branches: master` instead and it looks a bit neater.

.buildkite/pipeline.yml
```diff
@@ -9,3 +9,12 @@ steps:
     label: crossref-verify itself (import as if done externally)
   - command: nix run -f $(nix eval -f nix/sources.nix --raw nixpkgs) reuse -c reuse lint
     label: reuse lint
+  - command: nix-build release -o result-release
+    label: "Build the release"
+    artifact_paths:
+      - "result-release/bin/*"
+  - command:
+      - GITHUB_TOKEN=$(cat ~/niv-bot-token) NIX_PATH=nixpkgs=channel:nixos-unstable nix-shell -p git gitAndTools.hub --run "curl https://raw.githubusercontent.com/serokell/scratch/587aded2cd1bb624f85b49727272616fdc49f4fc/scripts/release-binary.sh | bash"
+    label: Create a release
```
gromakovsky: Might be better to call it a "pre-release" here.

release/default.nix
balsoft: Yeah, better be safe than sorry and 28kB (~7kB compressed) is not a big deal compared to the size of the executable anyways.

.buildkite/pipeline.yml
balsoft: Fixed.

.buildkite/pipeline.yml
balsoft: Fixed.

balsoft:
@gromakovsky so the changelogs are not my responsibility, correct?

gromakovsky:
> @gromakovsky so the changelogs are not my responsibility, correct?

Yep, we'll add it later. I didn't plan to make a release right after merging this PR, we want to at least rename the repo, so at least one more PR will be made anyway.

In [`universum`](https://github.com/serokell/universum/releases) I tend to write 
```
For more details please see [the changelog](CHANGES.md#161).
```

in each release.

And the situation is similar to `README.md`: it's not obvious whether `CHANGES.md` should be part of the archive or not. Anyway, adding it to the archive later looks trivial (1 line in  `release/default.nix`).

balsoft:
@gromakovsky also, build artifacts seem to function correctly now.

release/default.nix
balsoft: I mean there are no pre-built libraries and compilers for musl upstream.

.buildkite/pipeline.yml
```diff
@@ -14,6 +14,6 @@ steps:
     artifact_paths:
       - "result-release/bin/*"
   - command:
-      - GITHUB_TOKEN=$(cat ~/niv-bot-token) NIX_PATH=nixpkgs=$(nix eval --raw '(import ./nix/sources.nix).nixpkgs') nix-shell -p curl git gitAndTools.hub --run "curl https://raw.githubusercontent.com/serokell/scratch/587aded2cd1bb624f85b49727272616fdc49f4fc/scripts/release-binary.sh | bash"
+      - GITHUB_TOKEN=$(cat ~/niv-bot-token) NIX_PATH=nixpkgs=$(nix eval --raw '(import ./nix/sources.nix).nixpkgs') nix-shell -p curl git gitAndTools.hub --run "curl https://raw.githubusercontent.com/serokell/scratch/master/scripts/release-binary.sh | bash"
```
gromakovsky: I'll copy from `scratch` PR:

> I'd rather not do it this way because one day someone may move this script into another location and things will stop working for everybody. Let's use something permanent instead of master.

Such things do happen all the time. Let's push `release-binary` tag and use it here?

balsoft: Done

release/default.nix
zhenyavinogradov: Ok, let's include the license for safety. But are you sure we need the whole LICENSES and not the LICENSE file?

release/default.nix
zhenyavinogradov: In most cases they will rebuild only when we repin nixpkgs, and the effect is negligible compared to the fact that we have to rebuild ghc and all haskell packages as well. But it would be simpler for us to have to maintain the static version only

balsoft:
@zhenyavinogradov Now everything is static. If it takes too long, building dynamically is a really simple change.

zhenyavinogradov:
Ok, good. What do you think about the `dontStrip` option?

release/default.nix
gromakovsky: Well, different licenses apply to different source files. In fact, the `LICENSE` file is present primarily because GitHub recognizes it and shows on the main repo page.

But it doesn't mean that all source files are covered by only this license. E. g. `sources.nix` belongs to nmattia and there the MIT license applies and it requires to put its copyright notice into anything you distribute. And `source.nix` participates in building the archive.

zhenyavinogradov: Ok, I'm clearly far from being a lawyer, let's keep it like that. But out of interest, I wonder what would be different if we imported the `sources.nix` file from an external url without including it into the repository, like we do with the libraries that we fetch from git. Would we not have to include the copyright notice then?

gromakovsky: I think no, but I don't have any explanation. Feel free to ask in #legal.

balsoft:
I'll merge now and will implement `dontStrip` logic some time soon.

PR #24, opened 2020-02-24
Add `ignored` CLI option

### LINKED ISSUE #19: Add --exclude option

Now that we have static binary in releases, one can use `xrefcheck` like this:
> wget -O - https://github.com/serokell/xrefcheck/releases/download/v0.1.0.0/release.tar.gz | tar xzv && xrefcheck/bin/xrefcheck

However, it will also check references in `xrefcheck/README.md` and report some errors.

Since we want to make it possible to add just one line to CI config to run `xrefcheck`, I think the easiest solution is to add `--exclude` option and pass `--exclude xrefcheck`.

Of course one can download/unpack tarball somewhere outside of current directory or supply appropriate config, but the former somewhat depends on particular CI system and the latter requires more than just one line.

# Acceptance criteria

It should be possible to pass `--exclude X` (maybe multiple times) to specify that all files in `X` (if it's a directory) or `X` itself (if it is a file) ought to be ignored.

Martoon-00:
## Description

Problem: it is often a case that in CI we copy xrefcheck repository into
the verified project, and we want to exclude folder with xrefcheck from
scan. Considering this in config is not too convenient, better add a CLI
option.

Solution: add this CLI option.

As in config, I call it `not-scanned`, because simpler `excluded` is too
general and it's not obvious what exactly it refers to - the set of
verified files or the set of supposedly non-existing files (in the
latter case references _to_ such files should also be prohibited).
## Related issue(s)

- Fixes partially #19 

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: MPL-2.0
 
 name:                xrefcheck
-version:             0.1.0.0
+version:             0.1.1
 github:              serokell/xrefcheck
 license:             MPL-2.0
 license-file:        LICENSE
--- a/src-files/def-config.yaml
+++ b/src-files/def-config.yaml
@@ -4,7 +4,7 @@
 
 # Parameters of repository traversal.
 traversal:
-  # Folders which we pretend do not exist
+  # Files and folders which we pretend do not exist
   # (so they are neither analyzed nor can be referenced).
   ignored:
     # Git files
--- a/src-files/def-config.yaml
+++ b/src-files/def-config.yaml
@@ -23,7 +23,7 @@ verification:
   # declaring "Response timeout".
   externalRefCheckTimeout: 10s
 
-  # File prefixes, references in which should not be analyzed.
+  # Prefixes of files, references in which should not be analyzed.
   notScanned:
     # GitHub-specific files
     - .github/pull_request_template.md
--- a/src/Xrefcheck/CLI.hs
+++ b/src/Xrefcheck/CLI.hs
@@ -11,6 +11,8 @@ module Xrefcheck.CLI
     , shouldCheckExternal
     , Command (..)
     , Options (..)
+    , TraversalOptions (..)
+    , addTraversalOptions
     , defaultConfigPaths
     , getCommand
     ) where
@@ -21,6 +23,7 @@ import Options.Applicative (Parser, ReadM, command, eitherReader, execParser, fu
                             short, strOption, switch, value)
 import Paths_xrefcheck (version)
 
+import Xrefcheck.Config
 import Xrefcheck.Core
 
 modeReadM :: ReadM VerifyMode
@@ -43,12 +46,24 @@ data Command
   | DumpConfig FilePath
 
 data Options = Options
-    { oConfigPath      :: Maybe FilePath
-    , oRoot            :: FilePath
-    , oMode            :: VerifyMode
-    , oVerbose         :: Bool
-    , oShowProgressBar :: Bool
+    { oConfigPath       :: Maybe FilePath
+    , oRoot             :: FilePath
+    , oMode             :: VerifyMode
+    , oVerbose          :: Bool
+    , oShowProgressBar  :: Bool
+    , oTraversalOptions :: TraversalOptions
     }
+
+data TraversalOptions = TraversalOptions
+    { toIgnored :: [FilePath]
+    }
+
+addTraversalOptions :: TraversalConfig -> TraversalOptions -> TraversalConfig
+addTraversalOptions TraversalConfig{..} (TraversalOptions ignored) =
+  TraversalConfig
+  { tcIgnored = tcIgnored ++ ignored
+  , ..
+  }
 
 -- | Where to try to seek configuration if specific path is not set.
 defaultConfigPaths :: [FilePath]
@@ -87,7 +102,16 @@ optionsParser = do
     oShowProgressBar <- fmap not . switch $
         long "no-progress" <>
         help "Do not display progress bar during verification."
+    oTraversalOptions <- traversalOptionsParser
     return Options{..}
+
+traversalOptionsParser :: Parser TraversalOptions
+traversalOptionsParser = do
+    toIgnored <- many . strOption $
+        long "ignored" <>
+        metavar "FILEPATH" <>
+        help "Files of files which we pretend do not exist."
+    return TraversalOptions{..}
 
 dumpConfigOptions :: Parser FilePath
 dumpConfigOptions = hsubparser $
--- a/src/Xrefcheck/CLI.hs
+++ b/src/Xrefcheck/CLI.hs
@@ -110,7 +110,7 @@ traversalOptionsParser = do
     toIgnored <- many . strOption $
         long "ignored" <>
         metavar "FILEPATH" <>
-        help "Files of files which we pretend do not exist."
+        help "Files and folders which we pretend do not exist."
     return TraversalOptions{..}
 
 dumpConfigOptions :: Parser FilePath
--- a/src/Xrefcheck/System.hs
+++ b/src/Xrefcheck/System.hs
@@ -12,8 +12,8 @@ module Xrefcheck.System
 import Data.Aeson (FromJSON (..), withText)
 import GHC.IO.Unsafe (unsafePerformIO)
 import System.Directory (canonicalizePath)
+import System.FilePath ((</>))
 import qualified System.FilePath.Glob as Glob
-import System.FilePath.Posix ((</>))
 
 -- | We can quite safely treat surrounding filesystem as frozen,
 -- so IO reading operations can be turned into pure values.
--- a/src/Xrefcheck/System.hs
+++ b/src/Xrefcheck/System.hs
@@ -5,25 +5,43 @@
 
 module Xrefcheck.System
     ( readingSystem
-    , CanonicalizedGlobPattern (..)
+    , RelGlobPattern (..)
+    , bindGlobPattern
     ) where
 
 import Data.Aeson (FromJSON (..), withText)
 import GHC.IO.Unsafe (unsafePerformIO)
 import System.Directory (canonicalizePath)
 import qualified System.FilePath.Glob as Glob
+import System.FilePath.Posix ((</>))
 
 -- | We can quite safely treat surrounding filesystem as frozen,
 -- so IO reading operations can be turned into pure values.
 readingSystem :: IO a -> a
 readingSystem = unsafePerformIO
 
--- | Glob pattern with 'canonicalizePath' applied O_o.
-newtype CanonicalizedGlobPattern = CanonicalizedGlobPattern Glob.Pattern
+-- | Glob pattern relative to repository root.
+newtype RelGlobPattern = RelGlobPattern FilePath
 
-instance FromJSON CanonicalizedGlobPattern where
-    parseJSON = withText "Repo-rooted glob pattern" $ \path -> do
-        let !cpath = readingSystem $ canonicalizePath (toString path)
-        cpat <- Glob.tryCompileWith Glob.compDefault cpath
-                & either fail pure
-        return $ CanonicalizedGlobPattern cpat
+bindGlobPattern :: FilePath -> RelGlobPattern -> Glob.Pattern
+bindGlobPattern root (RelGlobPattern relPat) = readingSystem $ do
+  -- TODO [#26] try to avoid using canonicalization
+  absPat <- canonicalizePath (root </> relPat)
+  case Glob.tryCompileWith globCompileOptions absPat of
+    Left err ->
+      error $ "Glob pattern compilation failed after canonicalization: " <>
+              toText err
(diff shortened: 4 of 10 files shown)


src/Xrefcheck/CLI.hs
```diff
@@ -87,8 +103,18 @@ optionsParser = do
...
         help "Do not display progress bar during verification."
+    oVerifyOptions <- verifyOptionsParser
     return Options{..}
 
+verifyOptionsParser :: Parser VerifyOptions
+verifyOptionsParser = do
+    voNotScanned <- many . strOption $
+        long "not-scanned" <>
+        metavar "FILEPATH" <>
+        help "Prefixes of files, references inwhich should not be analyzed.\n\
```
gromakovsky:
```suggestion
        help "Prefixes of files, references in which should not be analyzed.\n\
```

src/Xrefcheck/CLI.hs
```diff
@@ -48,8 +51,21 @@ data Options = Options
     , oMode            :: VerifyMode
     , oVerbose         :: Bool
     , oShowProgressBar :: Bool
+    , oVerifyOptions   :: VerifyOptions
     }
 
+data VerifyOptions = VerifyOptions
+    { voNotScanned :: [FilePath]
+      -- ^ Folders, references in files of which we should not analyze.
```
gromakovsky: Here you say "Folders", below you say "Prefixes of files". It sounds a bit obscure and ambiguous, can you clarify the exact semantic? Note that in the issue I requested that it should be a folder or a file (in the former case everything in the folder is not scanned, in the latter case the file is not scanned).

Martoon-00: Mda.

I fixed to be "file prefix" as in config, in case of file it should be just filepath, in case of folder - filepath terminated with `/` (file prefix `src` would match both `src` and `src-files` folders, but `src/` would match only the former).

Actually, what do you think if I try to change it to glob pattern? That seems to me a more generic solution, but may be breaking for other projects. 

In either case, I think I will change those `FilePath`s to some custom type in order to make their semantics expressed in types.

gromakovsky: > I fixed to be "file prefix" as in config, in case of file it should be just filepath

Filepath or any prefix of filepath? Will `a` match everything that starts from `a` (folders and files)? Will empty string match everything (or is it not allowed)?

> Actually, what do you think if I try to change it to glob pattern? That seems to me a more generic solution, but may be breaking for other projects.

If `a` matches everything that starts from `a` then I think glob patterns are much saner. If `a` matches only `a`, using glob patterns seems to be a breaking change only in some weird corner cases (like when you have `*` in a file name). So in the former case I would do it because it's saner, in the latter case I would it unless it appears to be complicated.

Martoon-00: > Filepath or any prefix of filepath? Will a match everything that starts from a (folders and files)? Will empty string match everything (or is it not allowed)?

Yes, that's all true. Sounds not very good.

I'm going to try glob patterns then.

src-files/def-config.yaml
```diff
@@ -4,14 +4,14 @@
 
 # Parameters of repository traversal.
 traversal:
-  # Folders which we pretend do not exist
+  # Glob patterns describing files which we pretend do not exist
   # (so they are neither analyzed nor can be referenced).
   ignored:
     # Git files
-    - .git
+    - .git/**/*
```
gromakovsky: Is it equivalent to `.git/`?

src/Xrefcheck/System.hs
```diff
@@ -5,25 +5,43 @@
...
-    , CanonicalizedGlobPattern (..)
+    , RelGlobPattern (..)
+    , bindGlobPattern
     ) where
 
 import Data.Aeson (FromJSON (..), withText)
 import GHC.IO.Unsafe (unsafePerformIO)
 import System.Directory (canonicalizePath)
 import qualified System.FilePath.Glob as Glob
+import System.FilePath.Posix ((</>))
```
gromakovsky: Why `Posix`? I mean, is it intentional and do you want it to still be `Posix` if someone tries to build it on Windows?

src-files/def-config.yaml
Martoon-00: In glob language (at least supposed by library I use), this means "match any file in `.git` folder", and this is the only way to express that.

gromakovsky: I see there is `filterExcludedDirs` in `gatherRepoInfo`, how is it supposed to work? Which dirs are excluded?

src/Xrefcheck/System.hs
Martoon-00: Looks so, in glob patterns they (authors of that library) consider whatever filepath separator the system has, and `</>` also inserts `\\` in Windows.

src-files/def-config.yaml
Martoon-00: I just came to the same thing :man_facepalming: 

Apparently, here I should still expect folder (or, alternatively, file) name, this should be easy to handle without globs.

gromakovsky:
Now my question is: why can't we use the same logic for `ignored` and `not-scanned`? I think it makes sense to use glob patterns in both cases and treat them as either files or folders. So that I can write `doc*` in `ignored` and in `not-scanned` and it will match `doc.txt` file and `doccop/` folder.

You said that the only way to not scan `doccop/` folder that is supported by the library you use is `doccop/**/*`, but I think it's because you treat these things only as files, but not as folders. I am not sure what's the best way to implement what I suggest, but for example can't you do something like the following: when you traverse the repo and encounter a folder that is not present in `ignored`, but is present in `not-scanned`, can't you somehow store the information that you are in `not-scanned` folder and mark all files in that folder as `not-scanned`?

Martoon-00:
> Now my question is: why can't we use the same logic for `ignored` and `not-scanned`? I think it makes sense to use glob patterns in both cases and treat them as either files or folders. So that I can write `doc*` in `ignored` and in `not-scanned` and it will match `doc.txt` file and `doccop/` folder.
> 
> You said that the only way to not scan `doccop/` folder that is supported by the library you use is `doccop/**/*`, but I think it's because you treat these things only as files, but not as folders. I am not sure what's the best way to implement what I suggest, but for example can't you do something like the following: when you traverse the repo and encounter a folder that is not present in `ignored`, but is present in `not-scanned`, can't you somehow store the information that you are in `not-scanned` folder and mark all files in that folder as `not-scanned`?

Do I understand it correctly that you propose to consider exclusion patterns simultaneously as glob for files and as precise folder name? So that if the user specifies `doccop/**/*`, this matches against all the `doccop` folder content, and if he specifies just `doccop`, then it matches against this folder and all its contents?

gromakovsky:
I propose to handle it the same as `.gitignore` is handled by `git`. It is well known and I think nobody will have issues with that, also it is quite flexible and intuitive (IMO). Patterns are globs for files as well as for folders.

> So that if the user specifies `doccop/**/*`, this matches against all the doccop folder content

Yes, all folders and files in `doccop/` (recursively).

>  if he specifies just doccop, then it matches against this folder and all its contents?

Yes, also all files named `doccop`. And if there is `foo/doccop` it should match as well. But `/doccop` will match only the root `doccop` (file or folder). And `/doccop/` will match only the root `doccop` folder.

`/docc?p` should also match the root `doccop/` folder or `docccp` folder or `docc2p` file.

I think it's the best we can do, and the only reason not to do it (IMO) is if it is hard. But currently it doesn't sound much harder than what's already done, feel free to convince me that it actually is hard :wink:

gromakovsky:
There is one difference with `gitignore`, though: `git` does not really care about folders in the end, it keeps track of files. If you create an empty folder `git` won't notice it (until you put a file there). We, on the other hand, do care about folders: one can have a reference to a folder. So a consequence is that `doccop/**/*` is apparently equivalent to `doccop/` in case of `.gitignore`, but in our case they are not (in `traversal`, not in `verification`) if I got everything right. But I think it's not a big problem anyway.

Martoon-00:
> I propose to handle it the same as `.gitignore` is handled by `git`. It is well known and I think nobody will have issues with that, also it is quite flexible and intuitive (IMO). Patterns are globs for files as well as for folders.

That's what I wanted to do from the very beginning, but it seemed difficult.

Okay, let me try, maybe it is easy to express .gitignore patterns via that library.

gromakovsky:
> but it seemed difficult.

I initially imagined this task to be quick and simple and didn't expect any glob patterns. But then you proposed them (maybe to a smaller extent), some discussions started, the task stopped being simple (and quick), so now it seems reasonable to invest a bit more time and come up with a perfect solution.

Martoon-00:
Okay, going with .gitignore-like patterns is not trivial. There is a case that `**` in gitignore can match zero folders, while in `Glob` package `**` matches one or more folders, so it is not trivial to a adopt. Probably I will switch to an alternative `filepattern` package which seemingly supplies less efficient matching, but `**` there works in the same way as in gitignore. Though this apparently is better done in a separate PR.

I removed commit with glob stuff for now, please check.

gromakovsky:
AFAIU `--exclude` option I describe in #19 should behave as `ignored` in config, not as `notScanned`. The difference is that references TO files in `ignored` folders are invalid, but references TO files in `notScanned` folders are ok.

If some file has an invalid reference to `xrefcheck` folder that does not exist, it may accidentally become valid when we do `wget -O - https://github.com/serokell/xrefcheck/releases/download/v0.1.0.0/release.tar.gz | tar xzv `. Why did you choose `notScanned` semantic instead of `ignored`?

That being said, the difference is quite minor and is unlikely to matter in practice, so we can postpone it till a follow-up PR (which will hopefully appear in this life). Ideally I think all options from the config should be available in CLI as well, so `--not-scanned` makes sense in any case.

gromakovsky:
> I think this change is worth bumping the version btw.

Bump. I'd set the version to 0.1.1 and update the changelog.

gromakovsky:
Okay, I have tested and my immediate problem is resolved. Let's make a new release and I will be satisfied. Will approve after you update version and changelog and rebase.

> Fixed #19

I propose to keep this issue open and create a new one for now. Hopefully one day you will find time to implement glob stuff properly and add tests for corner cases, because currently it's hard to understand how `--not-scanned` option and some things in config work.

Martoon-00:
Ah right, your concern about using `--ignored` is valid.

I will replace `--not-scanned` with `--ignored` then as I want to keep the number of CLI options as few as possible. If we reflect all the config in CLI options, this will encourage CI care-taker specifying all options in CLI, while it should be possible for a mere developer having a local copy of repo to just run `xrefcheck` without any options and check links in the same way CI does it.

gromakovsky:
Okay, let's do this way. Btw,
> Partially fixes #19

I think it will still close the issue. Probably `Fixes partially` won't.

Martoon-00:
Oh right, changed.

gromakovsky:
LGTM, please rebase on latest `master`, squash commits and then I will make final test and hopefully approve.

src/Xrefcheck/System.hs
gromakovsky: @Martoon-00 I am sorry, buy I still don't understand why `Posix`. And now that there was some effort to make this thing work on Windows, I'd like to get back to this question.

Just in case:
*    System.FilePath.Posix manipulates POSIX/Linux style FilePath values (with / as the path separator).
*    System.FilePath.Windows manipulates Windows style FilePath values (with either \ or / as the path separator, and deals with drives).
*    System.FilePath is an alias for the module appropriate to your platform.

So your `</>` is always `/`, even on Windows. You do: `canonicalizePath (root </> relPat)`. On Windows this `root </> relPat` can be something like `foo/bar` and `canonicalizePath` (I guess) won't treat `/` as a separator. So why is it done this way?

package.yaml
```diff
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: MPL-2.0
 
 name:                xrefcheck
-version:             0.1.0.0
+version:             0.1.1
```
gromakovsky, line 6: Shouldn't changelog be updated accordingly?

src/Xrefcheck/System.hs
Martoon-00: Oh I got your point, that was really weird :facepalm:

Martoon-00: I inserted a commit prior to version bump. There are many places where import should be fixed.

Martoon-00: And that could be the reason why `qReadFileBS` didn't work and @balsoft had to replace it with `embedFile`, my bad.

balsoft: No, it's not indeed the reason. The problem with `qReadFileBS` lies deeper within the architecture of Template Haskell itself and it requires fixing it and not this.

(Obviously I tried replacing `Posix` with `Windows` here and it didn't help a bit)

Martoon-00: Ah okay, thanks for the note.

PR #29, opened 2020-04-18
Add a windows build

balsoft:
## Description

Build an executable for Windows. This requires a proper review from Haskell people, as well as some updates to documentation and release process. To build a windows executable, `nix build -f. --argstr platform windows` .

## Related issue(s)

None

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -9,6 +9,10 @@ steps:
     label: Executable
     artifact_paths:
       - "result/bin/xrefcheck"
+  - command: nix-build --argstr platform windows
+    label: Windows executable
+    artifact_paths:
+      - "result/bin/*"
   - command: nix run -f. -c xrefcheck
     label: Xrefcheck itself
   - command: nix run -f $(nix eval --raw '((import ./nix/sources.nix).nixpkgs)') reuse -c reuse lint
--- a/src/Xrefcheck/Config.hs
+++ b/src/Xrefcheck/Config.hs
@@ -11,9 +11,11 @@ import Data.Aeson.Options (defaultOptions)
 import Data.Aeson.TH (deriveFromJSON)
 import Data.Yaml (FromJSON (..), decodeEither', prettyPrintParseException, withText)
 import Instances.TH.Lift ()
-import qualified Language.Haskell.TH.Syntax as TH
-import System.FilePath ((</>))
-import TH.RelativePaths (qReadFileBS)
+
+-- FIXME: Use </> from System.FilePath
+-- </> from Posix is used only because we cross-compile to Windows and \ doesn't work on Linux
+import System.FilePath.Posix ((</>))
+import Data.FileEmbed (embedFile)
 import Time (KnownRatName, Second, Time, unitsP)
 
 import Xrefcheck.System (RelGlobPattern)
@@ -50,7 +52,7 @@ data VerifyConfig = VerifyConfig
 -- would be lost.
 defConfigText :: ByteString
 defConfigText =
-  $(TH.lift =<< qReadFileBS ("src-files" </> "def-config.yaml"))
+  $(embedFile ("src-files" </> "def-config.yaml"))
 
 defConfig :: HasCallStack => Config
 defConfig =
--- a/xrefcheck.nix
+++ b/xrefcheck.nix
@@ -2,18 +2,18 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-{ static ? false }:
+{ linux ? false, linux-static ? false, windows ? false }:
 let
   sources = import ./nix/sources.nix;
   nixpkgs = import sources.nixpkgs (import sources."haskell.nix" {}).nixpkgsArgs;
-  pkgs = if static then nixpkgs.pkgsCross.musl64 else nixpkgs;
+  pkgs = if linux-static then nixpkgs.pkgsCross.musl64 else if windows then nixpkgs.pkgsCross.mingwW64 else nixpkgs;
   project = pkgs.haskell-nix.stackProject {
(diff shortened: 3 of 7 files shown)


src/Xrefcheck/Config.hs
```diff
@@ -50,7 +50,7 @@ data VerifyConfig = VerifyConfig
 -- would be lost.
 defConfigText :: ByteString
 defConfigText =
-  $(TH.lift =<< qReadFileBS ("src-files" </> "def-config.yaml"))
+  $(embedFile ("src-files" </> "def-config.yaml"))
```
Martoon-00, line 53: So it happened that `qReadFileBS` works bad on Windows, but `embedFile` does fine?

balsoft: Yes. The reason is that `qReadFileBS` gets an absolute path from `iserv`, which for windows cross-compilation runs inside wine, thus producing a path like `Z:\build\xrefcheck\src-files`, which makes no sense on Linux. `embedFile`, on the other hand, works with relative paths AFAIU and doesn't talk to `iserv` at all, so it works fine.

Martoon-00: Cool, nice that you managed to find a workaround for this.

balsoft:
Update to the release-binary script here: https://github.com/serokell/scratch/pull/11

balsoft:
Build artifacts (tested in wine, works): [xrefcheck.tar.gz](https://github.com/serokell/xrefcheck/files/4499997/xrefcheck.tar.gz)

Unfortunately, I can't build libffi for windows statically (as a `.lib`), so we'll have to figure out a solution for that.

src/Xrefcheck/Config.hs
gromakovsky: Does it mean that we should always prefer `embedFile` to `qReadFileBS`?

balsoft: @gromakovsky yes, if we want to support cross-compilation to windows, we shouldn't use `qReadFileBS`, because it simply won't work.

Martoon-00: I originally used `qReadFileBS` because it states its return type explicitly (`Q ByteString` instead of `Q Exp`), but that seems not that important taking cross-platform issues into account)

zhenyavinogradov:
Maybe we should build windows executable in CI for all branches, to see if there are windows-specific problems with the build

xrefcheck.nix
```diff
@@ -2,16 +2,17 @@
...
   sources = import ./nix/sources.nix;
   nixpkgs = import sources.nixpkgs (import sources."haskell.nix");
-  pkgs = if static then nixpkgs.pkgsCross.musl64 else nixpkgs;
+  pkgs = if static then nixpkgs.pkgsCross.musl64 else if windows then nixpkgs.pkgsCross.mingwW64 else nixpkgs;
   project = pkgs.haskell-nix.stackProject {
     src = pkgs.haskell-nix.haskellLib.cleanGit { src = ./.; };
     modules = [{
       packages.xrefcheck = {
         package.ghcOptions = "-Werror";
+        preConfigure = "ls /build/xrefcheck/src-files";
```
zhenyavinogradov, line 15: What is this for?

gromakovsky:
Oh, maybe one day someone even will resurrect company account on appveyor (or any other Windows CI) and we will get back to https://issues.serokell.io/issue/OPS-415 :pled:

balsoft:
@gromakovsky I guess we can cross-compile universum too...

balsoft:
@zhenyavinogradov now it build a windows exe for every branch.

.buildkite/pipeline.yml
```diff
@@ -9,6 +9,10 @@ steps:
     label: Executable
     artifact_paths:
       - "result/bin/xrefcheck"
+  - command: nix-build --argstr platform windows
```
zhenyavinogradov, line 12: It looks like an error happens during build, but it does not fail the build: https://buildkite.com/serokell/xrefcheck/builds/284#c03312be-8e80-4bcc-baa4-84505668ae18/65-144. Do you know if this error is critical?

balsoft:
Looks like it doesn't affect much. The executable still works in Wine, so I assume it's fine.

src/Xrefcheck/Config.hs
```diff
@@ -11,9 +11,11 @@ import Data.Aeson.Options (defaultOptions)
 import Data.Aeson.TH (deriveFromJSON)
 import Data.Yaml (FromJSON (..), decodeEither', prettyPrintParseException, withText)
 import Instances.TH.Lift ()
-import qualified Language.Haskell.TH.Syntax as TH
-import System.FilePath ((</>))
-import TH.RelativePaths (qReadFileBS)
+
+-- FIXME: Use </> from System.FilePath
```
gromakovsky, line 15: Ermm, looks like a big hack. How are we going to fix it? If we import from Posix, why not use `/` directly?

balsoft: We import from POSIX specifically so that when a fix comes, it's a small diff.

I think we have to write our own implementation of `</>` at some point that's `/` if build == POSIX and `\` if build == Windows. I don't know how to do that properly though.

Another solution would be to place the config at the root of the repository so that we don't need `</>` at all :P

gromakovsky: > I think we have to write our own implementation of `</>` at some point that's `/` if build == POSIX and `\` if build == Windows.

Shouldn't it be handled by `filepath`? Maybe it's worth creating an issue there?

balsoft: It probably should, but I believe that getting the build platform will be GHC-specific.

PR #32, opened 2020-05-15
Avoid loot-prelude dependency

### LINKED ISSUE #30: Use cabal's mixins instead of loot-prelude

Currently we use `loot-prelude` which allows us to use `Universum` as custom prelude and not type `import Universum` everywhere (it effectively exposes `Universum` as `Prelude`).

An alternative is to use `cabal`'s feature: [`mixins`](https://www.haskell.org/cabal/users-guide/developing-packages.html?highlight=mixins#pkg-field-mixins).

Pros:
* less dependencies
* no dependencies that are not on Hackage (can be solved by uploading `loot-prelude` of course)

Less dependencies ⇒ less maintenance effort. `mixins` are very unlikely to require any maintenance, bot libraries often do require something.

I see this:
> Mixin packages are part of the Backpack extension to the Haskell module system.

So maybe it won't work with `stack` or `hpack`, I don't know, let's try.

# Acceptance criteria

`loot-prelude` is not used anymore, `cabal`'s `mixins` feature is used instead.

Martoon-00:
## Description

Problem: lootbox is not uploaded to hackage, so when we upload
xrefcheck, its build on hackage fails.

Solution: we can use `mixins` feature to set Universum as prelude
without any intermediate packages.

This seems to be supported even by quite old versions of stack, like
1.7.3, so that should not bring build problems in most cases.

## Related issue(s)

Fixes #30

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,13 @@
 Unreleased
 ==========
 
+
+0.1.1.1
+=======
+
+* [#30](https://github.com/serokell/xrefcheck/pull/32)
+  + Do not depend on `loot-prelude` package.
+
 0.1.1
 =======
 
--- a/package.yaml
+++ b/package.yaml
@@ -5,7 +5,7 @@
 spec-version: 0.31.0
 
 name:                xrefcheck
-version:             0.1.1
+version:             0.1.1.1
 github:              serokell/xrefcheck
 license:             MPL-2.0
 license-file:        LICENSE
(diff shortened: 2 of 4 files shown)


Martoon-00:
I'm truly surprised (and quite glad) that it worked.

Right, since nothing seems to change since the last release, it would be nice to make a new one as a fix. I will do tomorrow.

gromakovsky:
Btw, I just noticed that 0.1.1 was not uploaded to Hackage. I don't know if we should do it :)

Martoon-00:
Oh right, I have uploaded 0.1.1 just now

PR #41, opened 2020-06-23
Fix release

balsoft:
## Description

Release script is broken. Fix it.

Also instead of tarball, add the binary itself to the artifacts.

## Related issue(s)

None

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/release/default.nix
+++ b/release/default.nix
@@ -2,19 +2,21 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-{ pkgs ? import (import ../nix/sources.nix).nixpkgs { } }:
+{ pkgs ? import (import ../nix/sources.nix).nixpkgs { }, timestamp }:
 let
-  executable = (import ../xrefcheck.nix {
-    static = true;
-  }).components.exes.xrefcheck;
-  joinByBasename = name: paths:
-    pkgs.runCommandNoCC name { } (''
-      mkdir $out
-    '' + pkgs.lib.concatMapStrings (path: ''
-      ln -s ${path} $out/${builtins.baseNameOf path}
-    '') paths);
-in joinByBasename "xrefcheck-release" [
-  "${executable}/bin"
-  ../LICENSES
-  ../README.md
+  executable =
+    (import ../xrefcheck.nix { static = true; }).components.exes.xrefcheck;
+
+  releaseNotes = pkgs.writeText "release-notes.md" ''
+    Automatic release on ${timestamp}
+  '';
+in pkgs.linkFarm "xrefcheck-release" [
+  {
+    name = "release-notes.md";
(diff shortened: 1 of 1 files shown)


gromakovsky:
Was `release-binary` tag in `scratch` force-pushed to a new revision? Why? The whole point of using a tag instead of `master` or any other branch (that changes over time) was to prevent breakages like this one. If you wanted to update that script (especially in a BREAKING way) you should have pushed a new tag, e. g. `release-binary-v2` or something. `git tag --help` calls it "The insane thing":
```
   On Re-tagging
       What should you do when you tag a wrong commit and you would want to re-tag?

       If you never pushed anything out, just re-tag it. Use "-f" to replace the old one. And you’re done.

       But if you have pushed things out (or others could just read your repository directly), then others will have already seen
       the old tag. In that case you can do one of two things:

        1. The sane thing. Just admit you screwed up, and use a different name. Others have already seen one tag-name, and if you
           keep the same name, you may be in the situation that two people both have "version X", but they actually have different
           "X"'s. So just call it "X.1" and be done with it.

        2. The insane thing. You really want to call the new version "X" too, even though others have already seen the old one. So
           just use git tag -f again, as if you hadn’t already published the old one.
```

balsoft:
Yep, I fucked up. WIll try to avoid in the future.

release/default.nix
```diff
@@ -2,19 +2,19 @@
...
+  releaseNotes = pkgs.writeTextDir "release-notes.md" ''
+    Automatic release on ${timestamp}
+  '';
+in pkgs.buildEnv {
+  name = "xrefcheck-release";
+  paths = [
+    releaseNotes
+    "${executable}/bin"
+  ];
+}
```
gromakovsky, line 20: Also I would suggest renaming the executable in release artifacts from `xrefcheck` to `xrefcheck-linux-x86_64-static-binary`. I can imagine that one day we may have other artifacts and `xrefcheck` may become ambiguous, so we will have to rename it to something longer and it will break `wget -O xrefcheck bla-bla && ./xrefcheck`. Moreover, this name is simply more descriptive, it makes it clear that this thing is a static binary for linux x86_64.

balsoft: I think long executable names (especially with `static-binary` in the end) aren't a very good idea. Would you be satisfied with `xrefcheck-x86_64-linux` ?

gromakovsky: Yes, I am fine with `xrefcheck-x86_64-linux`. It'd be nice if github releases had something like symlinks so that we could have both `xrefcheck-x86_64-linux` and `xrefcheck` without duplicating this file, but I guess there is no such feature :)

PR #44, opened 2020-06-26
Disable progress bar in CI automatically

### LINKED ISSUE #42: Invert --no-progress

By default `xrefcheck` reports progress by printing some nice progress bars. That's a cool feature when you run this tool on your computer but it's usually quite useless on CI. And CI often doesn't handle progress bars well, it doesn't update it in place, rather prints multiple bars (that depends on particular CI system of course).

Since according to our vision the primary use-case for `xrefcheck` is CI, it seems better to disable progress reporting by default. And replace `--no-progress` with `--progress` with the opposite semantics.

# Acceptance criteria

1. Progress reporting is disabled by default.
2. There is no `--no-progress`.
3. There is `--progress` that enables the existing progress reporting.

Martoon-00:
Problem: we have two major use cases:

1. Using xrefcheck in CI, in this case it is annoying for devops to
always pass mostly always reasonable `--no-progress` option.
2. Using it locally, here we want progress bar to be enabled by default
so that user could run `xrefcheck` without any options most of the time.

Solution:

Detect whether are we running in CI and use the respective default for
whether progress bar should be displayed.

Detecting whether we are within CI is possible using `CI` env variable,
at least Github actions, Gitlab CI and Buildkite pass this env.

`--no-progress` option still can be passed if user wants to suppress
progress bar in local run. The opposite `--progress` seems to also make
sense, because in case if high timeouts are set, user probably wants to
track progress somehow to make sure that CI didn't hang. So we add the
latter, and these two options can be used to override the default
behaviour defined by `CI` env var.

## Related issue(s)

- Fixed #42.

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,9 @@
 Unreleased
 ==========
 
+0.1.2
+=======
+
 * [#44](https://github.com/serokell/xrefcheck/pull/44)
   + Decide whether to show progress bar by default depending on `CI` env variable.
   + Added `--progress` option.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@
 Unreleased
 ==========
 
+* [#44](https://github.com/serokell/xrefcheck/pull/44)
+  + Decide whether to show progress bar by default depending on `CI` env variable.
+  + Added `--progress` option.
+
 0.1.1.2
 =======
 
--- a/package.yaml
+++ b/package.yaml
@@ -5,7 +5,7 @@
 spec-version: 0.31.0
 
 name:                xrefcheck
-version:             0.1.1.2
+version:             0.1.2
 github:              serokell/xrefcheck
 license:             MPL-2.0
 license-file:        LICENSE
--- a/src/Xrefcheck/CLI.hs
+++ b/src/Xrefcheck/CLI.hs
@@ -18,7 +18,7 @@ module Xrefcheck.CLI
     ) where
 
 import Data.Version (showVersion)
-import Options.Applicative (Parser, ReadM, command, eitherReader, execParser, fullDesc, help,
+import Options.Applicative (Parser, ReadM, command, eitherReader, execParser, flag', fullDesc, help,
                             helper, hsubparser, info, infoOption, long, metavar, option, progDesc,
                             short, strOption, switch, value)
 import Paths_xrefcheck (version)
@@ -50,7 +50,7 @@ data Options = Options
     , oRoot             :: FilePath
     , oMode             :: VerifyMode
     , oVerbose          :: Bool
-    , oShowProgressBar  :: Bool
+    , oShowProgressBar  :: Maybe Bool
     , oTraversalOptions :: TraversalOptions
     }
 
@@ -99,9 +99,16 @@ optionsParser = do
         short 'v' <>
         long "verbose" <>
         help "Report repository scan and verification details."
-    oShowProgressBar <- fmap not . switch $
-        long "no-progress" <>
-        help "Do not display progress bar during verification."
+    oShowProgressBar <- asum
+        [ flag' (Just True) $
+            long "progress" <>
+            help "Display progress bar during verification. \
+                 \This is enabled by default unless `CI` env var is set to true."
+        , flag' (Just False) $
+            long "no-progress" <>
+            help "Do not display progress bar during verification."
+        , pure Nothing
+        ]
     oTraversalOptions <- traversalOptionsParser
     return Options{..}
 
--- a/src/Xrefcheck/System.hs
+++ b/src/Xrefcheck/System.hs
@@ -5,13 +5,16 @@
 
 module Xrefcheck.System
     ( readingSystem
+    , askWithinCI
     , RelGlobPattern (..)
     , bindGlobPattern
     ) where
 
 import Data.Aeson (FromJSON (..), withText)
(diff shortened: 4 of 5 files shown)


src/Xrefcheck/System.hs
```diff
@@ -20,6 +23,14 @@ import qualified System.FilePath.Glob as Glob
 readingSystem :: IO a -> a
 readingSystem = unsafePerformIO
 
+-- | Heuristics to check whether we are running within CI.
+-- Check the respective env variable which is usually set in all CIs.
+withinCI :: Bool
+withinCI = case readingSystem $ lookupEnv "CI" of
```
gromakovsky, line 29: Do we really need to `unsafePerformIO` here? Why not `withinCI :: IO Bool`?

src/Xrefcheck/CLI.hs
```diff
@@ -99,9 +99,16 @@ optionsParser = do
...
-        help "Do not display progress bar during verification."
+    oShowProgressBar <- asum
+        [ flag' (Just True) $
+            long "progress" <>
+            help "Display progress bar during verification. \
+                 \This is enabled by default unless `CI` env var is set to true."
+        , flag' (Just False) $
+            long "no-progress" <>
+            help "Do not display progress bar during verification."
+        , pure Nothing
```
gromakovsky, line 110: What will happen if I pass both (maybe multiple times)? I guess it should be documented somewhere.

gromakovsky:
Also since we want to use this functionality immediately (it's needed for my github action), I suppose it makes sense to make a new release (hopefully #41 will be finished soon).

src/Xrefcheck/System.hs
Martoon-00: I honestly don't see much difference between these two. `IO` has too many capital letters and was too tiresome to write. One more minor reason if that if one day I need to use it in another place, it may be easier to use if it is pure.

src/Xrefcheck/CLI.hs
Martoon-00: optparse-applicative prompts an error if it sees an extra option, and the same applies for "passed option multiple times" or "pass contradicting options" cases. (the latter case seems to be a corner case of an extra option because once I parsed "--progress", I forget that "--no-progress" was ever expected).

src/Xrefcheck/System.hs
gromakovsky: I don't see a big difference between reading env variables and CLI parsing in this sense. You could as well read command line options or config file inside `unsafePerformIO`. Command line options don't change during program lifetime, but how often do you see them being read using `unsafePerformIO`? And I think `IO` is intentionally short to be easy and quick to type :roll_eyes:

package.yaml
```diff
@@ -5,7 +5,7 @@
 spec-version: 0.31.0
 
 name:                xrefcheck
-version:             0.1.1.2
+version:             0.1.1.3
```
gromakovsky, line 8: I am not sure but I think when you add some user-visible functionality you should update minor version number (i. e. the third one).

We don't explicitly follow PVP, but I'd use it as guidance: https://pvp.haskell.org/

It says:
> Non-breaking change. Otherwise, if only new bindings, types, classes, non-orphan instances or modules (but see below) were added to the interface, then A.B MAY remain the same but the new C MUST be greater than the old C. Note that modifying imports or depending on a newer version of another package may cause extra non-orphan instances to be exported and thus force a minor version change.

That's more applicable to libraries while our package is unlikely to be used as a library. But if we extend it to executables, I think the adding of a new command line option fits into this point and hence we should increase C.

src/Xrefcheck/CLI.hs
gromakovsky: Are `--progress` and `--no-progress` considered contradicting in this case?

src/Xrefcheck/System.hs
Martoon-00: Meh, okay, I have something to oppose to this.

src/Xrefcheck/CLI.hs
Martoon-00: Yes. If I pass `--progress --no-progress`, I get `Invalid option '--no-progress'`.

package.yaml
Martoon-00: Right. A bit surprising that we got so many releases (2 = "many" today) which do not change anything significant.

src/Xrefcheck/CLI.hs
gromakovsky: AFAIU more common behavior is to consider the last option, but I think it's not so important at this stage. For example, `stack build --test --no-test --no-test --test` is fine and runs tests, but `stack build --test --no-test --no-test --test --no-test` is also fine and it doesn't run tests.

CHANGES.md
```diff
@@ -7,6 +7,13 @@
 Unreleased
 ==========
 
+0.1.1.3
```
gromakovsky, line 10:
```suggestion
0.1.2
```

src/Xrefcheck/CLI.hs
Martoon-00: I will do this because it may be convenient, but some later, because once I rebased on top of lts-15 I realized that I cannot build without exceeding my hard drive memory.

PR #53, opened 2020-11-21
Ignoring global links in markdown files

alyoanton9:
## Description

Sometimes we want to ignore some global links in markdown files. The reasons can be different. For example, we know that CI runner does not have the certain level of access. Or we just want to have broken link in our markdown. This PR adds a couple of ways to do it.

**The first** is about including links we want to ignore in config file. It enables including not only links, but regular expressions for them.
Example:

If you want to ignore https://gitlab.com/morley-framework/morley/-/issues/new and https://gitlab.com/morley-framework/morley/-/merge_requests/new, you should include these lines in your config yaml file (`src-files/def-config.yaml` by default):
```
ignoreRefs:
    - https://gitlab.com/morley-framework/morley/-/(issues|merge_requests)/new
```
If you don't want any link in your repository to be ignored, leave `ignoreRefs` empty:

```
ignoreRefs:
   []
```

**The second** way to ignore a link is to annotate it right in-place. There are 3 modes of ignoring supported - link, paragraph and file. It becomes clear by example:
Example:
1. Use `<!-- xrefcheck: ignore file -->`  to ignore the whole file. It won't be scanned by parser. Note that it must be either at the top of markdown:
```
<!-- xrefcheck:ignore file -->

file content
```
or right after the comment blocks:
```
<!-- the comment -->
<!-- another comment -->
<!-- xrefcheck:ignore file -->

file content
```
otherwise the following error will be produced (`11:1-20` is a position where "ignore file" occurred):
`Error when scanning ./tests/markdowns/unexpected_ignore_file.md: "ignore file" must be at the top of markdown or right after comments at the top (11:1-20)`

2. Use `<!-- xrefcheck: ignore paragraph -->`  to ignore the following paragraph.

`web-site` and `team` links will be ignored, `blog` will be still validated:
```
<!-- xrefcheck:ignore paragraph  -->
Serokell [web-site](https://serokell.io/)
[team](https://serokell.io/team)

Serokell [blog](https://serokell.io/blog)
```
 If there is not a paragraph ignore comment is followed by, then it will be the respected error (`9:1-17` is a position of element that follows ignore comment):
`Error when scanning ./tests/markdowns/no_paragraph.md: expected a PARAGRAPH after "ignore paragraph", but found HEADING (9:1-17)`

3. Use ` <!-- xrefcheck: ignore link-->` to ignore the following link.

You can include a text before link, but there must not be line breaks.
So there are correct usages (ignore `web-site`, but not `team`)
```
<!-- xrefcheck: ignore link-->
Serokell [web-site](https://serokell.io/)
Serokell [team](https://serokell.io/team)
```
```
<!-- xrefcheck: ignore link-->
[web-site](https://serokell.io/)
[team](https://serokell.io/team)
```
and this one is not:
```
<!-- xrefcheck: ignore link-->
Serokell
[web-site](https://serokell.io/)
```
So the result will be (`8:1-10` is a position of "not-a-link" element):
`Error when scanning ./tests/markdowns/no_link.md: expected a LINK after "ignore link" (8:1-10)`

The description of ignoring modes is also located in the footer of help message:
```
To ignore a link in your markdown, include "<!-- xrefcheck: ignore mode -->"
comment with one of these modes:
  "link"                   Ignore the following link after the comment. If there
                           is no link, then throw an error.
  "paragraph"              Ignore the whole paragraph after the comment.
  "file"                   This mode can only be used at the top of markdown or
                           right after comments at the top.
```

## Related issue(s)
INT-135

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,79 @@ For description of other options:
 xrefcheck --help
 ```
 
+
+### Special functionality
+
+<details>
+  <summary>Ignoring external links</summary>
+
+  If you want some external links to not be verified, you can use one of the following ways to ignore those links:
+
+1. Add the regular expression that matches the ignoring link to the optional `ignoreRefs` parameter of your config file.
+
+    For example: 
+    ```yaml
+    ignoreRefs:
+      - https://bad.reference.(org|com)(/?)
+    ```
+    allows to ignore both `https://bad.reference.org` and `https://bad.reference.com` with or without last "/".
+
+2. Add right in-place annotation using one of the following ignoring modes (each mode is just a comment with a certain syntax).
+
+    * Ignore the link:
+
+        There are several ways to add this annotation:
+
+      * Just add it like a regular text before the ignoring link.
+
+        ```markdown
+        Bad ['com' reference](https://bad.reference.com) <!-- xrefcheck: ignore link --> and bad ['org' reference](https://bad.reference.org)
+        ```
+
+      * Separate the ignoring link from the annotation and the following text with single new lines.
+
+        ```markdown
+        Bad ['com' reference](https://bad.reference.com) and bad <!-- xrefcheck: ignore link -->
+        ['org'](https://bad.reference.org)
+        reference
+        ```
+
+        Therefore only `https://bad.reference.org` will be ignored.
+
+      * If the ignoring link is the first in a paragraph, then the annotation can also be added before a paragraph.
+
+        ```markdown
+        <!-- xrefcheck: ignore link -->
+        [Bad 'org' reference](https://bad.reference.org)
+        [Bad 'com' reference](https://bad.reference.com)
+        ```
+
+        It is still the same `https://bad.reference.org` will be ignored in this case.
+
+    * Ignore the paragraph:
+
+        ```markdown
+        <!-- xrefcheck: ignore paragraph -->
+        Bad ['org' reference](https://bad.reference.org)
+        Bad ['com' reference](https://bad.reference.com)
+
+        Bad ['io' reference](https://bad.reference.io)
+        ```
+
+        In this way, `https://bad.reference.org` and `https://bad.reference.com` will be ignored and `https://bad.reference.io` will still be verified.
+
+    * Ignore the whole file:
+        ```markdown
+        <!-- a comment -->
+        <!-- another comment -->
+
+        <!-- xrefcheck: ignore file -->
+        ...the rest of the file...
+        ```
+
+        Using this you can ignore the whole file.
+        </details>
+
 ## Configuring
 
 Configuration template (with all options explained) can be dumped with:
--- a/src-files/def-config.yaml
+++ b/src-files/def-config.yaml
@@ -13,6 +13,7 @@ traversal:
     # Stack files
     - .stack-work
 
+
 # Verification parameters.
 verification:
   # On 'anchor not found' error, how much similar anchors should be displayed as
--- a/src-files/def-config.yaml
+++ b/src-files/def-config.yaml
@@ -49,3 +49,9 @@ verification:
     - ../../issues/*
     - ../../merge_requests
     - ../../merge_requests/*
+
+  # POSIX extended regular expressions that match external references
+  # that have to be ignored (not verified).
+  # It is an optional parameter, so it can be omitted. 
+  ignoreRefs:
+    []
--- a/src/Xrefcheck/CLI.hs
+++ b/src/Xrefcheck/CLI.hs
@@ -17,12 +17,14 @@ module Xrefcheck.CLI
     , getCommand
     ) where
 
+import qualified Data.List as L
 import Data.Version (showVersion)
-import Options.Applicative (Parser, ReadM, command, eitherReader, execParser, flag', fullDesc, help,
-                            helper, hsubparser, info, infoOption, long, metavar, option, progDesc,
+import Options.Applicative (Parser, ReadM, command, eitherReader, execParser, flag', footerDoc, fullDesc,
+                            help, helper, hsubparser, info, infoOption, long, metavar, option, progDesc,
                             short, strOption, switch, value)
-import Paths_xrefcheck (version)
+import Options.Applicative.Help.Pretty (Doc, displayS, fill, fillSep, indent, renderPretty, text)
 
+import Paths_xrefcheck (version)
 import Xrefcheck.Config
 import Xrefcheck.Core
 
@@ -150,4 +152,32 @@ getCommand = do
         info (helper <*> versionOption <*> totalParser) $
         fullDesc <>
         progDesc "Cross-references verifier for markdown documentation in \
-                 \Git repositories."
+                 \Git repositories." <>
+        (footerDoc $ pure ignoreModesMsg)
+
+ignoreModesMsg :: Doc
+ignoreModesMsg = text $ header <> body
+    where
+        header = "To ignore a link in your markdown, \
+                 \include \"<!-- xrefcheck: ignore <mode> -->\"\n\
+                 \comment with one of these modes:\n"
+        body = displayS (renderPretty pageParam pageWidth doc) ""
+
+        pageWidth = 80
+        pageParam = 1
+
+        doc = fillSep $ map formatDesc modeDescr
+
+        modeDescr =
+            [ ("  \"link\"",      L.words $ "Ignore the link right after the comment.")
+            , ("  \"paragraph\"", L.words $ "Ignore the whole paragraph after the comment.")
+            , ("  \"file\"",      L.words $ "This mode can only be used at the top of markdown \
+                                            \or right after comments at the top.")
+            ]
+            
+        modeIndent = length ("\"paragraph\"" :: String) + 2
+        descrIndent = 27 - modeIndent
+
+        formatDesc (mode, descr) =
+            (fill modeIndent $ text mode) <>
+            (indent descrIndent $ fillSep $ map text descr)
--- a/src/Xrefcheck/Config.hs
+++ b/src/Xrefcheck/Config.hs
@@ -11,6 +11,8 @@ import Data.Aeson.Options (defaultOptions)
 import Data.Aeson.TH (deriveFromJSON)
 import Data.Yaml (FromJSON (..), decodeEither', prettyPrintParseException, withText)
 import Instances.TH.Lift ()
+import Text.Regex.TDFA (CompOption (..), ExecOption (..), Regex)
+import Text.Regex.TDFA.Text (compile)
 
 -- FIXME: Use </> from System.FilePath
 -- </> from Posix is used only because we cross-compile to Windows and \ doesn't work on Linux
@@ -40,6 +42,8 @@ data VerifyConfig = VerifyConfig
       -- ^ Files which we pretend do exist.
     , vcNotScanned                :: [FilePath]
       -- ^ Prefixes of files, references in which we should not analyze.
+    , vcIgnoreRefs                :: Maybe [Regex]
+      -- ^ Regular expressions that match external references we should not verify.
     }
 
 -----------------------------------------------------------
@@ -70,3 +74,26 @@ deriveFromJSON defaultOptions ''VerifyConfig
 instance KnownRatName unit => FromJSON (Time unit) where
     parseJSON = withText "time" $
         maybe (fail "Unknown time") pure . unitsP . toString
+
+instance FromJSON Regex where
+    parseJSON = withText "regex" $ \val -> do
+        let errOrRegex =
+                compile defaultCompOption defaultExecOption val
+        either (error . show) return errOrRegex
+
+-- Default boolean values according to
+-- https://hackage.haskell.org/package/regex-tdfa-1.3.1.0/docs/Text-Regex-TDFA.html#t:CompOption
+defaultCompOption :: CompOption
+defaultCompOption =
+    CompOption
+    { caseSensitive = True
+    , multiline = True
+    , rightAssoc = True
+    , newSyntax = True
+    , lastStarGreedy = False
+    }
+
+-- Default boolean value according to
+-- https://hackage.haskell.org/package/regex-tdfa-1.3.1.0/docs/Text-Regex-TDFA.html#t:ExecOption
+defaultExecOption :: ExecOption
+defaultExecOption = ExecOption {captureGroups = False}
--- a/src/Xrefcheck/Config.hs
+++ b/src/Xrefcheck/Config.hs
@@ -7,6 +7,7 @@
 
 module Xrefcheck.Config where
 
+import Control.Lens (makeLensesWith)
 import Data.Aeson.Options (defaultOptions)
 import Data.Aeson.TH (deriveFromJSON)
 import Data.Yaml (FromJSON (..), decodeEither', prettyPrintParseException, withText)
@@ -21,6 +22,7 @@ import Data.FileEmbed (embedFile)
 import Time (KnownRatName, Second, Time, unitsP)
 
 import Xrefcheck.System (RelGlobPattern)
+import Xrefcheck.Util (postfixFields)
 
 -- | Overall config.
 data Config = Config
@@ -45,6 +47,9 @@ data VerifyConfig = VerifyConfig
     , vcIgnoreRefs                :: Maybe [Regex]
       -- ^ Regular expressions that match external references we should not verify.
     }
+
+makeLensesWith postfixFields ''Config
+makeLensesWith postfixFields ''VerifyConfig
 
 -----------------------------------------------------------
 -- Default config
@@ -93,7 +98,6 @@ defaultCompOption =
     , lastStarGreedy = False
     }
 
--- Default boolean value according to
--- https://hackage.haskell.org/package/regex-tdfa-1.3.1.0/docs/Text-Regex-TDFA.html#t:ExecOption
+-- ExecOption value to improve speed
 defaultExecOption :: ExecOption
 defaultExecOption = ExecOption {captureGroups = False}
--- a/src/Xrefcheck/Scanners/Markdown.hs
+++ b/src/Xrefcheck/Scanners/Markdown.hs
@@ -10,6 +10,7 @@
 module Xrefcheck.Scanners.Markdown
     ( markdownScanner
     , markdownSupport
+    , parseFileInfo
     ) where
 
 import CMarkGFM (Node (..), NodeType (..), PosInfo (..), commonmarkToNode)
--- a/src/Xrefcheck/Scanners/Markdown.hs
+++ b/src/Xrefcheck/Scanners/Markdown.hs
@@ -14,12 +14,13 @@ module Xrefcheck.Scanners.Markdown
 
 import CMarkGFM (Node (..), NodeType (..), PosInfo (..), commonmarkToNode)
 import Control.Lens ((%=))
+import Control.Monad.Trans.Except (Except, runExcept, throwE)
 import qualified Data.ByteString.Lazy as BSL
+import Data.Char (isSpace)
 import Data.Default (Default (..))
 import qualified Data.Text as T
 import qualified Data.Text.Lazy as LT
 import Fmt (Buildable (..), blockListF, nameF, (+|), (|+))
-import GHC.Conc (par)
 
 import Xrefcheck.Core
 import Xrefcheck.Scan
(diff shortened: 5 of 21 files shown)


alyoanton9:
I faced one problem while implementing my features:

[The following build](https://buildkite.com/serokell/xrefcheck/builds/333#1875fe70-2579-46eb-9b73-178bdf6b71ce) fails because there are broken markdowns I use for tests (`./tests/markdowns/`) in `Test.Xrefcheck.IgnoranceSpec`. I tried to add `tests/markdowns/` (and even `tests/markdowns/unrecognised_option.md`, `tests/markdowns/unexpected_ignore_file.md` and so on) to `notScanned` of `src-files/def-config.yaml`. But it didn't work and AFAIU have never worked (I may be wrong). To check this I included `<!-- xrefcheck: ignore file -->` to `.github/pull_request_template.md` incorrectly (not at the top):
```
## Description

<!--
Describes the nature of your changes. If they are substantial, you should
further subdivide this into a section describing the problem you are solving and
another describing your solution.
-->

<!-- xrefcheck: ignore file -->
```
and got the respected error:
`Error when scanning ./.github/pull_request_template.md: "ignore file" must be at the top of markdown or right after the license (10:1-19)`
despite the fact there is
```
  notScanned:
    # GitHub-specific files
    - .github/pull_request_template.md
```
in `src-files/def-config.yaml`
So I suppose it can be a problem in `verifyRepo` of `Xrefcheck.Verify` (or somewhere else, or with me, who knows...)

alyoanton9:
I also have a couple of questions:

1. I decided to use `die` right in `markdownScanner`. So if we have broken markdown in our repo, then we get an error message and don't get other info about repo's links until we fix markdown.
But now I'm not sure it's correct approach. Perhaps it would be better to collect error messages about broken markdowns somehow and then print them with other broken links? However, this implies bigger code changes...

2. I also doubt we need  to support the following syntax `[[//]: # (xrefcheck: ignore link)`. It's trickier to parse this comment, they are not included in AST AFAIU (tested with `readFile "./someMD.md" >>= (putStr . (pretty :: Node -> String) . commonmarkToNode [] [])`). But I may be wrong again and I want to know your opinion (I just know it needs additional time and I'm not sure it worth).

3. Are tests for regex needed too (It's not difficult, just decided to clarify bcs all tests in this repo looks tiny and this feature is not volume)?

Martoon-00 (commented):
Hey, here is the first potion of comments from my side. 

I failed to properly allocate time for work at the beginning of the week, so managed to take a look only now. Gonna review further tomorrow / on Fri.

Martoon-00:
Btw I managed to look only at in-place annotations for now, and its implementation looks quite solid for me.

Martoon-00:
> I also doubt we need to support the following syntax [[//]: # (xrefcheck: ignore link). It's trickier to parse this comment, they are not included in AST AFAIU

Hm, this is indeed so, pretty sad. And I don't see an option not to ignore such comments, looks like they are just stripped as Markdown says to do.

I tried some other markdown parsers, without much success.

@gromakovsky what do you think about this?

> Are tests for regex needed too?

If possible, I would add one test since the proper match on regex occured to be non-trivial.

gromakovsky:
image

What is it? :roll_eyes:

gromakovsky:
> If you don't want any link in your repository to be ignored, leave `ignoreRefs` empty:

Can `ignoreRefs` be omitted?

gromakovsky:
> There are 3 modes of ignorance supported

Ignorance means "lack of knowledge or information". Shouldn't we call it "ignoring" or "ignore"?

gromakovsky:
> Use `<!-- xrefcheck: ignore file -->` to ignore the whole file.

I know we also have `notScanned` and `ignored` in the config which make this possibility look kinda redundant. Do we really need it? Is this `<!-- xrefcheck: ignore file -->` more powerful in any way?

gromakovsky:
> @gromakovsky what do you think about this?

Absence of built-in comments in markdown is sad true reality that we have to admit. If these :smart: commits are hard to parse, let's not support them, not many people are aware of this trick anyway.

alyoanton9:
> image
> 
> What is it? roll_eyes

I just were in doubt if it's okay to give [this link](https://issues.serokell.io/issue/INT-135) here (bcs it's from YT, not github).

alyoanton9:
> > There are 3 modes of ignorance supported
> 
> Ignorance means "lack of knowledge or information". Shouldn't we call it "ignoring" or "ignore"?

Actually we should, I'll do it.

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -49,54 +50,174 @@ nodeExtractText = mconcat . map extractText . nodeFlatten
...
+nodeExtractInfo (Node _ _ docNodes) =
+    if checkIgnoreFile docNodes
+    then return def
+    else finaliseFileInfo <$> extractionResult
+    where
+        extractionResult :: Except Text FileInfo
+        extractionResult =
+            let exceptRes = runExceptT $ loop docNodes Nothing
+                (eitherError, fileInfo) = runState exceptRes def
+            in except $ either Left (const $ Right fileInfo) eitherError
```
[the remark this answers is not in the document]
alyoanton9: > > If you don't want any link in your repository to be ignored, leave `ignoreRefs` empty:
> 
> Can `ignoreRefs` be omitted?

[Now](https://github.com/serokell/xrefcheck/pull/53/commits/d40f57a78ef63cf5bf65d552461c8a5b3eddb499) it can be

alyoanton9:
> > If you don't want any link in your repository to be ignored, leave `ignoreRefs` empty:
> 
> Can `ignoreRefs` be omitted?

[Now](https://github.com/serokell/xrefcheck/pull/53/commits/d40f57a78ef63cf5bf65d552461c8a5b3eddb499) it can be

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -49,54 +50,174 @@ nodeExtractText = mconcat . map extractText . nodeFlatten
...
 
 markdownScanner :: ScanAction
-markdownScanner path = liftIO $ do
-    res <- parseFileInfo path . decodeUtf8 <$> BSL.readFile path
-    force res `par` return res
+markdownScanner path = do
+    errOrInfo <- parseFileInfo . decodeUtf8 <$> BSL.readFile path
+    case errOrInfo of
+        Left errTxt -> do
+            die $ "Error when scanning " <> path <> ": " <> T.unpack errTxt
```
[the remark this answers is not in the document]
alyoanton9: I fixed everything you mentioned and I also added more safety (with `safeHead` function), you can look at it [here](https://github.com/serokell/xrefcheck/pull/53/commits/14f70d4f6bdffa0fb96cec9b16647f93cc39698a)

alyoanton9:
> > There are 3 modes of ignorance supported
> 
> Ignorance means "lack of knowledge or information". Shouldn't we call it "ignoring" or "ignore"?

Done (later will

> > Use `<!-- xrefcheck: ignore file -->` to ignore the whole file.
> 
> I know we also have `notScanned` and `ignored` in the config which make this possibility look kinda redundant. Do we really need it? Is this `<!-- xrefcheck: ignore file -->` more powerful in any way?

To my way of thinking, it's better to have a single way/interface/tool to ignore smth, either a part of the markdown or the whole markdown file. Therefore, choosing between these two, I would prefer `<!-- xrefcheck: ignore file -->`.
However, as I stated at PR (and I hoped someone answer me :disappointed:), I don't see the way not to scan markdowns at the test directory (checks at github fail because of this). I tried to use `notScanned` but failed. Please, read [this comment](https://github.com/serokell/xrefcheck/pull/53#issuecomment-731626025) since I need a possible solution.

Martoon-00:
Regarding the issue with tests, note that there are separate `notScanned` and `ignored` properties. The former means that the file is parsed and its content is accounted when it is referenced from other files, but links in such files are not checked. And `ignored` means that the file is never read and should not be referenced from outside. So probably you wanted to put your tests into `ignored`.

If you think that this behavior was not clear from the comments in config, or one needs to read them multiple times to come to what I wrote above (and I willingly believe this is so), please update those comments to clarify things (in a separate pull request?). Or I can try to do this myself and request your review.

Martoon-00:
> Can ignoreRefs be omitted?

I suspect that no, that the same holds for other properties. And I think this is not so bad, because it is easier to have all options with comments at hand rather than grep for the documentation when something goes wrong (e.g. need to adjust timeouts). Though it's not for me, as the library author, to enforce, let's handle this separately if you think fields should be optional.

Martoon-00:
> Do we really need it? Is this  more powerful in any way?

It seems to have its own use cases. If a file is moved often or is autogenerated, it may be simpler to have the disabling comment embedded.

On the other hand, gitignore never proposes such practices, so maybe this is not justified :thinking:

src/Xrefcheck/Verify.hs
```diff
@@ -259,11 +259,17 @@ checkExternalResource VerifyConfig{..} link
...
-        in null b && null a && not (null m)) vcIgnoreRefs
+    isIgnored =
+        let mIsIgnored = (doesMatchAnyRegex link) <$> vcIgnoreRefs
+        in maybe True id mIsIgnored
+
     doesReferLocalhost = any (`T.isInfixOf` link) ["://localhost", "://127.0.0.1"]
 
+    doesMatchAnyRegex :: Text -> ([Text] -> Bool)
+    doesMatchAnyRegex txt = any $ \t ->
+        let (before, match, after) = txt =~ t :: (Text, Text, Text)
```
Martoon-00, line 270: Some more comments:

1. I heard from Java world that compiling regex pattern may be an expensive operation, it is better to do once and then apply multiple times.
2. Regex compilation can fail.

Taking these two into account, maybe let's store not `Text`s in `vsIngoreRefs` rather [`Regex`](https://hackage.haskell.org/package/regex-tdfa-1.3.1.0/docs/Text-Regex-TDFA-Text.html#t:Regex)s, and implement a JSON instance that compiles the pattern, reporting an appropriate error if any occurs? Failing fast sounds like a good thing.

The `Config.hs` module even has an example of defining `FromJSON` instance manually, the implementation for `Regex`  should look pretty similar.

src/Xrefcheck/Scanners/Markdown.hs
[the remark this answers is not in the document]
Martoon-00: Thanks, looks much better to me :+1:

>  I would like to create another special issue.

Sounds good. Post it here once you create it.

src/Xrefcheck/CLI.hs
```diff
@@ -150,4 +152,32 @@ getCommand = do
...
         progDesc "Cross-references verifier for markdown documentation in \
-                 \Git repositories."
+                 \Git repositories." <>
+        (footerDoc $ pure ignoreModesMsg)
+
+ignoreModesMsg :: Doc
+ignoreModesMsg = text $ header <> body
+    where
+        header = "To ignore a link in your markdown, \
+                 \include \"<!-- xrefcheck: ignore mode -->\"\n\
```
Martoon-00, line 162: I would wrap "mode" into something, braces or angle brackets, to make it clear that only this word it to be replaced with actual mode.

src/Xrefcheck/CLI.hs
```diff
@@ -150,4 +152,32 @@ getCommand = do
...
+        modeDescr =
+            [ ("  \"link\"",      L.words $ "Ignore the following link after the comment. \
+                                            \If there is no link, then throw an error.")
+            , ("  \"paragraph\"", L.words $ "Ignore the whole paragraph after the comment.")
+            , ("  \"file\"",      L.words $ "This mode can only be used at the top of markdown \
+                                            \or right after the license. Otherwise throw an error.")
+            ]
+            
+        modeIndent = length ("\"paragraph\"" :: String) + 2
+        descrIndent = 27 - modeIndent
```
Martoon-00, line 179: Mda, nice hacks. But it looks neat, so it's worth it.

src/Xrefcheck/CLI.hs
```diff
@@ -150,4 +152,32 @@ getCommand = do
...
+        pageWidth = 80
+        pageParam = 1
+        doc = fillSep $ map formatDesc modeDescr
+
+        modeDescr =
+            [ ("  \"link\"",      L.words $ "Ignore the following link after the comment. \
+                                            \If there is no link, then throw an error.")
+            , ("  \"paragraph\"", L.words $ "Ignore the whole paragraph after the comment.")
+            , ("  \"file\"",      L.words $ "This mode can only be used at the top of markdown \
+                                            \or right after the license. Otherwise throw an error.")
```
Martoon-00, line 175: I would not mention throwing error, this seems not important for me since we said having an "ignore" command at the top is a must.

Also, the practice of using licenses is in fact not wide-spread at all and not every reader will understand why we mention them, I would rather say something like "or right after comments at the top, for instance in case the file has to start from a license note."

tests/Test/Xrefcheck/IgnoringSpec.hs
```diff
@@ -0,0 +1,50 @@
...
+            , "tests/markdowns/no_paragraph.md"
+            , "tests/markdowns/unexpected_ignore_file.md"
+            , "tests/markdowns/unrecognised_option.md"
+            ]
+        
+        defFI = FileInfo [] []
+
+        parse path = parseFileInfo . decodeUtf8 <$> BSL.readFile path
+        
+        getFI path = fromRight defFI <$> parse path
```
Martoon-00, line 44: It's better to just fail in such cases in tests, otherwise, if this error hits, investigating what exactly is going on may take a while :)  And a failure is quite unexpected here, so we can afford failing.

tests/Test/Xrefcheck/IgnoringSpec.hs
```diff
@@ -0,0 +1,50 @@
...
+
+import Xrefcheck.Core
+import Xrefcheck.Scanners.Markdown
+
+spec :: Spec
+spec = do
+    describe "Parsing failures" $ do
+        it "Check if parsing incorrect markdowns produce exceptions" $ do
+            areIncorrect <- mapM isIncorrectMD failPaths
+            any id areIncorrect `shouldBe` True
```
Martoon-00, line 19: `any id` = `or`

tests/Test/Xrefcheck/IgnoringSpec.hs
```diff
@@ -0,0 +1,50 @@
...
+
+        parse path = parseFileInfo . decodeUtf8 <$> BSL.readFile path
+        
+        getFI path = fromRight defFI <$> parse path
+        
+        getRefs fi = map rName $ fi ^. fiReferences 
+
+        isIncorrectMD path = do
+            errOrInfo <- parse path
+            return $ isLeft errOrInfo
```
Martoon-00, line 50: :+1:

Martoon-00: I would like to see the following test: if a file contains "ignore file" command, this does not affect the list of anchors found in it (so if another file references an anchor in ignored file, this still works). 

I now understand that probably my initial suggestion to just erase the entire file on "ignore file" seems wrong in this regard. Indeed, it's debatable whether do we want to exclude file entirely or just skip checking links in it - I think ideally we could support both somehow, but if not, I think it's better to have the latter behavior - a file marked as ignored should still be referenced outside.

(though first let's complete the discussion about the necessity of `exclude file` just in case)

alyoanton9:
> Regarding the issue with tests, note that there are separate `notScanned` and `ignored` properties. The former means that the file is parsed and its content is accounted when it is referenced from other files, but links in such files are not checked. And `ignored` means that the file is never read and should not be referenced from outside. So probably you wanted to put your tests into `ignored`.
> 
> If you think that this behavior was not clear from the comments in config, or one needs to read them multiple times to come to what I wrote above (and I willingly believe this is so), please update those comments to clarify things (in a separate pull request?). Or I can try to do this myself and request your review.

Oh, I missed this "ignored" actually. Perhaps because it is inside of "traversal"...Nevertheless I think it's my mistake. Actually I would like to mention that files from "notScanned"  are still parsed. But it's a minor change. May I add this at the current PR?

tests/Test/Xrefcheck/IgnoringSpec.hs
Martoon-00: Well okay, as we discussed separately, with such an approach `ignore paragraph`  would also have to be updated. That's all quite tricky to do and motivation for this is unclear. We will create a separate ticket for implementing commands in `notScanned` way.

alyoanton9:
> > Can ignoreRefs be omitted?
> 
> I suspect that no, that the same holds for other properties. And I think this is not so bad, because it is easier to have all options with comments at hand rather than grep for the documentation when something goes wrong (e.g. need to adjust timeouts). Though it's not for me, as the library author, to enforce, let's handle this separately if you think fields should be optional.

@gromakovsky Additionally with comment above we also decided it would be strange to have all options obligatory but optional `ignoreRefs`. Perhaps, it doesn't worth it, do it?

gromakovsky:
> I just were in doubt if it's okay to give this link here (bcs it's from YT, not github).

Maybe it's not okay to give that link, but it would definitely be better to write INT-135. 135 makes one think it's #135 on github, which currently doesn't exist (so one may be confused), but may be created later (and then someone who opens this PR in a couple of years will be even more confused).

gromakovsky:
> I suspect that no, that the same holds for other properties. And I think this is not so bad, because it is easier to have all options with comments at hand rather than grep for the documentation when something goes wrong (e.g. need to adjust timeouts). Though it's not for me, as the library author, to enforce, let's handle this separately if you think fields should be optional.

I think optional fields are better at least because they can be added in backwards-compatible way. If someone uses an older config without `ignoreRefs`, we still want it to be parsed, I guess.

gromakovsky:
> Additionally with comment above we also decided it would be strange to have all options obligatory but optional ignoreRefs. Perhaps, it doesn't worth it, do it?

As I said, I think optional fields are good from the backwards-compatibility perspective. Even if we say "xrefcheck doesn't have many users as of now, so we can break things for people when they update xrefcheck version" (it still doesn't sound good btw), at some point we hope to have more users and this argument should stop being appropriate. So I think the sooner we start caring about it, the better.

Regarding `xrefcheck:ignore file`, I think I don't mind having this feature, it's up to you. However, the logic with ignoring licensing information sounds weird to me. This licensing headers are used by us, but I think 99.99% developers don't follow these REUSE practices. And I think there are various other tools which have a certain convention about adding some headers. Favoring license headers over other possible headers doesn't sound justified to me.
I propose to relax this rule and check all comments before the first non-comment text occurrence. So if there is
```
<!-- foo bar baz -->
<!-- my-editor:some-setting:
  - q
  - w
-->
<!-- xrefcheck:ignore file -->
```

we still ignore the file.

gromakovsky:
Also can you please update the PR title so that it looks like a title written by a human rather than an auto-generated string from the branch name?

Martoon-00:
> I think optional fields are better at least because they can be added in backwards-compatible way. If someone uses an older config without ignoreRefs, we still want it to be parsed, I guess.

On the other hand, updating config seems not that hard, and all options will be in place.

Though I suspect that it will be not me who will propagate this to most of our repositories, and looks like we have some external users who may don't like that compatibility breaks, so okay, let's make it optional. Though it still should be mentioned in the default config (`ignoredRefs: []`) with a comment.

alyoanton9:
@Martoon-00 @gromakovsky I suppose I've fixed all according to your reviews
I assume we need to inform users about adding these ways to ignore global links. Need I to update `README`? And If I need then where(in what paragraph) to put this information?

gromakovsky:
First of all, you should update the changelog file as stated in the PR template. Besides, since this functionality can be hard to discover for a random user, I think it makes sense to mention it in the "Usage" chapter of README.

gromakovsky:
I didn't review code closely because I am not really familiar with it (and not sure I want to learn it closely, I heard it's not of the highest quality :trollface:), I believe @Martoon-00's review should be sufficient.

Judging by descriptions, comments, tests, examples and all such stuff it looks good to me now. I'll approve after you update README and CHANGES.

gromakovsky:
So apparently now only this is left:
> Besides, since this functionality can be hard to discover for a random user, I think it makes sense to mention it in the "Usage" chapter of README.

alyoanton9:
> So apparently now only this is left:
> 
> > Besides, since this functionality can be hard to discover for a random user, I think it makes sense to mention it in the "Usage" chapter of README.

Unfortunately, there are also tests for regex are left, I'm implementing them (I faced with some problems, but I'll finish today).

alyoanton9:
@gromakovsky I've changed `README`
@Martoon-00 Could you please review the last fixes and regex tests ([since Dec 16](https://github.com/serokell/xrefcheck/pull/53/commits)) and also look at `README`?

src/Xrefcheck/Config.hs
```diff
@@ -72,3 +74,26 @@ deriveFromJSON defaultOptions ''VerifyConfig
...
+    , multiline = True
+    , rightAssoc = True
+    , newSyntax = True
+    , lastStarGreedy = False
+    }
+
+-- Default boolean value according to
+-- https://hackage.haskell.org/package/regex-tdfa-1.3.1.0/docs/Text-Regex-TDFA.html#t:ExecOption
+defaultExecOption :: ExecOption
+defaultExecOption = ExecOption {captureGroups = False}
```
Martoon-00, line 99: Do I understand correctly that `False` is not the default value provided by regex-tdfa as mentioned in the comment above, but we pass it to speed up things (which seems reasonable)?

src/Xrefcheck/Config.hs
```diff
@@ -72,3 +74,26 @@ deriveFromJSON defaultOptions ''VerifyConfig
 instance KnownRatName unit => FromJSON (Time unit) where
     parseJSON = withText "time" $
         maybe (fail "Unknown time") pure . unitsP . toString
+
+instance FromJSON Regex where
+    parseJSON = withText "regex" $ \val -> do
+        let errOrRegex =
+                compile defaultCompOption defaultExecOption val
+        either (error . show) return errOrRegex
```
Martoon-00, line 82: :+1:

src/Xrefcheck/Verify.hs
```diff
@@ -260,15 +260,16 @@ checkExternalResource VerifyConfig{..} link
...
     doesReferLocalhost = any (`T.isInfixOf` link) ["://localhost", "://127.0.0.1"]
 
-    doesMatchAnyRegex :: Text -> ([Text] -> Bool)
-    doesMatchAnyRegex txt = any $ \t ->
-        let (before, match, after) = txt =~ t :: (Text, Text, Text)
-        in null before && null after && not (null match)
+    doesMatchAnyRegex :: Text -> ([Regex] -> Bool)
+    doesMatchAnyRegex src = any $ \regex ->
+        let rightResult = fromRight Nothing $ regexec regex src
+        in case rightResult of
```
Martoon-00, line 270: Why not pattern match on `Either` directly?

src-files/def-config.yaml
```diff
@@ -53,5 +53,7 @@ verification:
     - ../../merge_requests
     - ../../merge_requests/*
 
+  # Regular expressions that match external references that have to be ignored.
```
Martoon-00: I would also mention which regex syntax do we support (looks like it is POSIX in our case)

README.md
```diff
@@ -56,6 +56,51 @@ We provide the following ways for you to use xrefcheck:
...
+
+* Ignore the paragraph:
+
+```sh
+<!-- xrefcheck: ignore paragraph -->
+[Bad 'ru' reference](https://bad.reference.ru)
+[Bad 'com' reference](https://bad.reference.com)
+
+[Bad 'io' reference](https://bad.reference.io)
+```
```
Martoon-00, line 89: I think it's also wroth including some text, not bare links, otherwise the reader may think that ignorance works only on standalone links.

README.md
```diff
@@ -56,6 +56,51 @@ We provide the following ways for you to use xrefcheck:
 
 If none of those are suitable for you, please open an issue!
 
+If you want some external links to not be verified, you can use one of the following ways to ignore that links:
+1. Add the regular expression that match ignoring link to optional `ignoreRefs` parameter of your config file.
+
+For example: 
+```sh
+ignoreRefs:
+  - https://bad.reference.(ru|com)(/?)
```
Martoon-00, line 65: Some small preference: let's not mention `ru`, I feel uneasy about this :smile_cat: 

What about `org`?

README.md
```diff
@@ -56,6 +56,51 @@ We provide the following ways for you to use xrefcheck:
...
+* Ignore the whole file:
+```sh
+<!-- a comment -->
+<!-- another comment -->
+
+<!-- xrefcheck: ignore file -->
+...the rest of the file...
+```
+
+Using this you are able to ignore the whole file besides  comments before `<!-- xrefcheck: ignore file -->`.
```
Martoon-00, line 102: Comments cannot contain links, I think we could mention not the "besides comments before" part.

README.md
```diff
@@ -56,6 +56,51 @@ We provide the following ways for you to use xrefcheck:
...
+
 To find all broken links in a repository, run from within its folder:
```
Martoon-00, line 104: Is overall this looks good, nice that we have this documented. 

Usually, we apply some spell-checker to public documents, please use some checker to proof-read the document (e.g. grammarly.com)

README.md
```diff
@@ -56,6 +56,51 @@ We provide the following ways for you to use xrefcheck:
 
 If none of those are suitable for you, please open an issue!
 
+If you want some external links to not be verified, you can use one of the following ways to ignore that links:
```
Martoon-00, line 59: To avoid making README too big, let's put this into a separate section and hide under a spoiler, something like this:

```markdown
### Special functionality

<details>
  <summary>Ignoring external links</summary>
  Text goes here
</details>
```

README.md
```diff
@@ -56,6 +56,51 @@ We provide the following ways for you to use xrefcheck:
...
+ignoreRefs:
+  - https://bad.reference.(ru|com)(/?)
+```
+allows to ignore both `https://bad.reference.ru` and `https://bad.reference.com` with or without last "/".
+
+2. Add right in-place annotation using one of the following ignoring mode (each mode is just a comment with a certain syntax).
+
+* Ignore the link:
+
+```sh
```
Martoon-00, line 73: Should be `markdown`.

README.md
```diff
@@ -56,6 +56,51 @@ We provide the following ways for you to use xrefcheck:
 
 If none of those are suitable for you, please open an issue!
 
+If you want some external links to not be verified, you can use one of the following ways to ignore that links:
+1. Add the regular expression that match ignoring link to optional `ignoreRefs` parameter of your config file.
+
+For example: 
+```sh
```
Martoon-00, line 63: Should be `yaml`

tests/Test/Xrefcheck/IgnoreRegexSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+                    ]
+
+            let notMatchedLinks =
+                    [ "https://non-existent.reference/"
+                    , "https://bad.externall.reference"
+                    , "https://bad.reference.io"
+                    ]
+
+            (matchedLinks `isInfixOf` brokenLinks) `shouldBe` False
+            (notMatchedLinks `isInfixOf` brokenLinks) `shouldBe` True
```
Martoon-00, line 55: What do you want to test here?

tests/configs/test-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - ../../issues
+    - ../../issues/*
+    - ../../merge_requests
+    - ../../merge_requests/*
+
+  # Regular expressions that match external references that have to be ignored.
+  # It is an optional parameter, so it can be omitted.
+  ignoreRefs:
+    - https://bad.((external.)?)reference(/?)
+    - https://bad.reference.(ru|com)
```
Martoon-00, line 60: It is better to avoid copy-pasting the default config and doing small changes - just for the sake of tests, because when we later add more test configs and more options, maintaining all the config files would become extremely tiresome.

I see 3 options for how to improve this:
1. Remove comments and all the unnecessary entries and hope that one day we make all the fields here optional.
2. Initialize the necessary part of config in Haskell, e.g. `config = defConfig{ vcIgnoreRefs = Aeson.decode "<config text>" }`. You may need to add lenses for config to make it easier to update fields.
3. An ideal option - like 2, but do `config = defConfig{ vsIgnoreRefs = ["regex1", "regex2"] }`, and add tests on JSON decoding of regex separately.

Martoon-00 (commented):
Sorry, this time I brought an awful lot of comments.

It's fine if you don't manage to fix them all before your vacation, maybe I pick finishing this PR later if we need it urgently and you don't object.

src/Xrefcheck/Config.hs
alyoanton9: Oh, you are right. I've just "copied" this comment from the previous default value.

gromakovsky (approved):
I don't have anything else to add to what @Martoon-00 wrote.

tests/Test/Xrefcheck/IgnoreRegexSpec.hs
alyoanton9: Due to the fact all links for these tests are invalid, they are either reported as invalid or ignored by the verifier.

```(matchedLinks `isInfixOf` brokenLinks) `shouldBe` False``` checks that links matched with regexs from `test-config.yaml` are not in broken links (not verified)

And ```(notMatchedLinks `isInfixOf` brokenLinks) `shouldBe` True``` checks that links that are not matched with regexs from `test-config.yaml` are contained in broken links (verified)

**Comment by alyoanton9** [comment]
@Martoon-00 I've [added lenses](https://github.com/serokell/xrefcheck/pull/53/commits/9e48fcb2a00198bf0b3c350d97b5f0651636ce0c) for `Config` and related data types. But now they are used only for my tests (I didn't even renamed the rest of the fields). There are a lot of usages and I suppose It should be a separated PR like "use lenses everywhere".

File: tests/Test/Xrefcheck/IgnoreRegexSpec.hs (hunk 21 above)
Comment (Martoon-00) [reply tests/Test/Xrefcheck/IgnoreRegexSpec.hs:62]: As far as I understand:

* If `brokenLinks = [a, b, c]` and `matchedLinks = [a, d]`, then `isInfixOf` will return `False` and test will pass, despite presence of `a` should cause a failure. 
* If `brokenLinks = [a, b, c]` and `notMatchedLinks = [a, c]`, `isInfixOf` will return `False` because the latter is a subsequence of the former, but not a sublist, and the test with fail despite everything is good.

Actually, I would be a bit more verbose here and write something like

```haskell
forM links $ \link -> do
  when (link `elem` links0) $
    assertFailure $ "Link is considered broken despite it should be ignored: " <> show link
```

because having relatively proper error messages in tests is also valuable :slightly_smiling_face:.

Martoon-00:
> @Martoon-00 I've [added lenses](https://github.com/serokell/xrefcheck/pull/53/commits/9e48fcb2a00198bf0b3c350d97b5f0651636ce0c) for `Config` and related data types. But now they are used only for my tests (I didn't even renamed the rest of the fields). There are a lot of usages and I suppose It should be a separated PR like "use lenses everywhere".

In fact, I personally a bit dislike the practice of prefixing field records with `_` and treating lenses as a primary way to work with fields. One of the reasons is exactly that to be consistent we now have to use this practice everywhere across the project, even for datatypes where using lenses does not make much benefit.

Often we declare

```
-- | For datatype with "myNyan" field it will create "myNyanL" lens.
postfixLFields :: LensRules
postfixLFields = lensRules & lensField .~ mappingNamer (\s -> [s++"L"])
```

and use it to derive lenses via `makeLensesWith` instead of default `makeLenses`, then the existing use of field accessors as getters (which is often neater) remains justified, and lenses can be used only for updates and access in nested structures.

What do you think? If we do so, then there will be no need to update anything in the remainder of the project.

alyoanton9:
> > @Martoon-00 I've [added lenses](https://github.com/serokell/xrefcheck/pull/53/commits/9e48fcb2a00198bf0b3c350d97b5f0651636ce0c) for `Config` and related data types. But now they are used only for my tests (I didn't even renamed the rest of the fields). There are a lot of usages and I suppose It should be a separated PR like "use lenses everywhere".
> 
> In fact, I personally a bit dislike the practice of prefixing field records with `_` and treating lenses as a primary way to work with fields. One of the reasons is exactly that to be consistent we now have to use this practice everywhere across the project, even for datatypes where using lenses does not make much benefit.
> 
> Often we declare
> 
> ```
> -- | For datatype with "myNyan" field it will create "myNyanL" lens.
> postfixLFields :: LensRules
> postfixLFields = lensRules & lensField .~ mappingNamer (\s -> [s++"L"])
> ```
> 
> and use it to derive lenses via `makeLensesWith` instead of default `makeLenses`, then the existing use of field accessors as getters (which is often neater) remains justified, and lenses can be used only for updates and access in nested structures.
> 
> What do you think? If we do so, then there will be no need to update anything in the remainder of the project.

I didn't know this approach. It seems to be more reasonable, I'll do it!

Martoon-00:
Okay, I think we reached the stage where this work should be tested against real projects :tada:.

Though sad news is that somehow all external links now seem to be not scanned :thinking: 

So I tried [morley](https://gitlab.com/morley-framework/morley/), [indigo](https://gitlab.com/morley-framework/indigo/) and [tzip](https://gitlab.com/tzip/tzip/) repositories and when I e.g. do

```sh
stack exec xrefcheck -- -r ../morley -m external-only
``` 
from xrefcheck's master I get some failing external links, however in this branch I don't.

You are welcome to do some debugging, or maybe I can investigate this in January myself, I believe there should be an easy fix.

For now, please squash commits and let's assume that everything is mostly good.

alyoanton9:
> Okay, I think we reached the stage where this work should be tested against real projects tada.
> 
> Though sad news is that somehow all external links now seem to be not scanned thinking
> 
> So I tried [morley](https://gitlab.com/morley-framework/morley/), [indigo](https://gitlab.com/morley-framework/indigo/) and [tzip](https://gitlab.com/tzip/tzip/) repositories and when I e.g. do
> 
> ```shell
> stack exec xrefcheck -- -r ../morley -m external-only
> ```
> 
> from xrefcheck's master I get some failing external links, however in this branch I don't.
> 
> You are welcome to do some debugging, or maybe I can investigate this in January myself, I believe there should be an easy fix.
> 
> For now, please squash commits and let's assume that everything is mostly good.

Oh, I checked it on `morley` at the earlier stages of "developing" and everything was fine...
I'll try to do smth now, but if I fail then I'll finish it at the beginning of January

Martoon-00:
I'm convincing @alyoanton9 to offload finishing this PR to me :)

Should be relatively easy to complete, and finally after almost an year we will be able to keep xrefcheck CI jobs passing.

alyoanton9:
> 
> 
> Okay, I think we reached the stage where this work should be tested against real projects 🎉.
> 
> Though sad news is that somehow all external links now seem to be not scanned 🤔
> 
> So I tried [morley](https://gitlab.com/morley-framework/morley/), [indigo](https://gitlab.com/morley-framework/indigo/) and [tzip](https://gitlab.com/tzip/tzip/) repositories and when I e.g. do
> 
> ```shell
> stack exec xrefcheck -- -r ../morley -m external-only
> ```
> 
> from xrefcheck's master I get some failing external links, however in this branch I don't.
> 
> You are welcome to do some debugging, or maybe I can investigate this in January myself, I believe there should be an easy fix.
> 
> For now, please squash commits and let's assume that everything is mostly good.

I tested this work on `morley` and `indigo` and got the same results as with a master branch, that is, the same external links were reported as broken.

The only problem is about instability, e.g. in one time there are 2 definitely broken links and after a minute there are a lot of links having errors like "Too many requests" or "Response timeout". But I suppose it depends on internet connection, not xrefcheck.

@Martoon-00 Could you please check it one more time? It's strange.

src-files/def-config.yaml
```diff
@@ -13,6 +13,9 @@ traversal:
     # Stack files
     - .stack-work
 
+    # Test files
+    - tests/markdowns
+
```
Martoon-00, line 18: Btw, this should not be here as this config is what `xrefcheck dump-config` produces. Consider updating this field from code if necessary.

Martoon-00:
Hmm, really, somehow everything works for me now, though I'm quite sure I checked everything properly back then.
Okay, sorry for the confusion, let's proceed with finishing this. "Too many requests" is indeed a different issue.

src/Xrefcheck/Verify.hs
```diff
@@ -251,14 +252,26 @@ checkExternalResource :: VerifyConfig
...
+    | isIgnored = return mempty
     | doesReferLocalhost = return mempty
     | otherwise = fmap toVerifyRes $ do
         makeRequest HEAD 0.3 >>= \case
             Right () -> return $ Right ()
             Left   _ -> makeRequest GET 0.7
   where
+    isIgnored =
+        let maybeIsIgnored = (doesMatchAnyRegex link) <$> vcIgnoreRefs
+        in fromMaybe True maybeIsIgnored
```
Martoon-00, line 264: Ah okay, back to that issue: I get all broken external links unnoticed if `ignoreRefs` field is absent. Apparently this code is not exactly what we want.

Martoon-00:
Does `<!-- xrefcheck: ignore link -->` work for you if it is in the middle of a paragraph? I'm trying to make [this link](https://gitlab.com/morley-framework/morley/-/blob/913ff3c25ddf9688ac140ce218ac17c9bb5f8753/CONTRIBUTING.md#L11) ignored, but this does not seem to work, Wonder whether does it happen because our markdown parser does not parse this comment as a separate node. If so, it's worth thinking about what to do regarding this.

alyoanton9:
> 
> 
> Does `<!-- xrefcheck: ignore link -->` work for you if it is in the middle of a paragraph? I'm trying to make [this link](https://gitlab.com/morley-framework/morley/-/blob/913ff3c25ddf9688ac140ce218ac17c9bb5f8753/CONTRIBUTING.md#L11) ignored, but this does not seem to work, Wonder whether does it happen because our markdown parser does not parse this comment as a separate node. If so, it's worth thinking about what to do regarding this.

What do you mean by "in the middle of the paragraph"?
```
<!-- xrefcheck: ignore link -->
Please [open an issue](https://gitlab.com/morley-framework/morley/-/issues/new)
if you find a bug or have a feature request.
Before submitting a bug report or feature request, check to make sure it hasn't already been submitted.
```
^ works fine, the link is ignored
We have `<!--xrefcheck: ignore link-->` as separate paragraph, that is, `<!-- xrefcheck: ignore *-->` is always followed by the newline

src-files/def-config.yaml
alyoanton9: Oh
AFAIU we need `tests/markdowns` folder to be at `ignored` when we execute `xrefcheck` on itself, especially on `xrefcheck`'s root or `tests` (execution on another directory of `xrefcheck` must work fine). But how can we define it properly? Can we just check if the root's last directories are `xrefcheck` or `xrefcheck/tests`?

src/Xrefcheck/Verify.hs
alyoanton9: Fixed

src-files/def-config.yaml
alyoanton9: [Updating from code](https://github.com/serokell/xrefcheck/pull/53/commits/13e551dc48ce602082f778ded55781aed4fb47ed) ✔️

Martoon-00:
> What do you mean by "in the middle of the paragraph"?

I saw it like this:

```
Please <!-- xrefcheck: ignore link --> [open an issue](https://gitlab.com/morley-framework/morley/-/issues/new)
if you find a bug or have a feature request.
```

so that it is possible to ignore one link within a paragraph and leave the other. Though in this case the need to put an `ignore link` annotation right into the text is indeed a bit sad.

Currently, if a paragraph contains two links and I want to ignore the second one, how can I express that?

src-files/def-config.yaml
Martoon-00: While this assumption that we appear in `xrefcheck/test` folder only in tests seems plausible, I think that `Main` on itself should not know anything about test-related stuff, because if we start doing so for other similar demands of test code, then the production logic will eventually become a sort of mess.

Normally the necessary behavior should be configurable and updated respectively in test logic. In our case this can look as follows:
* Extract a piece of logic from `defaultAction`, so that this piece accepts `Config` and performs the necessary checks.
* Invoke this method in tests, passing a config constructed manually.

src/Xrefcheck/Scanners/Markdown.hs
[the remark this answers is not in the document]
Martoon-00: So now there is #56.

src-files/def-config.yaml
alyoanton9: We discussed it separately that `tests/markdowns` should be ignored through github CLI options, not from `Main`

alyoanton9:
There wasn't an ability to ignore a link in the middle of a paragraph, as it was observed. So, we discussed with @Martoon-00 separately what the ways of rendering could be.
As a result, I added a couple of commits to handle this option for now.
There are [ignoring itself](https://github.com/serokell/xrefcheck/pull/53/commits/2287a2a3d9bc9a19a693e70c99d943ae5db4172e), [tests ](https://github.com/serokell/xrefcheck/pull/53/commits/662419f17b3ea03fe68eed89e88ed504dc055e81) and [changed `README`](https://github.com/serokell/xrefcheck/pull/53/commits/1e0bafe9e2ce1e4df03cd9d699549b9bcc06a64d)

* One "issue" that isn't stated in `README` is the fact that `<!-- xrefcheck: ignore link -->`  is considered as a blank line (after converting to html), so that this way of formatting can't be used to ignore the link in the middle of the paragraph:
Example:
    ```markdown
    html comment is the same as blank line
    <!-- xrefcheck: ignore link -->
    [link name](https://link.com) blablabla
    ```

Martoon-00:
If I write a comment like `xrefcheck: ignore`, I get the error message like:

```
Error when scanning ./contracts/tezos_examples/README.md: unrecognised option " "ignore" perhaps you meant <"ignore link"|"ignore paragraph"|"ignore file"> (19:24-49)
```

Something seems wrong here.

Bwt, the fact that source location is attached is really cool :+1:

PR #57, opened 2021-02-25
[#55] GitLab support

### LINKED ISSUE #55: [BUG] Support Gitlab's header-to-anchor conversion properly
# Description

When I implemented anchors derivation from headers, I used Github as an example, thinking that since both Github and Gitlab are using CommonMark, they should behave in the same way. And meh, I was wrong.

While `A & B` or `A / B` are translated to `a--b` anchor in Github, in Gilab the result will be just `a-b`.

# To Reproduce

Make a page with `A & B` header and `#a-b` / `#a--b` anchors. The former is treated as a valid anchor, not the later, which is wrong for Gitlab repositories.

# Expected behavior

Behaviour for the mentioned anchors should be the opposite.

# Environment

- Ubuntu
- `5ba3605a`

Martoon-00:
## Description

It turned out that anchors generated from headers have minor differences in GitHub and GitLab, we handle this here.

## Related issue(s)

Resolves #55 

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/src-files/def-config.yaml
+++ b/src-files/def-config.yaml
@@ -56,3 +56,12 @@ verification:
   # It is an optional parameter, so it can be omitted.
   ignoreRefs:
     []
+
+# Parameters of scanners for various file types.
+scanners:
+
+  markdown:
+    # Flavor of markdown, e.g. GitHub-flavor.
+    #
+    # This affects which anchors are generated for headers.
+    flavor: GitHub
--- a/src/Xrefcheck/CLI.hs
+++ b/src/Xrefcheck/CLI.hs
@@ -19,14 +19,14 @@ module Xrefcheck.CLI
 
 import qualified Data.List as L
 import Data.Version (showVersion)
-import Options.Applicative (Parser, ReadM, command, eitherReader, execParser, flag', footerDoc, fullDesc,
-                            help, helper, hsubparser, info, infoOption, long, metavar, option, progDesc,
-                            short, strOption, switch, value)
+import Options.Applicative
+  (Parser, ReadM, command, eitherReader, execParser, flag', footerDoc, fullDesc, help, helper,
+  hsubparser, info, infoOption, long, metavar, option, progDesc, short, strOption, switch, value)
 import Options.Applicative.Help.Pretty (Doc, displayS, fill, fillSep, indent, renderPretty, text)
 
 import Paths_xrefcheck (version)
-import Xrefcheck.Config
 import Xrefcheck.Core
+import Xrefcheck.Scan
 
 modeReadM :: ReadM VerifyMode
 modeReadM = eitherReader $ \s ->
--- a/src/Xrefcheck/CLI.hs
+++ b/src/Xrefcheck/CLI.hs
@@ -17,7 +17,9 @@ module Xrefcheck.CLI
     , getCommand
     ) where
 
+import qualified Data.Char as C
 import qualified Data.List as L
+import qualified Data.Text as T
 import Data.Version (showVersion)
 import Options.Applicative
   (Parser, ReadM, command, eitherReader, execParser, flag', footerDoc, fullDesc, help, helper,
@@ -45,7 +47,7 @@ modeReadM = eitherReader $ \s ->
 
 data Command
   = DefaultCommand Options
-  | DumpConfig FilePath
+  | DumpConfig Flavor FilePath
 
 data Options = Options
     { oConfigPath       :: Maybe FilePath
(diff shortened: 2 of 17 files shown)


alyoanton9 (commented):
LGTM, from what I can tell (with my experience).

I just wish you to answer my questions in the comments on code.

src/Xrefcheck/Config.hs
```diff
@@ -7,33 +7,38 @@
 
 module Xrefcheck.Config where
 
+import qualified Unsafe
```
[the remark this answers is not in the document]
Martoon-00: We [tend](https://gitlab.com/morley-framework/morley/-/blob/master/docs/code-style.md#imports) to group imports, prelude-related first, then third party libraries, then local packages. `Unsafe` is from prelude, so appears at the top.

src/Xrefcheck/Config.hs
```diff
@@ -57,18 +65,119 @@ makeLensesWith postfixFields ''VerifyConfig
...
+      assert (off > processedLen) $
+        pickConfigSubstring processedLen (off - processedLen) :
+        replaceHole (pickConfigSubstring off len) ++
+        replaceHoles (off + len) locs
+
+    holeItemRegex :: R.Regex
+    holeItemRegex = R.makeRegex ("(^|:)([ ]*):PLACEHOLDER:([^:]+):" :: Text)
+
+    holeListRegex :: R.Regex
+    holeListRegex = R.makeRegex ("^([ ]*-[ ]*):PLACEHOLDER:([^:]+):" :: Text)
```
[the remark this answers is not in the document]
Martoon-00: I supposed that list hole allows inserting a list of entries, for instance [`virtualFiles`](https://github.com/serokell/xrefcheck/pull/57/files#diff-5b93233819d5f91ba783808ea00b7ec4f93d64c58b6b2d1b38e11d692facbd52R34) is now defined as a list placeholder.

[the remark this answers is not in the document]
alyoanton9: Yeah, it's clear. I mean that `:PLACEHOLDER:virtualFiles` and, for example, `:PLACEHOLDER:repositoryType` can be matched with `holeItemRegex`, and it seems that the hole always looks like `:PLACEHOLDER:name`, where `name` is just sequence of characters and numbers(?). So that I can't catch why we need 2 different regex and ask for an example.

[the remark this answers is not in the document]
Martoon-00: Ah, I see.

Well, in the case of the list, I need to prepend each entry with a `-`. And it is not trivial to figure out automatically where this `-` should be, because yaml has to be properly indented, and indentation may differ depending on the practices used in this particular file. So the simplest I can do is to assume that the placeholder is also prepended with `-` and replicate this `-` at whatever indentation it is.

So the difference is that for lists I expect this `-` before the placeholder, and for single entries the is no such prefix. It turned out that fetching the `-` was simpler with regexps.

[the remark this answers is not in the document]
alyoanton9: Oooh, I see now. Thanks!

alyoanton9 (approved):

src-files/def-config.yaml
```diff
@@ -56,3 +56,12 @@ verification:
...
     []
+
+# Parameters of scanners for various file types.
+scanners:
+
+  markdown:
+    # Type of repository.
+    #
+    # This affects which anchors are generated for headers.
+    repositoryType: GitHub
```
gromakovsky: Maybe let's call it `flavor` or `format` or something like that? There are various specifications of markdown: [CommonMark](https://commonmark.org/) is probably the most popular basic specification and GitHub-flavored markdown is an extension of it. GitLab-flavored markdown is another extension of CommonMark (it's CommonMark + some stuff AFAIK). So I think we are interested not in _repository type_ or _service where you push your repository_, but rather in markdown flavor that you follow. For example, if you use neither GitLab nor GitHub, there is still a chance that your service mandates GitHub-flavored markdown for example, so it will make sense to write `flavor: GitHub`, but less sense to write `repositoryType: GitHub`.

In future versions I guess we can also support `flavor: CommonMark` here for example.

src/Xrefcheck/Core.hs
```diff
@@ -38,6 +38,14 @@ data RepoType
   | GitLab
   deriving (Show)
 
+allRepoTypes :: [RepoType]
+allRepoTypes = [GitHub, GitLab]
```
gromakovsky, line 42: Btw, you could also use `Bounded` and `Enum` I suppose.

src-files/def-config.yaml
Martoon-00: Hmm, sounds pretty good, gonna do.

src/Xrefcheck/Core.hs
Martoon-00: Yeah, I wanted to do so first, but then imagined that there can be some GitHub-specific settings in the future (some field in `GitHub` ctor), so preferred to go with this thing from the beginning.

src/Xrefcheck/CLI.hs
```diff
@@ -47,7 +47,7 @@ modeReadM = eitherReader $ \s ->
 
 data Command
   = DefaultCommand Options
-  | DumpConfig RepoType FilePath
+  | DumpConfig Flavor FilePath
```
gromakovsky, line 50: And for this command it actually makes sense to take `RepoType` rather than `Flavor`. Currently these are equal sets (of repo types and MD flavors), but one day we may support another repo type that may use an existing flavor (e. g. GH-flavored MD). Then `RepoType` will become a different entity and it will be appropriate here. That repo type may also have different `notScanned` and `virtualFiles` for example.

That said, I don't think it's _necessary_ to bother with this stuff _now_, I am just pointing it out. So I think that `RepoType` and `Flavor` are 2 different entities, currently they would be equal and we use `Flavor` to denote `RepoType` also and that's fine for now.

gromakovsky (approved):
I reviewed superficially, didn't check complex logic.

src/Xrefcheck/CLI.hs
Martoon-00: Yeah, at least half of me also thinks so.

I guess I'll leave a comment about this distinction and rename the CLI option.

PR #63, opened 2021-03-30
Fix docker image upload

balsoft:
## Description

The previous way of uploading images to dockerhub exposed our password.
Prevent this from happening by using a separate script instead of calling
skopeo directly.


scripts/upload-docker-image.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
zhenyavinogradov, line 1: `reuse lint` in CI is complaining about this file

zhenyavinogradov (approved):

PR #66, opened 2021-07-08
Handle ftp links

### LINKED ISSUE #47: Handle ftp links

Currently we support only `http` and `https` links. If there is an `ftp://` link, you will get something like
```
   ➥  In file ./Haskell and Dimensions/article.md
     bad reference (external) at src:149:644-713:
       - text: "here"
       - link: ftp://statgen.ncsu.edu/pub/thorne/molevoclass/AtchleyOct19.pdf
       - anchor: -
     ⛂  Bad url (expected 'http' or 'https')
```

# Acceptance criteria

If an `ftp://` link is encountered, one of the following should happen:
1. It's checked the same way as `http://` and `https://` links.
2. If (1) is not doable, such a link is automatically ignored and a message is printed saying that the link was ignored.

worm2fed:
## Description

Problem:
Currently we support only http and https links. If there is an `ftp://` link, you will get exception.

Solution:
Use `ftp-client` to check connection to ftp, see response statuses and check file existence. This produces adding new error types and small refactoring.

## Related issue(s)

Resolves #47 

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -4,21 +4,21 @@
 
 spec-version: 0.31.0
 
-name:                xrefcheck
-version:             0.2
-github:              serokell/xrefcheck
-license:             MPL-2.0
-license-file:        LICENSE
-author:              Kostya Ivanov, Serokell
-maintainer:          Serokell <hi@serokell.io>
-copyright:           2018-2019 Serokell <https://serokell.io>
+name: xrefcheck
+version: 0.2
+github: serokell/xrefcheck
+license: MPL-2.0
+license-file: LICENSE
+author: Kostya Ivanov, Serokell
+maintainer: Serokell <hi@serokell.io>
+copyright: 2018-2019 Serokell <https://serokell.io>
 
 extra-source-files:
-- README.md
-- CHANGES.md
-- src-files/*
+  - README.md
+  - CHANGES.md
+  - src-files/*
 
-description:         Please see the README on GitHub at <https://github.com/serokell/xrefcheck#readme>
+description: Please see the README on GitHub at <https://github.com/serokell/xrefcheck#readme>
 
 default-extensions:
   - AllowAmbiguousTypes
@@ -71,6 +71,7 @@ dependencies:
   - filepath
   - file-embed
   - fmt
+  - ftp-client
   - Glob
   - http-client
   - http-types
@@ -100,31 +101,31 @@ library:
   source-dirs: src
 
   generated-other-modules:
-  - Paths_xrefcheck
+    - Paths_xrefcheck
 
 executables:
   xrefcheck:
-    main:                Main.hs
-    source-dirs:         exec
+    main: Main.hs
+    source-dirs: exec
     generated-other-modules:
-    - Paths_xrefcheck
+      - Paths_xrefcheck
     ghc-options:
-    - -threaded
-    - -rtsopts
-    - -with-rtsopts=-N
-    - -O2
+      - -threaded
+      - -rtsopts
+      - -with-rtsopts=-N
+      - -O2
     dependencies:
-    - xrefcheck
+      - xrefcheck
 
 tests:
   xrefcheck-tests:
-    main:                Main.hs
-    source-dirs:         tests
+    main: Main.hs
+    source-dirs: tests
     generated-other-modules:
-    - Paths_xrefcheck
+      - Paths_xrefcheck
     dependencies:
-    - xrefcheck
-    - hspec
-    - QuickCheck
+      - xrefcheck
+      - hspec
+      - QuickCheck
     build-tools:
-    - hspec-discover
+      - hspec-discover
--- a/src/Xrefcheck/Verify.hs
+++ b/src/Xrefcheck/Verify.hs
@@ -22,23 +22,29 @@ module Xrefcheck.Verify
     , checkExternalResource
     ) where
 
-import Control.Concurrent.Async (forConcurrently, withAsync)
-import Control.Monad.Except (MonadError (..))
+import qualified Data.ByteString as BS
+import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Map as M
 import qualified Data.Text as T
+import qualified GHC.Exts as Exts
+import qualified System.FilePath.Glob as Glob
+
+import Control.Concurrent.Async (forConcurrently, withAsync)
+import Control.Monad.Except (MonadError (..))
+import Data.Maybe (fromJust)
 import Data.Text.Metrics (damerauLevenshteinNorm)
 import Fmt (Buildable (..), blockListF', listF, (+|), (|+))
-import qualified GHC.Exts as Exts
+import Network.FTP.Client (FTPResponse (..), ResponseStatus (..), login, nlst, withFTP)
 import Network.HTTP.Client (HttpException (..), HttpExceptionContent (..), responseStatus)
-import Network.HTTP.Req (GET (..), HEAD (..), HttpException (..), NoReqBody (..), defaultHttpConfig,
-                         ignoreResponse, req, runReq, useURI)
+import Network.HTTP.Req
+  (GET (..), HEAD (..), HttpException (..), NoReqBody (..), defaultHttpConfig, ignoreResponse, req,
+  runReq, useURI)
 import Network.HTTP.Types.Status (Status, statusCode, statusMessage)
 import System.Console.Pretty (Style (..), style)
 import System.Directory (canonicalizePath, doesDirectoryExist, doesFileExist)
 import System.FilePath (takeDirectory, (</>))
-import qualified System.FilePath.Glob as Glob
 import Text.Regex.TDFA.Text (Regex, regexec)
-import Text.URI (mkURI)
+import Text.URI (Authority (..), URI (..), mkURI, unRText)
 import Time (RatioNat, Second, Time (..), ms, threadDelay, timeout)
 
 import Xrefcheck.Config
@@ -60,9 +66,7 @@ deriving instance Semigroup (VerifyResult e)
 deriving instance Monoid (VerifyResult e)
 
 instance Buildable e => Buildable (VerifyResult e) where
-    build vr = case verifyErrors vr of
-        Nothing   -> "ok"
-        Just errs -> listF errs
+    build vr = maybe "ok" listF $ verifyErrors vr
 
 verifyOk :: VerifyResult e -> Bool
 verifyOk (VerifyResult errors) = null errors
@@ -97,7 +101,8 @@ data VerifyError
     | AmbiguousAnchorRef FilePath Text (NonEmpty Anchor)
     | ExternalResourceInvalidUri
     | ExternalResourceUnknownProtocol
-    | ExternalResourceUnavailable Status
+    | ExternalHttpResourceUnavailable Status
+    | ExternalFtpResourceUnavailable FTPResponse
     | ExternalResourceSomeError Text
     deriving (Show)
 
@@ -118,10 +123,13 @@ instance Buildable VerifyError where
         ExternalResourceInvalidUri ->
             "⛂  Invalid url\n"
         ExternalResourceUnknownProtocol ->
-            "⛂  Bad url (expected 'http' or 'https')\n"
-        ExternalResourceUnavailable status ->
+            "⛂  Bad url (expected 'http','https' or 'ftp')\n"
+        ExternalHttpResourceUnavailable status ->
             "⛂  Resource unavailable (" +| statusCode status |+ " " +|
             decodeUtf8 @Text (statusMessage status) |+ ")\n"
+        ExternalFtpResourceUnavailable status ->
+            "⛂  Resource unavailable (" +| frStatus status |+ " " +|
+            decodeUtf8 @Text (frMessage status) |+ ")\n"
         ExternalResourceSomeError err ->
             "⛂  " +| build err |+ "\n\n"
       where
@@ -129,6 +137,9 @@ instance Buildable VerifyError where
             []  -> "\n"
             [h] -> ",\n   did you mean " +| h |+ "?\n"
             hs  -> ", did you mean:\n" +| blockListF' "    -" build hs
+
+instance Buildable ResponseStatus where
+  build = show
 
 verifyRepo
     :: Rewrite
@@ -248,20 +259,25 @@ verifyReference config@VerifyConfig{..} mode progressRef (RepoInfo repoInfo)
                         givenAnchors
                 in throwError $ AnchorDoesNotExist anchor similarAnchors
 
-checkExternalResource :: VerifyConfig
-                      -> Text
-                      -> IO (VerifyResult VerifyError)
+checkExternalResource :: VerifyConfig -> Text -> IO (VerifyResult VerifyError)
 checkExternalResource VerifyConfig{..} link
     | isIgnored = return mempty
     | doesReferLocalhost = return mempty
     | otherwise = fmap toVerifyRes $ do
-        makeRequest HEAD 0.3 >>= \case
-            Right () -> return $ Right ()
-            Left   _ -> makeRequest GET 0.7
+          uri <- mkURI link
+            -- `catchAny` \_ ->
+            -- runExceptT $ throwError ExternalResourceInvalidUri
+
+          case unRText <$> uriScheme uri of
+            Just "http" -> checkHttp uri
+            Just "https" -> checkHttp uri
+            Just "ftp" -> checkFtp uri
+            _ -> runExceptT $ throwError ExternalResourceUnknownProtocol
   where
     isIgnored =
-        let maybeIsIgnored = (doesMatchAnyRegex link) <$> vcIgnoreRefs
+        let maybeIsIgnored = doesMatchAnyRegex link <$> vcIgnoreRefs
         in fromMaybe False maybeIsIgnored
+
     doesReferLocalhost = any (`T.isInfixOf` link) ["://localhost", "://127.0.0.1"]
 
     doesMatchAnyRegex :: Text -> ([Regex] -> Bool)
@@ -272,10 +288,18 @@ checkExternalResource VerifyConfig{..} link
                 Nothing -> False
             Left _ -> False
 
-    makeRequest :: _ => method -> RatioNat -> IO (Either VerifyError ())
-    makeRequest method timeoutFrac = runExceptT $ do
-        uri <- mkURI link
-             & maybe (throwError ExternalResourceInvalidUri) pure
+    checkHttp uri = makeHttpRequest uri HEAD 0.3 >>= \case
+      Right () -> return $ Right ()
+      Left   _ -> makeHttpRequest uri GET 0.7
+
+
+    makeHttpRequest
+      :: _
+      => URI
+      -> method
+      -> RatioNat
+      -> IO (Either VerifyError ())
+    makeHttpRequest uri method timeoutFrac = runExceptT $ do
         parsedUrl <- useURI uri
                    & maybe (throwError ExternalResourceUnknownProtocol) pure
         let reqLink = case parsedUrl of
@@ -288,8 +312,8 @@ checkExternalResource VerifyConfig{..} link
 
         let maxTime = Time @Second $ unTime vcExternalRefCheckTimeout * timeoutFrac
 
-        mres <- liftIO (timeout maxTime $ void reqLink)
-                `catch` (either throwError (\() -> return (Just ())) . interpretErrors)
+        mres <- liftIO (timeout maxTime $ void reqLink) `catch`
+          (either throwError (\() -> return (Just ())) . interpretErrors)
         maybe (throwError $ ExternalResourceSomeError "Response timeout") pure mres
 
     isAllowedErrorCode = or . sequence
@@ -306,5 +330,46 @@ checkExternalResource VerifyConfig{..} link
             HttpExceptionRequest _ exc -> case exc of
                 StatusCodeException resp _
                     | isAllowedErrorCode (statusCode $ responseStatus resp) -> Right ()
-                    | otherwise -> Left $ ExternalResourceUnavailable (responseStatus resp)
+                    | otherwise -> Left $ ExternalHttpResourceUnavailable (responseStatus resp)
                 other -> Left . ExternalResourceSomeError $ show other
+
+
+    checkFtp :: URI -> IO (Either VerifyError ())
+    checkFtp uri = runExceptT $ do
+      -- get authority which stores host and port
+      authority <- case uriAuthority uri of
+        Right a -> pure a
+        Left _ -> throwError ExternalResourceInvalidUri
+      let host = T.unpack . unRText $ authHost authority
+      let port :: Int = fromIntegral . fromMaybe 21 $ authPort authority
+      let pieces = uriPath uri
+      -- in case there aro no pieces - smth wrong with url
+      unless (isJust pieces) $
+        throwError ExternalResourceInvalidUri
+      let (trailing, pathList) = fromJust pieces
+      -- in case there are trailing slash - smth wrong with url
+      when trailing $
+        throwError ExternalResourceInvalidUri
+      -- build path from pieces
+      let path = T.unpack . mconcat
+            . NonEmpty.toList
+            . NonEmpty.intersperse "/"
+            . NonEmpty.map unRText
+            $ pathList
+      makeFtpRequest host port path
+
+    makeFtpRequest :: String -> Int -> FilePath -> ExceptT VerifyError IO ()
+    makeFtpRequest host port path = withFTP host port $
+      \handle resp -> do
+        -- check connection status
+        when (frStatus resp /= Success) $
+          Left $ ExternalFtpResourceUnavailable resp
+        -- anonymous login
+        loginResp <- login handle "anonymous" ""
+        -- check login status
+        when (frStatus loginResp /= Success) $
+          Left $ ExternalFtpResourceUnavailable loginResp
+        file <- nlst handle [ path ]
+        -- check file exists
+        when (BS.null file) $
+          Left $ FileDoesNotExist path


src/Xrefcheck/Verify.hs
```diff
@@ -81,230 +87,324 @@ toVerifyRes = VerifyResult . either one (\() -> [])
...
+      let host = T.unpack . unRText $ authHost authority
+      let port :: Int = fromIntegral . fromMaybe 21 $ authPort authority
+      let pieces = uriPath uri
+      -- in case there aro no pieces - smth wrong with url
+      unless (isJust pieces) $
+        throwError ExternalResourceInvalidUri
+      let (trailing, pathList) = fromJust pieces
+      -- in case there are trailing slash - smth wrong with url
+      when trailing $
+        throwError ExternalResourceInvalidUri
```
dcastro, line 379: Same here: In this function, we're throwing `ExternalResourceInvalidUri` in three places, but we should be throwing a `ExternalResourceInvalidURL` instead.

Also, since we know _why_ the URI is not a valid URL (no authority, no path, etc), the error message should include this piece of information.

src/Xrefcheck/Verify.hs
```diff
@@ -81,230 +87,324 @@ toVerifyRes = VerifyResult . either one (\() -> [])
...
+      authority <- case uriAuthority uri of
+        Right a -> pure a
+        Left _ -> throwError ExternalResourceInvalidUri
+      let host = T.unpack . unRText $ authHost authority
+      let port :: Int = fromIntegral . fromMaybe 21 $ authPort authority
+      let pieces = uriPath uri
+      -- in case there aro no pieces - smth wrong with url
+      unless (isJust pieces) $
+        throwError ExternalResourceInvalidUri
+      let (trailing, pathList) = fromJust pieces
```
dcastro, line 376: We shouldn't use `fromJust`.
We should pattern match on `pieces :: Maybe etc`, and then branch out.

Yes, in this instance, it technically _is_ safe to use `fromJust` here, because we're guarded by `isJust` 2 lines up.
_But_, during a refactoring, while moving things around, it's easy to move/copy-paste the `fromJust` somewhere else, and leave the `isJust` behind. When that happens, the code will still compile, but then fail to run.
I know it sounds a bit farfetched, but this does happen in practice.

For this reason, pattern matching and branching out is considered a more "refactoring-safe" solution.

src/Xrefcheck/Verify.hs
```diff
@@ -81,230 +87,324 @@ toVerifyRes = VerifyResult . either one (\() -> [])
...
+      -- in case there are trailing slash - smth wrong with url
+      when trailing $
```
dcastro, line 378: Is it really necessary to exclude a trailing slash?

src/Xrefcheck/Orphans.hs
```diff
@@ -0,0 +1,37 @@
...
+    case message of
+      SingleLine s -> s
+      MultiLine ss -> BS.concat ss
+    )
+
+instance Buildable FTPResponse where
+  build FTPResponse{..} =  frStatus |+ " (" +| frCode |+ ") " +| frMessage |+ ""
+
+instance Buildable FTPException where
+  build (BadProtocolResponseException _) = "Raw FTP exception"
```
dcastro, line 32: Why not display the argument?

src/Xrefcheck/Verify.hs
```diff
@@ -81,230 +87,324 @@ toVerifyRes = VerifyResult . either one (\() -> [])
...
+
+    ExternalResourceUnknownProtocol ->
+      "⛂  Bad url (expected 'http','https' or 'ftp')\n"
+
+    ExternalHttpResourceUnavailable status ->
+      "⛂  Resource unavailable (" +| statusCode status |+ " " +|
+      decodeUtf8 @Text (statusMessage status) |+ ")\n"
+
+    ExternalFtpResourceUnavailable status ->
+      "⛂  Resource unavailable (" +| status |+ ")\n"
```
dcastro, line 140:
```suggestion
    ExternalFtpResourceUnavailable response ->
      "⛂  Resource unavailable (" +| response |+ ")\n"
```

dcastro: Also, the `response` might span multiple lines, so it's probably best to display it below `⛂  Resource unavailable`, instead of in between `(` `)`

src/Xrefcheck/Orphans.hs
```diff
@@ -0,0 +1,37 @@
...
+
+instance Buildable FTPMessage where
+  build message = build $ decodeUtf8 @Text (
+    case message of
+      SingleLine s -> s
+      MultiLine ss -> BS.concat ss
+    )
+
+instance Buildable FTPResponse where
+  build FTPResponse{..} =  frStatus |+ " (" +| frCode |+ ") " +| frMessage |+ ""
```
dcastro, line 29: This is probably too much info to display in one single line. Try formatting it across multiple lines (Hint: `unlinesF` will make this easier)

src/Xrefcheck/Verify.hs
```diff
@@ -81,230 +87,324 @@ toVerifyRes = VerifyResult . either one (\() -> [])
...
+    makeFtpRequest host port path secure = handler host port $
+      \handle response -> do
+        -- check connection status
+        when (frStatus response /= Success) $
+          throwError $ ExternalFtpResourceUnavailable response
+        -- anonymous login
+        loginResp <- login handle "anonymous" ""
+        -- check login status
+        when (frStatus loginResp /= Success) $
+          throwError $ ExternalFtpResourceUnavailable loginResp
```
dcastro, line 404: This does not mean the resource is unavailable, it means the server does not support anonymous login.

package.yaml
```diff
@@ -100,31 +100,44 @@ library:
...
-    - -O2
+      - -threaded
+      - -rtsopts
+      - -with-rtsopts=-N
+      - -O2
     dependencies:
-    - xrefcheck
+      - xrefcheck
+
+  links-tests:
```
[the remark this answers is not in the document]
worm2fed: we discussed it in slack, so Kirill asked to make it as separate executable
this is the reason I've add it to `executables` section

src/Xrefcheck/Orphans.hs
dcastro: Nevermind, just saw the conversation on slack.

package.yaml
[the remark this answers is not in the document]
dcastro: I think he meant it in the broader sense of the word. Technically, both `tests` and `executables` components are executable files. It's just that the former is an executable with tests, and the latter is an executable that is meant to be distributed.

src/Xrefcheck/Verify.hs
worm2fed: if there are trailing slash - this is directory, not a file

dcastro: I think it's fine to have a link to an FTP directory though, right?

src/Xrefcheck/Verify.hs
worm2fed: I've got it! thanks

src/Xrefcheck/Verify.hs
worm2fed: Discussed with Kirill, that this row exception can be a few words or few gigabytes

src/Xrefcheck/Verify.hs
worm2fed: Hw.. you can open this link in your browser
ftp://statgen.ncsu.edu/pub/thorne/molevoclass/AtchleyOct19.pdf

but you can't open this one ftp://statgen.ncsu.edu/pub/thorne/molevoclass/ - it's required to open external app to connect to this server

src/Xrefcheck/Verify.hs
```diff
@@ -81,230 +87,324 @@ toVerifyRes = VerifyResult . either one (\() -> [])
...
+      case regexec regex src of
+        Right res -> case res of
+          Just (before, match, after, _) ->
+            null before && null after && not (null match)
+          Nothing -> False
+        Left _ -> False
+
+    checkHttp :: URI -> ExceptT VerifyError IO ()
+    checkHttp uri = do
+      res <- liftIO $ runExceptT $ makeHttpRequest uri HEAD 0.3
```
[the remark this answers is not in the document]
worm2fed: > I think we could use tryE instead of liftIO . runExceptT here

hm... I can see we have `transformers` in our dependencies, but in my generated `cabal.project.freeze` there are no
I have only 
```
any.exception-transformers ==0.4.0.9,
any.data-accessor-transformers ==0.2.1.7,
any.transformers-base ==0.4.5.2,
any.transformers-bifunctors ==0.1,
transformers-compat +five-three,
any.transformers-compat ==0.6.6,
any.transformers-fix ==1.0,
any.writer-cps-transformers ==0.5.6.1,
```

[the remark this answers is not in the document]
worm2fed: I'll use `catchError`

src/Xrefcheck/Verify.hs
dcastro: I can open `ftp://statgen.ncsu.edu/pub/thorne/molevoclass/` in my file explorer, which shows a folder with files in it.

src/Xrefcheck/Orphans.hs
worm2fed: not sure I've got your point..
do you mean this?
```
instance Buildable FTPResponse where
  build FTPResponse{..} = unlinesF
    [ mconcat [build frStatus, " (", build frCode, "):"]
    , build frMessage
    ]
```

src/Xrefcheck/Verify.hs
worm2fed: okay, if you insist I'll delete this check)

src/Xrefcheck/Orphans.hs
dcastro: Yup, I meant that using `unlinesF` is better than concatenating strings and manually adding `\n` in between them.

Btw, for formatting a single line, I think `frStatus |+ " (" +| frCode |+ "):"` reads better than `mconcat [build frStatus, " (", build frCode, "):"]`

src/Xrefcheck/Verify.hs
dcastro: Thanks :)

We should also add a couple of tests for ftp/ftps links to 1) an empty directory and 2) a non-empty directory

worm2fed: What we suppose to do in case of empty dir? Is it valid case for us?

dcastro: I'd say so. If it's a valid FTP link (which it is), then we should consider it valid.

links-tests/Test/Xrefcheck/FtpLinks.hs
```diff
@@ -0,0 +1,84 @@
...
+
+test_FtpLinks :: TestTree
+test_FtpLinks = askOption $ \(FtpHostOpt host) -> do
+  let notRaisedMsg = "expected exceptions was not raised"
+
+  testGroup "Ftp links handler"
+    [ testCase "handles correct link to file" $ do
+        let link = host <> "/pub/file_exists.txt"
+        result <- checkExternalResource config link
+        verifyOk result @?= True
```
dcastro, line 55: If `checkExternalResource` returns some errors, then comparing two booleans using `verifyOk result @?= True` will display a very generic, and not really helpful message in the test logs, something like "Expected True, got False"

It would be better to compare the full data type, e.g. `result @?= VerifyResult []`.
That way, if the response contains some errors, those errors will be displayed in the test logs as well, giving the reader some idea of _why_ the test failed.

worm2fed: Good idea, you're right

now instead of "Expected True, got False" I can see 
```
expected: VerifyResult []
         but got: VerifyResult [FileDoesNotExist "pub/file_exists.txt"]
```
for missing file

src/Xrefcheck/Verify.hs
```diff
@@ -272,12 +290,23 @@ checkExternalResource VerifyConfig{..} link
...
+      => URI
+      -> method
+      -> RatioNat
+      -> ExceptT VerifyError IO ()
+    makeHttpRequest uri method timeoutFrac = do
+        parsedUrl <- case useURI uri of
+          -- accordingly to source code - Nothing can be only in case when
+          -- protocol is not http or https, but we've checked it already
+          -- so just in case we throw exception here
+          Nothing -> throwError $ ExternalResourceInvalidUrl "Something wrong"
```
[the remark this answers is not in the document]
worm2fed: good

src/Xrefcheck/Verify.hs
worm2fed: I've added a test, but not sure it's working... I need ftp with empty dir to test it)

package.yaml
```diff
@@ -1,24 +1,23 @@
...
+copyright: 2018-2021 Serokell <https://serokell.io>
+description: Please see the README on GitHub at <https://github.com/serokell/xrefcheck#readme>
 
 extra-source-files:
-- README.md
-- CHANGES.md
-- src-files/*
-
-description:         Please see the README on GitHub at <https://github.com/serokell/xrefcheck#readme>
+  - README.md
```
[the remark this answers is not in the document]
worm2fed: please check

src/Xrefcheck/Verify.hs
```diff
@@ -327,5 +360,49 @@ checkExternalResource VerifyConfig{..} link
...
+
+    checkFtp :: URI -> Bool -> ExceptT VerifyError IO ()
+    checkFtp uri secure = do
+      -- get authority which stores host and port
+      authority <- case uriAuthority uri of
+        Right a -> pure a
+        Left _ -> throwError $
+          ExternalResourceInvalidUrl (Just "FTP path must be absolute")
+      let host = T.unpack . unRText $ authHost authority
+      let port :: Int = fromIntegral . fromMaybe 21 $ authPort authority
```
Martoon-00, line 374: Here you cast `Word` to `Int`, which may cause an overflow. Indeed, such an issue is very unlikely, but nothing prevents the user from trying some weird port - and getting a link with a negative port in the error message. I would add an explicit check on that conversion is valid (for instance with the help of [toIntegralSized](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Bits.html#v:toIntegralSized)).

links-tests/Test/Xrefcheck/FtpLinks.hs
```diff
@@ -0,0 +1,79 @@
...
+
+test_FtpLinks :: TestTree
+test_FtpLinks = askOption $ \(FtpHostOpt host) -> do
+  let notRaisedMsg = "expected exceptions was not raised"
+
+  testGroup "Ftp links handler"
+    [ testCase "handles correct link to file" $ do
+        let link = host <> "/pub/file_exists.txt"
+        result <- checkExternalResource config link
+        result @?= VerifyResult []
```
Martoon-00, line 56: Somehow tests pass locally, though I cannot figure out how is this possible taking into account that I have no `pub` directory or something like that. Is it some ftp magic?

src/Xrefcheck/Verify.hs
```diff
@@ -327,5 +360,49 @@ checkExternalResource VerifyConfig{..} link
...
+          throwError $ ExternalFtpResourceUnavailable response
+        -- anonymous login
+        loginResp <- login handle "anonymous" ""
+        -- check login status
+        when (frStatus loginResp /= Success) $
+          throwError $ ExternalFtpException $ UnsuccessfulException loginResp
+        file <- nlst handle [ path ]
+        -- check file exists
+        when (BS.null file) $
+          throwError $ FileDoesNotExist path
```
Martoon-00, line 406: And a minor thing: my opinion is that it's worth using a dedicated error for this and have separate `LocalFileDoesNotExist` and `FtpFileDoesNotExist`.

I tended to prefix error message with `⛀` for local files and `⛂` for remote resources, and now we got an exception from this rule which may cause a confusion to the user.

src/Xrefcheck/Verify.hs
```diff
@@ -327,5 +360,49 @@ checkExternalResource VerifyConfig{..} link
...
+      let host = T.unpack . unRText $ authHost authority
+      let port :: Int = fromIntegral . fromMaybe 21 $ authPort authority
+      -- build path from pieces
+      path <- case uriPath uri of
+        Nothing -> pure "/"
+        Just (_, pieces) -> pure . T.unpack . mconcat
+          . NonEmpty.toList
+          . NonEmpty.intersperse "/"
+          . NonEmpty.map unRText
+          $ pieces
```
Martoon-00, line 382: Quite annoying that the library does not provide an utility for this rendering :unamused: 

By the way, `T.unpack` is available in our prelude as `toString`, and `NonEmpty.toList` - probably just as `toList`. Since you already add a module with orphans, maybe it would be convenient to write your own `ToString (RText t)` instance and avoid calling `unRText` explicitly each time.

Martoon-00 (commented):
A really decent work, hope it finishes smoothly :+1: 

I left a bunch of comments, most of them should be easy to fix (assuming you find them reasonable).

links-tests/Test/Orphans.hs
```diff
@@ -0,0 +1,19 @@
...
+
+import Network.FTP.Client (FTPException (..))
+import Xrefcheck.Core (Anchor (Anchor), Position (Position))
+import Xrefcheck.Verify (VerifyError (..), VerifyResult (..))
+
+deriving instance Eq a => Eq (VerifyResult a)
+deriving instance Eq VerifyError
+deriving instance Eq FTPException
+deriving instance Eq Anchor
+deriving instance Eq Position
```
dcastro, line 19: Could all of these instances (apart from `Eq FTPException`) be moved to the core library, thus avoiding being orphans?

links-tests/Test/Xrefcheck/FtpLinks.hs
```diff
@@ -0,0 +1,79 @@
...
+
+
+config :: VerifyConfig
+config = cVerification $ defConfig GitHub
+
+test_FtpLinks :: TestTree
+test_FtpLinks = askOption $ \(FtpHostOpt host) -> do
+  let notRaisedMsg = "expected exceptions was not raised"
+
+  testGroup "Ftp links handler"
```
dcastro, line 52: Can we please add a test for an empty path (`let link = host`)?

README.md
```diff
@@ -45,6 +45,7 @@ Both relative and absolute local links are supported out of the box.
 * [url-checker](https://github.com/paramt/url-checker) - GitHub action which checks links in specified files.
 * [linkcheck](https://github.com/filiph/linkcheck) - advanced site crawler, checks for `HTML` files. There are other solutions for this particular task which we don't mention here.
 
+At the moment of writing, listed solutions doesn't support ftp/ftps links.
```
dcastro, line 48:
```suggestion
At the moment of writing, the listed solutions don't support ftp/ftps links.
```

src/Xrefcheck/Verify.hs
```diff
@@ -292,26 +315,36 @@ checkExternalResource VerifyConfig{..} link
...
+      => URI
+      -> method
+      -> RatioNat
+      -> ExceptT VerifyError IO ()
+    makeHttpRequest uri method timeoutFrac = do
+        parsedUrl <- case useURI uri of
+          -- accordingly to source code - Nothing can be only in case when
+          -- protocol is not http or https, but we've checked it already
+          -- so just in case we throw exception here
+          Nothing -> throwError $ ExternalResourceInvalidUrl Nothing
```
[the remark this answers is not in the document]
dcastro: `useHttpURI` and `useHttpsURI` also return a `Maybe`, I think we'd still have to handle the impossible case :/

package.yaml
```diff
@@ -1,24 +1,23 @@
...
-copyright:           2018-2019 Serokell <https://serokell.io>
+name: xrefcheck
+version: 0.2
+github: serokell/xrefcheck
+license: MPL-2.0
+license-file: LICENSE
+author: Kostya Ivanov, Serokell
+maintainer: Serokell <hi@serokell.io>
+copyright: 2018-2021 Serokell <https://serokell.io>
+description: Please see the README on GitHub at <https://github.com/serokell/xrefcheck#readme>
```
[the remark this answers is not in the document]
worm2fed: yes, dunno how to setup formatter do not format such a blocks...
but I think it's common practice for yaml files

src/Xrefcheck/Orphans.hs
```diff
@@ -0,0 +1,40 @@
...
+    , build frMessage
+    ]
+
+instance Buildable FTPException where
+  build (BadProtocolResponseException _) = "Raw FTP exception"
+  build (FailureRetryException e) = build e
+  build (FailureException e) = build e
+  build (UnsuccessfulException e) = build e
+  build (BogusResponseFormatException e) = build e
+
```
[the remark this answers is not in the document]
worm2fed: sorry, can you explain please?

src/Xrefcheck/Verify.hs
```diff
@@ -327,5 +360,49 @@ checkExternalResource VerifyConfig{..} link
...
+    makeFtpRequest host port path secure = handler host port $
+      \handle response -> do
+        -- check connection status
+        when (frStatus response /= Success) $
+          throwError $ ExternalFtpResourceUnavailable response
+        -- anonymous login
+        loginResp <- login handle "anonymous" ""
+        -- check login status
+        when (frStatus loginResp /= Success) $
+          throwError $ ExternalFtpException $ UnsuccessfulException loginResp
```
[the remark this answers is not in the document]
dcastro: Perhaps xrefcheck should also be able to emit warnings (not just errors) :thinking: In which case, we'd emit a warning for links requiring credentials.

And then the user could either use a flag to ignore warnings, or use a `<!-- -->` pragma to ask xrefcheck to ignore that specific link.


> So for instance, in HTTP case if I see that authentication is necessary, or a resource is forbidden for me, I still treat it as a good link because it's just that I have no access to the resource, *but some other people do*.

Nitpick: just wanted to point out that it's possible a URL might request authentication and, when authenticated, return a 404 and be an invalid link after all.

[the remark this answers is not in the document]
dcastro: But, to your point, I agree.

Given that we cannot emit warnings, we have a binary choice to make: treat it as valid or invalid. And we can't know for sure, so we must choose to treat it as valid to avoid false positives.

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
Martoon-00: Oh right, I thought this `Maybe` is about parsing failure, but it's not :/

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
Martoon-00: When I think about warnings, I can't find justified usages for them in the case of xrefcheck (

If we silence them globally, then they become just redundant. If we do not silence them and allow them to be, then the project may gather so much warnings that no one ever dares to rake them out.

If we suppress a warning locally - then this warning could be just an error instead.

I dunno, looks like in Haskell we have all the same issues with warnings, and the only justified use cases for them that I yet realize are: 1. temporarily build a bad code locally; 2. a library that does not yet comply with the modern practices, e.g. some `fmt` not yet using `DerivingStrategies` extension. But in xrefcheck case, there is no sense in "temporarily bad" or "outdated" Markdown, I suppose.

Generally, CI also has no choice but to be binary, either OK or not, everything in between is an eternal backlog. And since xrefcheck is used 98% of the time in CI, not locally, producing a ternary output will probably not change much.

---

I agree that adding warnings is a very tempting option, but I just don't yet see how to implement it in a way that would work.

Maybe we actually need some different mechanism in this case :thinking: 

There was a proposal to allow `xrefcheck` analyzing only git diff. Probably here warnings would shine, but there are some fundamental difficulties in implementing that.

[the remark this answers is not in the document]
dcastro: I agree completely with everything you said. On second thought, there is no use-case for warnings after all.

---

That being the case, we are indeed forced to either assume an "unauthorized link" is valid or invalid, it's a binary choice.

I'm starting to think we should throw an error though... We can't prove the link is valid, so xrefcheck should fail and tell the user "401 Unauthorized - Cannot verify whether the link is valid.".

If the user is _sure_ the link is indeed valid, then they can manually add a `<!-- xrefcheck: ignore ... -->` pragma.

---

Right now, we silently ignore the 401 and simply move on. We're not even giving the user the chance to decide what to do. The link may indeed be invalid, but they'll never know.

IOW, with the current approach, we're avoiding false positives (reporting that a valid link is broken) at the expense of being able to report positives (reporting that a broken link is broken).

---

Additionally, maybe we should add a ` <!-- xrefcheck: expected <statuscode> -->` pragma? :thinking:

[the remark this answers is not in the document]
worm2fed: > If the user is sure the link is indeed valid, then they can manually add a  pragma to silently ignore the link

> Right now, we always silently ignore the 401 and simply move on. We're not giving the user the chance to decide what to do. The link may indeed be invalid, but they'll never know.

I think this way too. 
it's better to give possibility to decide

links-tests/Test/Xrefcheck/FtpLinks.hs
worm2fed: hm...

I have this output for `cabal run -- links-tests --ftp-host "ftp://statgen.ncsu.edu"`
```
Ftp links handler
      handles correct link to file:                FAIL (2.02s)
        links-tests/Test/Xrefcheck/FtpLinks.hs:56:
        expected: VerifyResult []
         but got: VerifyResult [FileDoesNotExist "pub/file_exists.txt"]
      handles empty link (host only):              OK (1.70s)
      handles correct link to non empty directory: OK (1.71s)
      handles correct link to empty directory:     FAIL (1.68s)
        links-tests/Test/Xrefcheck/FtpLinks.hs:71:
        expected: VerifyResult []
         but got: VerifyResult [FileDoesNotExist "empty"]
      throws exception when file not found:        OK (1.86s)

2 out of 5 tests failed (8.97s)
```
this output is expected by me (at this server there are no `/pub/file_exists.txt` and `/empty/`

src/Xrefcheck/Verify.hs
worm2fed: Can't see your symbols(
image

is it a font?

src/Xrefcheck/Verify.hs
worm2fed: good idea

links-tests/Test/Orphans.hs
worm2fed: sure

links-tests/Test/Xrefcheck/FtpLinks.hs
worm2fed: oh, I've got it))

https://github.com/serokell/xrefcheck/blob/79b4e106f0a6d14dd8e4715fac3bbc47644e72ee/src/Xrefcheck/Verify.hs#L295-L295

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
Martoon-00: You want to let the user double-check links that we cannot check?

I think it produces more hassle than a benefit, since 99% of the time I just copy the link from the browser, and now I will also have to fight with CI and add boilerplate each time :/

Maybe we could really add an option, which status codes to ignore. And maybe a global one, not as pragma? So the users can either skip 401 and 403 by default and remember to be careful, or be explicit about ignoring such links.

---

Moving to better handling true positives (finding more broken links) is good, but I think we never can get near 100% accuracy here. Another example: we cannot check anchors in external pages, that would be just too expensive. 

So to provide at least some sensible guarantees, we have to either stick to another edge (0% of false positives) or really allow configuring the behaviour here. And no way the tool should become a burden for the user unless he wants so.

links-tests/Test/Xrefcheck/FtpLinks.hs
Martoon-00: Meeh, right :smile:

src/Xrefcheck/Verify.hs
Martoon-00: Wow that's super weird, I used just plan text there. For me it looks like 
Screenshot from 2021-07-30 16-56-36

src/Xrefcheck/Verify.hs
Martoon-00: Looks pretty now :+1:

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
worm2fed: > Maybe we could really add an option, which status codes to ignore. And maybe a global one, not as pragma? So the users can either skip 401 and 403 by default and remember to be careful, or be explicit about ignoring such links.

okay, we can add only one bool config option, something like "assume protected links broken"

src/Xrefcheck/Verify.hs
worm2fed: I can copy it and search, and google says its uff-8 symbol.. but I can't see it (same in source code)
image

what OS do you have?

worm2fed: image

Martoon-00: Oh sorry, I looked at the wrong position, I see what you mean now.

That's what I get:

Screenshot from 2021-07-30 17-59-11

I'm on Ubuntu 18, there these symbols render well for me in IDEs and in terminal,

worm2fed: can you check what is this font here?
https://fedoraproject.org/wiki/Identifying_fonts

anyway
is it good idea to use such a symbols in our code? maybe someone have same problem

Martoon-00: Here is this symbol, looks like fonts that support it are also mentioned there: https://www.fileformat.info/info/unicode/char/26c2/index.htm

> is it good idea to use such a symbols in our code? maybe someone have same problem

It is my little wish to use such fancy things in this project :stuck_out_tongue: 

In the worst case, you can just blindly copy-paste the symbols.

links-tests/Test/Xrefcheck/FtpLinks.hs
worm2fed: I've added config option for localhost links

src/Xrefcheck/Verify.hs
worm2fed: Check it please

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
worm2fed: done

worm2fed:
hm... can't understand why config test failed.. I've added default settings to our default config

links-tests/Test/Xrefcheck/FtpLinks.hs
Martoon-00: Good! Left a comment there.

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
Martoon-00: Nice!

src/Xrefcheck/Verify.hs
```diff
@@ -327,5 +370,62 @@ checkExternalResource VerifyConfig{..} link
...
+        dirList <- nlst handle [ "-a", path ]
+        when (BS.null dirList) $ do
+          fileSize <- size handle path `catch` \err ->
+            case err of
+              UnsuccessfulException _ -> throwError $ FtpEntryDoesNotExist path
+              FailureException FTPResponse{..} | frCode == 550 ->
+                throwError $ FtpEntryDoesNotExist path
+              _ -> liftIO $ throwIO err
+          when (fileSize == 0) $
+            throwError $ FtpEntryDoesNotExist path
```
Martoon-00, line 429: Could you leave a brief explanation as a comment here? Can't files in FTP really be empty and return `0` size?

worm2fed: @balsoft do you have an answer?

balsoft: This is a mistake on my part. At least per the standard:

>    The server-PI will respond to the SIZE command with a 213 reply
>   giving the transfer size of the file whose pathname was supplied, or
>   an error response if the file does not exist, the size is
>   unavailable, or some other error has occurred.

So if we get the size without any exceptions, it should mean that the file exists (even if the size is 0).

dcastro:
> The term "protected" is not common AFAIU, would be nice to elaborate on it here and in `vSkipProtected`'s haddock

Agreed. Since this check is specific to the `403 Forbidden` status code, I think `vSkipForbidden` would be more suitable.

On that note, since we're giving 403 special treatment, shouldn't we also include `401 Unauthorized`? If so, maybe `vSkipAuthFailures`?

dcastro:
Also, we already seem to be using the term `ignore` for other config fields (e.g. `ignoreRefs`), so maybe we should rename the new `skip*` fields to `ignore*` for consistency.

Martoon-00:
> > The term "protected" is not common AFAIU, would be nice to elaborate on it here and in `vSkipProtected`'s haddock
> 
> Agreed. Since this check is specific to the `403 Forbidden` status code, I think `vSkipForbidden` would be more suitable.
> 
> On that note, since we're giving 403 special treatment, shouldn't we also include `401 Unauthorized`? If so, maybe `vSkipAuthFailures`?

Yeah, it's really weird that I didn't add `401` to the list.

However, to me, 401 and 403 seem interchangeable for the purpose of links checking, they both mean that the tool has no access to the resource that other people can see.

Maybe we could provide a twofold solution:
* Have `vSkipAuthFailures` that is both for `401` and `403`;
* For low-level management - extend the interface that works with links ignorance, so that it is possible to specify error codes:
```yaml
ignoreRefs:
  - https?://google.com
  - link: https?://gitlab.com/morley-framework/morley/.*
    ignoredCodes: [403] 
```

Martoon-00:
> hm... can't understand why config test failed.. I've added default settings to our default config

@worm2fed If this still fires for you, you just need to do what the error message says, verify the changes, and include them into the commit. 

Maybe the error message is really unclear, if you know how to improve it, please feel free to do it.

worm2fed:
@Martoon-00 
> If this still fires for you, you just need to do what the error message says, verify the changes, and include them into the commit.

oh, really strange, because I've did it and no effect. But now it's ok, I found a problem pretty fast)

worm2fed:
I've made changes, please re-review
Hope we can close proceed this PR now)

src/Xrefcheck/CLI.hs
```diff
@@ -142,6 +158,13 @@ traversalOptionsParser = do
     help "Files and folders which we pretend do not exist."
   return TraversalOptions{..}
 
+verifyOptionsParser :: Parser VerifyOptions
+verifyOptionsParser = do
+  voIgnoreLocalhost <- flag True False $
+    long "check-localhost" <>
+    help "If provided localhost links will be checked"
```
Martoon-00, line 165: Minor nitpick: we should be consistent about trailing dots :stuck_out_tongue_closed_eyes: 

Same in some other places.

src/Xrefcheck/CLI.hs
```diff
@@ -69,6 +73,17 @@ addTraversalOptions TraversalConfig{..} (TraversalOptions ignored) =
   , ..
   }
 
+newtype VerifyOptions = VerifyOptions
+  { voIgnoreLocalhost :: Bool
+  }
+
+addVerifyOptions :: VerifyConfig -> VerifyOptions -> VerifyConfig
+addVerifyOptions VerifyConfig{..} VerifyOptions{..} =
```
Martoon-00, line 81: Minor: the reason why `addTraversalOptions` enlists `TraversalOptions`'s fields in its pattern-match explicitly is that, if someone adds one new option to this datatype, the compiler will complain on `addTraversalOptions` function until it is updated respectively.

But `addVerfiyOptions` currently lacks this property and one can forget to update this function when extending `VerifyOptions` datatype, and so things become slightly more difficult to maintain.

.xrefcheck.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # Skip localhost links
+  ignoreLocalhost: true
+
+# Parameters of scanners for various file types.
+scanners:
+  markdown:
+    # Flavor of markdown, e.g. GitHub-flavor.
+    #
+    # This affects which anchors are generated for headers.
+    flavor: GitHub
```
Martoon-00, line 58: Previously I tried to go without this file because it requires efforts duplication when adding new fields to config, but apparently there is no other choice.

What do you think about stripping the comments here and leaving bare yaml? That should make it easier to update this config once we require new fields in the new versions of xrefcheck.

Martoon-00 (commented):
Thanks, looks good now :+1: 

I have only a few more minor comments, and that's all from my side.

src/Xrefcheck/CLI.hs
worm2fed: Wow, interesting. I didn't think about it. Thanks.

.xrefcheck.yaml
worm2fed: seems like we really do not need this file
I accidentally created it when tried to fix tests

src-files/def-config.yaml
```diff
@@ -39,6 +39,12 @@ verification:
   ignoreRefs:
     []
 
+  # Skip protected links
+  skipProtected: true
+
+  # Skip localhost links
+  skipLocalhost: true
+
```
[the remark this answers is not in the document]
worm2fed: @dcastro can we do it at another PR? It's becomes hard to keep history properly with fixups and conflicts due rebase..

Martoon-00 (approved):

src/Xrefcheck/CLI.hs
```diff
@@ -69,6 +73,17 @@ addTraversalOptions TraversalConfig{..} (TraversalOptions ignored) =
...
+newtype VerifyOptions = VerifyOptions
+  { voIgnoreLocalhost :: Bool
+  }
+
+addVerifyOptions :: VerifyConfig -> VerifyOptions -> VerifyConfig
+addVerifyOptions VerifyConfig{..} (VerifyOptions ignoreLocalhost) =
+  VerifyConfig
+  { vIgnoreLocalhost = ignoreLocalhost
+  , ..
+  }
```
dcastro, line 85: Hm there seems to be a bug here.


1. If we have a `.xrefcheck.yaml` config file with `ignoreLocalhost: false`
2. and then call xrefcheck without any command line options

`xrefcheck` will still ignore localhost links, even though we've told it not to.

This happens because this function is overriding the `ignoreLocalhost: false` from the config file with the value parsed from the CLI (`--check-localhost`, which defaults to true when omitted).

src-files/def-config.yaml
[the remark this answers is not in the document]
dcastro: Sure, can you open an issue for us to keep track of any outstanding work?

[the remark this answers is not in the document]
dcastro: Sure, can you open an issue for us to keep track of any outstanding work?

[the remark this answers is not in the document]
worm2fed: https://github.com/serokell/xrefcheck/issues/81

src/Xrefcheck/CLI.hs
worm2fed: I've change it to `switch = flag False True`, so now default value is False

dcastro: Well, now you've reversed the problem :sweat_smile: 

1. If we have a `.xrefcheck.yaml` config file with `ignoreLocalhost: true`
2. and then call xrefcheck without any command line options

`xrefcheck` will still check localhost links, even though we've told it not to.

The problem is that `addVerifyOptions` _always_ ignores the value from the config, and uses only the value from the command line.

dcastro: (btw, I do agree with changing `--check-localhost` to `--ignore-localhost`, this way the naming matches the config option, so let's keep that!)

worm2fed: > xrefcheck will still check localhost links, even though we've told it not to.

yes, but there are a question about priority - we have to define, which values more important for us - from config file or CLA. I think CLA has higher priority, because it requires some extra actions.

So, you have your config - and you just use it all the time. But than - there are situation where you pass some CLA - and you suppose, that your intention will be proceeded. 

What do you think?

dcastro: Yeah, I think CLA should have priority over config options. In the absence of a CLA, the config option should kick in.

Using my example from above, since no command line options were specified, and I had `ignoreLocalhost: true` in my config, I would have expected the config option to kick in.

dcastro: Let me put it another way.

The algorithm should be:

1. Use command line option if available
2. Otherwise, use config option if available
3. Otherwise, use some default value (in this case, I think the default is `ignoreLocalhost = false`)

However, with the current implementation:

1. Use command line option if available
2. Otherwise, use a default value of `false`

worm2fed: oh, I've got your point, you are right

worm2fed: @dcastro as I understand, 

> Use command line option if available
Otherwise, use config option if available
Otherwise, use some default value (in this case, I think the default is ignoreLocalhost = false)

we always have config option, because it doesn't wrapped with `Maybe`, so, I think the correct algorithm should be:
1. Use command line option if available
2. Otherwise, use config option

dcastro: @worm2fed Oh I didn't realize the config option was mandatory. Yup, that's correct then.

dcastro: By the way, the current implementation still doesn't work:

```
$ stack build xrefcheck:exe:xrefcheck && stack exec xrefcheck --  -r tests/markdowns/with-annotations/no_link.md                   
Missing: --ignore-localhost
```

From [`flag'`'s docs](https://hackage.haskell.org/package/optparse-applicative-0.16.1.0/docs/Options-Applicative-Builder.html#v:flag-39-):

> Builder for a flag parser without a default value.
> Same as flag, but with no default value. In particular, this flag will never parse successfully by itself.

At this point, I think we should do two things:

1. We should move the commits related to `ignoreAuthFailures` and `ignoreLocalhost` to a separate PR. They're holding up the merging of the FTP feature.
2. Looks like we could use some _golden tests_ to cover the use of command line options (in the new PR). These tests would call the `xrefcheck` executable just like a user would.

dcastro: Just noticed you've already created an issue specifically for the tests, #81.

Maybe we should amend the issue, so that it mentions both the implementation + the tests. I think we should also mention we need _golden_ tests.

worm2fed: @dcastro make sense

dcastro: @worm2fed please ping me when the PR has been split and I'll approve it ^^

worm2fed: @dcastro take a look at comments here please #81

dcastro: Cool, I'll have a look. In the meantime, can you remove those extra commits from this PR so we can merge it?

dcastro: Oh wait, I think I missed something. Do the FTP tests **require** the ability to check localhost links?

If so, does that mean that this PR depends on #85?

worm2fed: > Oh wait, I think I missed something. Do the FTP tests require the ability to check localhost links?

yes) to run ftp tests we need to configure behaviour for localhost links

worm2fed: > can you remove those extra commits from this PR 

I would like to do it when we'll merge #85 
cause at this commits there are some small changes I need to move to the main commit of this PR

dcastro: I see, ok, then I guess this PR is blocked until #85 is merged. Let's focus on #85 instead then :+1:

PR #68, opened 2021-07-13
[#62] Add anchor recognition in headers

### LINKED ISSUE #62: [BUG] Anchors in headers should be detected
# Description

Apparently, hand-written anchors within header names are just ignored at the moment.

# To Reproduce

Write something like

```
## <a name="stuff-section"></a> Some stuff

[My link](#stuff-section)
```

and run the check.

# Expected behavior

The check either passes, or there is an error discouraging putting anchors in headers in case this error is justified.

# Environment

- xrefcheck 0.1.3

Heimdell:
## Description

Problem:
Currently `xrefcheck` ignores HTML-anchors in headers.

Solution:
When traversing node tree, go into `HEADER` subnodes as well.

## Related issue(s)

Resolves #62 

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/tests/Test/Xrefcheck/AnchorsInHeadersSpec.hs
+++ b/tests/Test/Xrefcheck/AnchorsInHeadersSpec.hs
@@ -16,7 +16,7 @@ spec = do
     describe "Anchors in headers" $ do
         it "Check if anchors in headers are recognized" $ do
             fi <- getFI "tests/markdowns/without-annotations/anchors_in_headers.md"
-            getAnchors fi `shouldBe` ["-some-stuff", "stuff-section"]
+            getAnchors fi `shouldBe` ["some-stuff", "stuff-section"]
     where
         parse :: FilePath -> IO (Either Text FileInfo)
         parse path =
--- a/tests/Test/Xrefcheck/AnchorsInHeadersSpec.hs
+++ b/tests/Test/Xrefcheck/AnchorsInHeadersSpec.hs
@@ -13,8 +13,8 @@ import Xrefcheck.Scanners.Markdown
 
 spec :: Spec
 spec = do
-    describe "\"ignore link\" mode" $ do
-        it "Check \"ignore link\" performance" $ do
+    describe "Anchors in headers" $ do
+        it "Check if anchors in headers are recognized" $ do
             fi <- getFI "tests/markdowns/without-annotations/anchors_in_headers.md"
             getAnchors fi `shouldBe` ["-some-stuff", "stuff-section"]
     where
--- a/tests/Test/Xrefcheck/AnchorsInHeadersSpec.hs
+++ b/tests/Test/Xrefcheck/AnchorsInHeadersSpec.hs
@@ -1,0 +1,31 @@
+{- SPDX-FileCopyrightText: 2019 Serokell <https://serokell.io>
+ -
+ - SPDX-License-Identifier: MPL-2.0
+ -}
+
+module Test.Xrefcheck.AnchorsInHeadersSpec where
+
+import qualified Data.ByteString.Lazy as BSL
+import Test.Hspec (Spec, describe, it, shouldBe)
+
+import Xrefcheck.Core
+import Xrefcheck.Scanners.Markdown
+
+spec :: Spec
+spec = do
+    describe "\"ignore link\" mode" $ do
+        it "Check \"ignore link\" performance" $ do
+            fi <- getFI "tests/markdowns/without-annotations/anchors_in_headers.md"
+            getAnchors fi `shouldBe` ["-some-stuff", "stuff-section"]
+    where
+        parse :: FilePath -> IO (Either Text FileInfo)
+        parse path =
+            parseFileInfo defGithubMdConfig . decodeUtf8 <$> BSL.readFile path
+
+        getFI :: FilePath -> IO FileInfo
+        getFI path =
+            let errOrFI = parse path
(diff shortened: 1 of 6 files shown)


tests/markdowns/without-annotations/anchors_in_headers.md
```diff
@@ -0,0 +1,9 @@
+<!--
+ - SPDX-FileCopyrightText: 2018-2019 Serokell <https://serokell.io>
+ -
+ - SPDX-License-Identifier: MPL-2.0
+ -->
+
+## <a name="stuff-section"></a> Some stuff
+
+[My link](#stuff-section)
```
dcastro, line 9: It seems anchors in headers are no longer ignored, but the validation still doesn't seem to work correctly.

```
$ stack exec xrefcheck -- -r tests/markdowns/without-annotations/anchors_in_headers.md

  ➥  In file tests/markdowns/without-annotations/anchors_in_headers.md
     bad reference (local) at src:9:1-25:
       - text: "My link"
       - link: 
       - anchor: stuff-section
     
     ⛀  Anchor 'stuff-section' is not present
     
     
```

tests/Test/Xrefcheck/AnchorsInHeadersSpec.hs
```diff
@@ -0,0 +1,31 @@
...
+import qualified Data.ByteString.Lazy as BSL
+import Test.Hspec (Spec, describe, it, shouldBe)
+
+import Xrefcheck.Core
+import Xrefcheck.Scanners.Markdown
+
+spec :: Spec
+spec = do
+    describe "\"ignore link\" mode" $ do
+        it "Check \"ignore link\" performance" $ do
```
dcastro, line 17: These test names seem to be wrong

tests/markdowns/without-annotations/anchors_in_headers.md
dcastro: Nevermind, this is indeed working as expected :D

tests/Test/Xrefcheck/AnchorsInHeadersSpec.hs
```diff
@@ -0,0 +1,31 @@
...
+            fi <- getFI "tests/markdowns/without-annotations/anchors_in_headers.md"
+            getAnchors fi `shouldBe` ["-some-stuff", "stuff-section"]
```
dcastro, line 19:
```hs
"-some-stuff"
```

Where did the `-` come from? :thinking: 

I think we may have accidentally discovered another bug, because this is not considered valid (even though it should be):

```
## <a name="stuff-section"></a> Some stuff

[My link](#some-stuff)
```

```

  ➥  In file tests/markdowns/without-annotations/anchors_in_headers.md
     bad reference (local) at src:9:1-22:
       - text: "My link"
       - link: 
       - anchor: some-stuff
     
     ⛀  Anchor 'some-stuff' is not present,
        did you mean -some-stuff (header II) at src:7:1-42?
```

Conversely, this is now valid (even though it shouldn't be):

```
## <a name="stuff-section"></a> Some stuff

[My link](#-some-stuff)
```

And if we remove the anchor tag, then the `-` prefix disappears and this is considered valid, as expected:

```
## Some stuff

[My link](#some-stuff)
```

Can you please open an issue to address this please?

alyoanton9: > Where did the - come from? 🤔

It comes from `headerToAnchor` which was added to close [#55](https://github.com/serokell/xrefcheck/issues/55). In this function, all spaces in heading are replaced with "-" to produce an anchor.

AFAIU when `## <a name="stuff-section"></a> Some stuff` is being parsed, " Some stuff" is considered as a heading instead of "Some stuff". When it's only the heading without HTML code, everything is okay. I tried to remove the space `## <a name="stuff-section"></a>Some stuff` and "some-stuff" became a valid anchor.

The first probable fix that comes to my mind is stripping the text in `nodeExtractText `:
```hs
nodeExtractText :: Node -> Text
nodeExtractText = T.strip . mconcat . map extractText . nodeFlatten
...
```
But perhaps it's a hack and there could be another solution.

dcastro: Ah that makes sense!

Yeah, I think, the `strip` solution makes sense, it seems to be what github does.

tests/Test/Xrefcheck/AnchorsInHeadersSpec.hs
alyoanton9: ➕ 

It seems these 2 strings were forgotten to change

tests/Test/Xrefcheck/AnchorsInHeadersSpec.hs
alyoanton9: Cool!

It just seems strange to me that `CMark` doesn't perform it by default.

dcastro (approved):

alyoanton9 (approved):

dcastro:
@Heimdell Just noticed you bumped the package version and added a new section in the changelog.

Can you please revert this?

The version is only supposed to be bumped when a release is made (see [an example of a release procedure here](https://github.com/serokell/xrefcheck/pull/65)), and this is only periodically (not whenever a PR is merged).

Your changelog entry should be under the `Unreleased` section.

Issue #64, opened 2021-04-23
Try to implement copy-paste protection checks


Martoon-00:
# Clarification and motivation

Imagine the following list of links:

* `[file](files/file.out)`
* `[Another file](files/another-file.out)`
* `[And a file once again](files/and-a-file-once-again.out)`

It is easy to make a mistake here during copy-pasting so that text is updated and the link is not. I think we can use some heuristics to spot such mistakes (but avoid false positives at all costs):
* If this check is enabled in config (it should be by default);
* And there are two links `[A1](B1)` and `[A2](B1)` in a file, and `A1` is equal to `B1` modulo casing and all the non-alphanum characters, while `A2` is _not_ equal to `B1`

then report an error at `[A2](B1)` position. And a similar check for `[A1](B2)`.

# Acceptance criteria

* The check is implemented.
* It can be disabled both in config and in place via some `<--! xrefcheck: no duplication check in {file/paragraph/link} -->`.

Martoon-00:
If possible, we should check reference-style links (`[text][link-id]`) too.

However not sure how much is it possible, AFAIR such links are automatically inlined by our markdown parser.

YuriRomanowski:
Do we want to check only links within one list? How about checking all the links within one file?

Martoon-00:
That's a good question. On the one hand, this increases the probability of getting a false positive. On the other, checking through the entire file may be more useful and will be a more transparent behaviour for the user.

Let's really go with checking across the entire file.

Over time we will collect some statistics on how this check works on real-life repositories and will revise the behaviour then.

PR #71, opened 2021-07-29
Refactor data collection.

Heimdell:
## Description

Problem: data collection is a function that calls itself 5 screens below the place it starts and is a pile of different, but unrelated actions, performed simultaneously.

Solution: divide the pile into parts and then combine parts externally.

### Concrete changes

1) The `xrefcheck: ignore X` functionality is implemented as a separate function.
3) The data collection is a fold over the tree, and other reducers can be connected.

### Results

`Anchor`/`Reference` colelctor fits into one screen now.

### The Visitor API

```haskell
foldNode :: (Monoid a, Monad m) => (Node -> m a) -> Node -> m a
```
Runs the provided action, collects the results.

## Related issue(s)

None, preparing for #64 

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/exec/Main.hs
+++ b/exec/Main.hs
@@ -7,7 +7,7 @@ module Main where
 
 import Universum
 
-import qualified Data.ByteString as BS
+import Data.ByteString qualified as BS
 import Main.Utf8 (withUtf8)
 
 import Xrefcheck.CLI (Command (..), getCommand)
--- a/exec/Main.hs
+++ b/exec/Main.hs
@@ -5,6 +5,8 @@
 
 module Main where
 
+import Universum
+
 import qualified Data.ByteString as BS
 import Main.Utf8 (withUtf8)
 
--- a/package.yaml
+++ b/package.yaml
@@ -35,6 +35,7 @@ default-extensions:
   - MultiParamTypeClasses
   - MultiWayIf
   - NamedFieldPuns
+  - NoImplicitPrelude
   - OverloadedStrings
   - RankNTypes
   - RecordWildCards
@@ -57,9 +58,7 @@ dependencies:
   - aeson
   - aeson-casing
   - async
-  - name: base
-    mixin: [hiding (Prelude)]
-    version: "< 4.15"
+  - base
   - bytestring
   - containers
   - cmark-gfm
@@ -91,8 +90,7 @@ dependencies:
   - th-lift-instances
   - th-utilities
   - transformers
-  - name: universum
-    mixin: [(Universum as Prelude), (Universum.Unsafe as Unsafe)]
+  - universum
   - yaml
   - with-utf8
 
--- a/package.yaml
+++ b/package.yaml
@@ -27,6 +27,7 @@ default-extensions:
   - DefaultSignatures
   - DeriveDataTypeable
   - DeriveGeneric
+  - DerivingStrategies
   - FlexibleContexts
   - FlexibleInstances
   - FunctionalDependencies
@@ -67,6 +68,7 @@ dependencies:
   - deepseq
   - directory-tree
   - directory
+  - dlist
   - filepath
   - file-embed
   - fmt
--- a/package.yaml
+++ b/package.yaml
@@ -31,6 +31,7 @@ default-extensions:
   - FlexibleInstances
   - FunctionalDependencies
   - GeneralizedNewtypeDeriving
+  - ImportQualifiedPost
   - LambdaCase
   - MultiParamTypeClasses
   - MultiWayIf
--- a/src/Xrefcheck/Scanners/Markdown.hs
+++ b/src/Xrefcheck/Scanners/Markdown.hs
@@ -15,6 +15,8 @@ module Xrefcheck.Scanners.Markdown
   , parseFileInfo
   ) where
 
+import Universum
+
 import CMarkGFM (Node (..), NodeType (..), PosInfo (..), commonmarkToNode)
 import Control.Lens ((%=))
 import Control.Monad.Trans.Except (Except, runExcept, throwE)
--- a/src/Xrefcheck/Scanners/Markdown.hs
+++ b/src/Xrefcheck/Scanners/Markdown.hs
@@ -21,11 +21,11 @@ import CMarkGFM (Node (..), NodeType (..), PosInfo (..), commonmarkToNode)
 import Control.Lens ((%=))
 import Control.Monad.Trans.Except (Except, runExcept, throwE)
 import Data.Aeson.TH (deriveFromJSON)
-import qualified Data.ByteString.Lazy as BSL
+import Data.ByteString.Lazy qualified as BSL
 import Data.Char (isSpace)
 import Data.Default (Default (..))
-import qualified Data.Text as T
-import qualified Data.Text.Lazy as LT
+import Data.Text qualified as T
+import Data.Text.Lazy qualified as LT
 import Fmt (Buildable (..), blockListF, nameF, (+|), (|+))
 
 import Xrefcheck.Core
--- a/src/Xrefcheck/Scanners/Markdown.hs
+++ b/src/Xrefcheck/Scanners/Markdown.hs
@@ -18,12 +18,9 @@ module Xrefcheck.Scanners.Markdown
 import Universum
 
 import CMarkGFM (Node (..), NodeType (..), PosInfo (..), commonmarkToNode)
-import Control.Lens ((%=))
-import Control.Monad.Trans.Except (Except, runExcept, throwE)
+import Control.Monad.Except (MonadError, throwError)
 import Data.Aeson.TH (deriveFromJSON)
 import Data.ByteString.Lazy qualified as BSL
-import Data.Char (isSpace)
-import Data.Default (Default (..))
 import Data.Text qualified as T
 import Data.Text.Lazy qualified as LT
 import Fmt (Buildable (..), blockListF, nameF, (+|), (|+))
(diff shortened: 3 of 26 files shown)


src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -15,15 +15,17 @@ module Xrefcheck.Scanners.Markdown
...
 import CMarkGFM (Node (..), NodeType (..), PosInfo (..), commonmarkToNode)
 import Control.Lens ((%=))
 import Control.Monad.Trans.Except (Except, runExcept, throwE)
 import Data.Aeson.TH (deriveFromJSON)
-import qualified Data.ByteString.Lazy as BSL
+import Data.ByteString.Lazy qualified as BSL
 import Data.Char (isSpace)
 import Data.Default (Default (..))
-import qualified Data.Text as T
-import qualified Data.Text.Lazy as LT
```
worm2fed, line 26: This is not accepted yet) but I do not mind
Can you 'vote' here?
https://github.com/serokell/style/pull/11

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -44,6 +46,11 @@ defGithubMdConfig = MarkdownConfig
 instance Buildable Node where
   build (Node _mpos ty subs) = nameF (show ty) $ blockListF subs
 
+-- foldWithSpine :: Monoid a => (Node -> a) -> Node -> a
```
worm2fed: I don't like this silent change. This commit is about removing mixing, but here you commented function for unknown reason

exec/Main.hs
```diff
@@ -7,7 +7,7 @@ module Main where
 
 import Universum
 
-import qualified Data.ByteString as BS
+import Data.ByteString qualified as BS
```
worm2fed, line 10: I can see that you make such a changes in your first commit and at the second one
Please, make smth with it - move this changes from first commit to this one or maybe it'll be ok to squash this two commits (but in this case you need update commit title - and provide commit body! with description of your changes)

src/Xrefcheck/Scanners/Markdown/ErrorMessage.hs
```diff
@@ -1,3 +1,7 @@
+{- SPDX-FileCopyrightText: 2018-2019 Serokell <https://serokell.io>
```
worm2fed, line 1: it's 2021 now))
same in other headers you've added

tests/Main.hs
```diff
@@ -3,6 +3,8 @@
  - SPDX-License-Identifier: MPL-2.0
  -}
 
+import Universum
```
worm2fed, line 6: This changes supposed to be at the first commit...

worm2fed (changes requested):
Great PR description! but so dirty commit history..
Can you do something with it? It's become really hard to review your changes.

1. Go through your commits and decide which small steps to your final result this commits lead you.
2. Add description for each commit (as described in our policy) so we can review this steps separately and understand it.
3. Use `git commit --fixup HASH` to fix smth in your previous commits, so we can understand to which commits related your fixes.

src/Xrefcheck/Scanners/Markdown.hs
Heimdell: I use `ImportQualifiedPost` that mitigates the problem anyway.

src/Xrefcheck/Scanners/Markdown/ErrorMessage.hs
```diff
@@ -0,0 +1,51 @@
...
+  , ""
+  ]
+
+unrecognisedMsg :: Text -> Text
+unrecognisedMsg txt = unwords
+  [ "unrecognised option"
+  , "\"" <> txt <> "\""
+  , "perhaps you meant \
+      \<\"ignore link\"|\"ignore paragraph\"|\"ignore file\"> "
+  ]
```
Martoon-00, line 37: I think this is not a good thing actually. Putting error messages far from the use site, we encourage their divergence from  the actual problem (e.g. the actual error can obtain one more condition when it fires, but it is now easy to forget to update the error message respectively).

I think error messages and where they are used should fit into screen, maybe but in `where`/`letl` block if found to make the code less readable.

Martoon-00 (commented):
My opinion, per change:

1. Using post-imports - kaef (though I wonder, didn't @worm2fed implemented this already somewhere else, taking https://github.com/serokell/style/pull/11 into account?)
1.5. Mixins removal - good; `import Universum` addition - ok.
2. Using `polysemy` - I would retract from doing this change until it actually appears necessary, because using it makes things more complex for newcomers, and there might be cases not well supported by it (as with any advanced framework).
3. Using the Visitor pattern - the same thing, the current Markdown scanner already does most of the things we ever want from it and it just works. After the rewrite, we use only `inside` event, so things just got more complex without immediate benefit. And it's not clear whether future changes in the scanner logic, if any, will fit into the current model without increasing its complexity in directions other than that you support now.
4. Error messages extraction to a separate module - responded in another comment. The fact of extraction is good, but error message and condition for it to fire should locate nearby IMO (they didn't before, but putting to a separate module sounds too harsh).
5. Global `-wno-orphans` - this sounds like a coolstory, I think orhpans are not a very good pattern so discouraging it with a warning is actually good. Or not? Also I heard that compilation time slightly increases with each module that allows orphan instances.
6. Trailing `--` in haddock - why not just use the same style as in the remainder of the project :pepe_stare:  If everyone comes with its own preferred style (that also probably changes every several years), that would be a mess.
7. The separation of ignored removal and info gathering itself - really nice to obtain it :+1: 
8. Moving functions from `where` to top-level - neat.

Well, quite a wall of comments, but this PR is sorta big too )

I would extract 1 & 1.5, maybe 4 (changed), 7 and 8 to a separate PR and leave other changes pending for now.

worm2fed:
@Martoon-00 
> Using post-imports - kaef (though I wonder, didn't @worm2fed implemented this already somewhere else, taking Add requirement for qualified imports style#11 into account?)

We didn't merge it yet. I'll update my PR soon!)

Martoon-00:
> @Martoon-00
> 
> > Using post-imports - kaef (though I wonder, didn't @worm2fed implemented this already somewhere else, taking Add requirement for qualified imports style#11 into account?)
> 
> We didn't merge it yet. I'll update my PR soon!)

:+1: 

Just coordinate with @Heimdell who switches this entire project to post-qualified-imports.

package.yaml
```diff
@@ -27,14 +27,17 @@ default-extensions:
   - DefaultSignatures
   - DeriveDataTypeable
   - DeriveGeneric
+  - DerivingStrategies
   - FlexibleContexts
   - FlexibleInstances
   - FunctionalDependencies
   - GeneralizedNewtypeDeriving
+  - ImportQualifiedPost
```
dcastro, line 35: Now stylish-haskell automatically rewrites `import A qualified as B` to `import qualified A as B`.

You can verify this behaviour by manually running this:

```
stylish-haskell --inplace src/Xrefcheck/Scanners/Markdown.hs
```

If we wanna use `ImportQualifiedPost`, we need to configure stylish-haskell to keep the "ImportQualifiedPost" format.

package.yaml
```diff
@@ -27,6 +27,7 @@ default-extensions:
   - DefaultSignatures
   - DeriveDataTypeable
   - DeriveGeneric
+  - DerivingStrategies
```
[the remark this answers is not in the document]
dcastro: Let's keep `DerivingStrategies`. In fact, we should always explicitly specify which strategy we're using. Implicit strategies are [a bad practice](https://kowainik.github.io/posts/deriving#anyclass-ambiguity).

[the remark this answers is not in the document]
dcastro: We should use `-Weverything` (like we do in morley) or, at the very least, `-Wmissing-deriving-strategies` to force us to specify a deriving strategy. Let's do this in a separate PR. @Heimdell can you please create an issue for this?

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -66,143 +65,153 @@ nodeExtractText = T.strip . mconcat . map extractText . nodeFlatten
...
+      case ty of
+        HTML_BLOCK _ -> do
+          return mempty
+
+        HEADING lvl -> do
+          flavor <- asks mcFlavor
+          let aType = HeaderAnchor lvl
+          let aName = headerToAnchor flavor $ nodeExtractText node
+          let aPos  = toPosition pos
+          return $ FileInfoDiff mempty $ pure $ Anchor {aType, aName, aPos}
```
dcastro, line 149: Personal suggestion: I think using the monomorphic `DList.empty` and `DList.singleton` (in this branch and the others) makes the code easier to read than the polymorphic `mempty` and `pure`.

Unnecessary polymorphism can sometimes hinder readability.

```suggestion
          return $ FileInfoDiff DList.empty $ DList.singleton $ Anchor {aType, aName, aPos}
```

This is arguable and subjective, so I'll leave this up to you :smile:

package.yaml
Heimdell: image

Google finds 0 results though. I'll try to do it, but can't guarantee it is possible.

Heimdell: Disregard that, it is possible https://github.com/akrmn/stylish-haskell/commit/c5c7b50445e930389a22e303ee0d84679af8a93e#diff-0924d0340532d02b8c3bacd84f133f5c6da1bf2494e872247f3abfc9e1c8d4b1

dcastro (approved):

dcastro:
@Martoon-00 do you wanna have another look at this before merging, or are you happy enough with merging as-is?

PR #80, opened 2021-09-01
[#72] Fix anchor generation of headers with spaces

### LINKED ISSUE #72: [BUG] Header anchors are prefixed with a dash
# Description

Say we have a header such as:

```md
## <a name="link-one"></a> Link two
```

This markdown is supposed to generate two anchors: `link-one` and `link-two`.

However, it looks like xrefcheck thinks the second anchor is `-link-two` (and not `link-two`).

See @alyoanton9's comment for a description of the issue and solution: https://github.com/serokell/xrefcheck/pull/68#discussion_r676461656

# To Reproduce

Create a markdown file with the following content, and then run `xrefcheck -r <filename>`

```md
## <a name="stuff-section"></a> Some stuff

[My link](#some-stuff)
```

# Expected behavior

`xrefcheck` should succeed.

# Actual behavior

`xrefcheck` fails:

```

  ➥  In file tests/markdowns/without-annotations/anchors_in_headers.md
     bad reference (local) at src:9:1-22:
       - text: "My link"
       - link: 
       - anchor: some-stuff
     
     ⛀  Anchor 'some-stuff' is not present,
        did you mean -some-stuff (header II) at src:7:1-42?
```

# Environment

- branch: slowpnir/anchors-in-headers

breakerzirconia:
## Description

Problem: 
A header name with its custom anchor appended after the octothorpe symbol(s) is parsed with the leading spaces present, thus generating an anchor from it will result in having a leading hyphen prepended to its beginning.

Solution: 
Having retrieved the header name, strip it to eradicate the leading spaces. This will result in the correct generation of the anchor which won't contain a hyphen at its beginning.

## Related issue(s)

Resolves #72 by applying `Data.Text.strip` to the extracted name of either a link or a header in `nodeExtractText`.

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/tests/Test/Xrefcheck/AnchorsSpec.hs
+++ b/tests/Test/Xrefcheck/AnchorsSpec.hs
@@ -5,18 +5,27 @@
 
 module Test.Xrefcheck.AnchorsSpec where
 
-import Test.Hspec (Spec, describe, it)
+import Test.Hspec (Spec, describe, it, shouldBe)
 import Test.QuickCheck ((===))
 
-import Xrefcheck.Core (Flavor (..), headerToAnchor)
+import Test.Xrefcheck.Util
+import Xrefcheck.Core
 
 checkHeaderConversions
   :: HasCallStack
   => Flavor -> [(Text, Text)] -> Spec
 checkHeaderConversions fl suites =
-  describe (show fl) $
+  describe (show fl) $ do
     forM_ suites $ \(a, b) ->
       it (show a <> " == " <> show b) $ headerToAnchor fl a === b
+    it "Non-stripped header name should be stripped" $ do
+      fi <- getFI fl "tests/markdowns/without-annotations/non_stripped_spaces.md"
+      getAnchors fi `shouldBe` [ case fl of GitHub -> "header--with-leading-spaces"
+                                            GitLab -> "header-with-leading-spaces"
+                               ]
+  where
+    getAnchors :: FileInfo -> [Text]
+    getAnchors fi = map aName $ fi ^. fiAnchors
 
 spec :: Spec
 spec = do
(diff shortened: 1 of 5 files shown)


tests/Test/Xrefcheck/AnchorsSpec.hs
```diff
@@ -5,18 +5,37 @@
...
+                               ]
+  where
+    parse :: FilePath -> IO (Either Text FileInfo)
+    parse path =
+      parseFileInfo MarkdownConfig{ mcFlavor = fl } . decodeUtf8 <$> BSL.readFile path
+
+    getFI :: FilePath -> IO FileInfo
+    getFI path =
+      let errOrFI = parse path
+      in either error id <$> errOrFI
```
Martoon-00, line 35: Oh, I suspect that this is copied from another place.

Could you please create an `Util` module in tests and place this reading function there?

Martoon-00 (approved):
Looks good to me :+1: 

Just left one minor comment.

tests/Test/Xrefcheck/AnchorsSpec.hs
breakerzirconia: Absolutely.

breakerzirconia: Hmm, on another note, the definitions of `parse` slightly differ in the respective test modules: in `AnchorSpec` `parse` depends on the `Flavor` in the local scope, but in `IgnoreAnnotationSpec` it doesn't and uses the default GitHub configuration. May I use the default config in `AnchorSpec` as well (for the test's primary purpose is to ensure that there are no leading hyphens, which works for both `Flavor`s) to then put `parse`, `getAnchors` and `getFI` in the `Util` module?

Martoon-00: I personally would make the function accept `Flavor` explicitly because generally there is no sense of "default" flavor. 

The one who writes his own test should decide, whether can he test that on just one arbitrary flavor or try every possible flavor, and it's good if we force him to take this decision.

dcastro (approved):
LGTM, well done! :+1:

PR #82, opened 2021-09-07
[#75] Fix the root with an appended slash support

### LINKED ISSUE #75: [BUG] `--root` does not accept paths ending with a slash
# Description

The `--root` command-line option does not accept paths ending with a slash.
It feels like it should.

# To Reproduce

1. Create a folder called `demo` with 2 files, `a.md` and `b.md`.
2. The file `a.md` contains a single line: 
      ```md
      [A reference](b.md)
      ```
3. Run these commands:
      ```
      stack build xrefcheck:exe:xrefcheck && stack exec xrefcheck --  --root ./demo
      stack build xrefcheck:exe:xrefcheck && stack exec xrefcheck --  --root ./demo/
      ```

# Expected behavior

Expected both commands to pass.

# Actual behavior

The first command passes (as expected), the second one fails:

```
$ stack build xrefcheck:exe:xrefcheck && stack exec xrefcheck --  --root ./demo

$ stack build xrefcheck:exe:xrefcheck && stack exec xrefcheck --  --root ./demo/

  ➥  In file ./demo/demo/a.md
     bad reference (relative) at src:1:1-19:
       - text: "A reference"
       - link: b.md
       - anchor: -
     
     ⛀  File does not exist:
        ./demo/demo/b.md
     
     
```

Notice _both_ paths in the error message, they contain `demo` twice: `./demo/demo/a.md` and `./demo/demo/b.md`

# Environment

- 7648d7c21efcaa57e5dd7853a7fc0b6e73f076b7, master

breakerzirconia:
## Description

Problem:
The recursive repository analysis always contains invalid references if given the root with a trailing forward slash. Suppose the root is `./a/b/`; then, after having gathered the repository information, the verification process will return a non-empty list of errors, because the lowest level directory will be duplicated (i.e. `./a/b/b/`). thus changing the root where the files may or may not exist.

Solution: 
The core of the issue lies within the definition of the `System.Directory.Posix.takeDirectory` function: if its argument — the file path — does indeed have a trailing slash, the function will simply **eradicate the slash**, when in our case what's expected is to **have the lowest level directory taken out**. The solution to that issue is to simply check whether the root contains a slash at its end, and if it does — remove it.

## Related issue(s)

Resolves #75 by modifying the definition of `dirOfRoot` in the `Xrefcheck.Scan` module to consider the case of having the forward slash character appended to the root.

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and


src/Xrefcheck/Scan.hs
```diff
@@ -61,12 +63,15 @@ gatherRepoInfo rw formatsSupport config root = do
...
-        $ Tree.zipPaths . (dirOfRoot Tree.:/)
+        $ Tree.zipPaths . (dirOfRoot root Tree.:/)
         $ filterExcludedDirs root repoTree
   return $ RepoInfo (M.fromList fileInfos)
   where
     rootNE = if null root then "." else root
-    dirOfRoot = if root == "" || root == "." then "" else takeDirectory root
+    dirOfRoot rt
+      | rt == "" || rt == "." = ""
+      | Unsafe.last rt == '/' = dirOfRoot $ takeDirectory rt
```
Martoon-00, line 73: I would use `Data.List.uncons` - it is simpler/quicker than `takeDirectory`, and safer than `Unsafe.last` (indeed using `Unsafe.last` is safe here, but it is always better to rely on compiler in ensuring correctness, this makes future edits less fragile).

Moreover, correct me if I'm wrong, but looks like a bug lurks here: according to the documentation, `takeDirectory "/"` is `"/"`, which means that passing `"/"` as a root would cause an infinite recursion.

tests/Test/Xrefcheck/TrailingSlashSpec.hs
```diff
@@ -0,0 +1,45 @@
...
+    forM_ roots $ \root -> do
+      repoInfo <- runIO $ allowRewrite False $ \rw ->
+        gatherRepoInfo rw format TraversalConfig{ tcIgnored = [] } root
+      res <- runIO $ allowRewrite False $ \rw ->
+        verifyRepo rw (cVerification config) FullMode root repoInfo
+      it ("Analysis results for the root with" <>
+          (if Unsafe.last root == '/' then "" else "out") <>
+          " the trailing forward slash should not contain the FileDoesNotExist error") $ do
+        let satisfactoryCase = maybe True (not . any checkFileDoesNotExist) $ verifyErrors res
+        satisfactoryCase === True
```
Martoon-00, line 34: Maybe it would be enough to check here that `repoInfo` contains something? This way the test would shorter, and would not fail when verification logic went wrong.

dcastro, line 34: Comparing two bools, like `bool === True`, usually leads to less-than-ideal error messages.

For example, if we expected files `a`, `b` and `c` to exist, but `b` didn't exist, the error message in the test logs will simply say:

> `Expected 'True', got 'False'`.

Ideally, the error message should say something like: 

> Expected all filepaths to be valid, but this filepath does not exist: b

tests/Test/Xrefcheck/TrailingSlashSpec.hs
```diff
@@ -0,0 +1,45 @@
...
+    checkFileDoesNotExist :: WithReferenceLoc VerifyError -> Bool
+    checkFileDoesNotExist WithReferenceLoc{..} = case wrlItem of
+      FileDoesNotExist _ -> True
+      _ -> False
+
+    roots :: [String]
+    roots =
+      [ "tests/markdowns/without-annotations"
+      , "tests/markdowns/without-annotations/"
+      ]
```
dcastro, line 45: These tests seem to assume that all files in the `tests/markdowns/without-annotation` folder are valid, and thus `verifyRepo` should pass.

That's a valid assumption _at the moment_, but the point of these folders is to contain some sample markdown files we want to test which may or may not be valid.

A solution might be to _not_ call `verifyRepo`. We could simply check that all the filepaths in `repoInfo` are indeed valid, i.e., they exists on disk.

dcastro: I see @Martoon-00 already made a similar point above :smile:

src/Xrefcheck/Scan.hs
```diff
@@ -61,12 +63,15 @@ gatherRepoInfo rw formatsSupport config root = do
   _ Tree.:/ repoTree <- liftIO $ Tree.readDirectoryWithL processFile rootNE
   let fileInfos = filter (\(path, _) -> not $ isIgnored path)
         $ dropSndMaybes . F.toList
-        $ Tree.zipPaths . (dirOfRoot Tree.:/)
+        $ Tree.zipPaths . (dirOfRoot root Tree.:/)
```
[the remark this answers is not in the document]
dcastro: Also, we should probably add tests for some of these edge cases as well, e.g.:
* `tests/markdowns/without-annotations/./`
* `tests/markdowns/without-annotations/all_checked.md`
* `tests/markdowns/without-annotations/./all_checked.md`
* `.`
* etc

tests/Test/Xrefcheck/TrailingSlashSpec.hs
dcastro: I stand corrected, it seems the tests make no such assumption. But still, it would be simpler to just check that the files from `repoInfo` exist on disk.

src/Xrefcheck/Scan.hs
breakerzirconia: Oh, you're right, this definition of `dirOfRoot` does indeed produce an infinite loop on `"/"`! I'll read the rest of the comments and think about how to redefine the function.

tests/Test/Xrefcheck/TrailingSlashSpec.hs
breakerzirconia: I've actually tried basing my observations only on the results of `gatherRepoInfo`, but it turns out, regardless of the appearance of the trailing forward slash in the root, `gatherRepoInfo` would perform just fine! As far as I understand, such behavior is present due to the file-wise analysis being performed **before** the composition 
`Tree.zipPaths . (dirOfRoot root Tree.:/)` is called. Because of that, I thought the actual verification with `verifyRepo` was the way to go in this situation.

dcastro: (We could use [`doesFileExist`](https://hackage.haskell.org/package/directory-1.3.7.0/docs/System-Directory.html#v:doesFileExist) for this)

breakerzirconia: Yeah, that's it, given the correct `root` where files did indeed exist, `Map FilePath FileInfo` would be non-empty, and the very problem here lied within the keys of that map, i.e. the `FilePath`s! I think I understand now, `verifyRepo` is indeed a redundant operation, and the only thing needed was to check whether these `FilePath`s do or do not exist.

Issue #74, opened 2021-08-17
Avoid checking duplicates


dcastro:
# Clarification and motivation

Say we have a file that contains multiple references to the same link:

```md
Some text and [a link](http://google.com).
More text and [another link](http://google.com).
```

In situations like these, it would be wasteful to check the same link twice.

We should avoid verifying links we've already checked..

If one reference is broken, `xrefcheck` should still report _both_ lines, individually.

PS: This, of course, also applies to when we have multiple files referencing the same link.

# Acceptance criteria

* When a link appears twice or more in a file (or a set of files), we only verify the link once.
* When the link is invalid, we still report each reference individually.

PR #83, opened 2021-09-09
[#74] Detect duplicating links during verification

breakerzirconia:
## Description

Problem: All the duplicating external links get verified independently, which is wasteful.

Solution: Store the verification results in the map. Verify the link once if it hasn't been already, and insert the results in the map, and if the link has been verified before, return the results by retrieving the respective value from the map.

## Related issue(s)

Implements #74 by storing previously verified links in a map.

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and


src/Xrefcheck/Verify.hs
```diff
@@ -155,14 +156,15 @@ verifyRepo
   let toScan = do
         (file, fileInfo) <- M.toList repoInfo
         guard . not $ any ((`isPrefixOf` file) . (root </>)) vcNotScanned
-        ref <- _fiReferences fileInfo
+        ref <- L.nubBy (\r1 r2 -> rLink r1 == rLink r2) $ _fiReferences fileInfo
```
dcastro, line 159: This means that if we have 2 references to the same link, and that link is broken, `xrefcheck` will only report 1 broken reference.

E.g.:

```md
[a](https://non-existent.reference/)
[b](https://non-existent.reference/)
```

```

  ➥  In file tests/markdowns/without-annotations/test.md
     bad reference (external) at src:7:1-36:
       - text: "a"
       - link: https://non-existent.reference/
       - anchor: -
     
     ⛂  ConnectionFailure Network.Socket.getAddrInfo (called with preferred socket type/protocol: AddrInfo {addrFlags = [AI_ADDRCONFIG], addrFamily = AF_UNSPEC, addrSocketType = Stream, addrProtocol = 0, addrAddress = <assumed to be undefined>, addrCanonName = <assumed to be undefined>}, host name: Just "non-existent.reference", service name: Just "443"): does not exist (Name or service not known)
```

We should report both references though (I've edited the issue's description to make this more clear).

src/Xrefcheck/Verify.hs
```diff
@@ -269,14 +273,24 @@ verifyReference
...
+checkExternalResource VerifyConfig{..} externalCheckedRef link
   | isIgnored = return mempty
   | doesReferLocalhost = return mempty
   | otherwise = fmap toVerifyRes $ do
-      makeRequest HEAD 0.3 >>= \case
-        Right () -> return $ Right ()
-        Left   _ -> makeRequest GET 0.7
+      externalChecked <- readIORef externalCheckedRef
+      if M.member link externalChecked
+      then return $ externalChecked M.! link
```
dcastro, line 287: `!` is a partial function, we should avoid it.

We should use `lookup` instead of `member`+`!`

src/Xrefcheck/Verify.hs
```diff
@@ -269,14 +273,24 @@ verifyReference
...
+      else do
+        result <- makeRequest HEAD 0.3 >>= \case
+          Right () -> return $ Right ()
+          Left   _ -> makeRequest GET 0.7
+        writeIORef externalCheckedRef (M.insert link result externalChecked)
+        return result
```
dcastro, line 293: This logic does not guarantee that a link won't be checked twice by two threads.

For example, imagine we have two references `r1` and `r2`, pointing to the same link `l`.
Now consider this sequence of events:
1. `verifyRepo` spawns two threads, `t1` and `t2`. `t1` will check `r1`, and `t2` will check `r2`.
2. `t1` uses `readIORef` to read the map, and checks whether the map contains `l`. It doesn't, so it proceeds with the check.
3. `t2` uses `readIORef` to read the map, and checks whether the map contains `l`. It doesn't, so it proceeds with the check.
4. `t1` finishes and writes the result to the `IORef`
5. `t2` finishes and writes the result to the `IORef`

dcastro: There's actually another bug here:

Imagine we have two references `r1` and `r2`, pointing to different links `l1` and `l2`.
Now consider this sequence of events:
1. `verifyRepo` spawns two threads, `t1` and `t2`. `t1` will check `r1`, and `t2` will check `r2`.
2. `t1` uses `readIORef` to read the **empty** map, and checks whether the map contains `l1`. It doesn't, so it proceeds with the check.
3. `t2` uses `readIORef` to read the **empty** map, and checks whether the map contains `l2`. It doesn't, so it proceeds with the check.
4. `t1` finishes, modifieds the **empty** map by adding an entry for `l1`, and writes the result to the `IORef` using `writeIORef`
5. `t2` finishes, modifieds the **empty** map by adding an entry for `l2`, and writes the result to the `IORef` using `writeIORef`

In the end, the IORef might contain a map with only 1 entry (instead of 2).

The problem here is that different thread's `read` and `write` operations might overlap, so they'll overwrite each other's work.

Martoon-00: My 50 cents :sweat_smile: : I think it would be helpful if you write down a design of the proper solution. It is likely simple but may need some relatively advanced knowledge to come to.

(I can think over it if you agree but are currently busy with other stuff)

dcastro: > are currently busy with other stuff

Yeah same :sweat_smile: 

>  I think it would be helpful if you write down a design of the proper solution. It is likely simple but may need some relatively advanced knowledge to come to.

Agreed! @breakerzirconia please try to come up with a design (maybe looking into IORef's `atomic*` operations and `MVar` _may_ be helpful), we can continue this discussion in Slack, feel free to ask any questions at all. When we agree on a plan, then we can implement it ^^

Martoon-00:
Btw, we tend to avoid merge commits at all costs. Always prefer rebase instead.

In some cases, if rebasing all the pack of commits is inconvenient, it is fine to postpone rebase until getting approvals and squashing fixup commits.

src/Xrefcheck/Verify.hs
```diff
@@ -269,14 +272,27 @@ verifyReference
...
+      case link `M.lookup` externalChecked of
+        Just resultMVar -> putMVar externalCheckedMVar externalChecked >> readMVar resultMVar
+        Nothing -> do
+          resultMVar <- newEmptyMVar
+          let externalChecked' = M.insert link resultMVar externalChecked
+          putMVar externalCheckedMVar externalChecked'
+          result <- makeRequest HEAD 0.3 >>= \case
+            Right () -> return $ Right ()
+            Left   _ -> makeRequest GET 0.7
+          putMVar resultMVar result
```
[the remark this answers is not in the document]
dcastro: :pogchamp: Yeah I loved that idea!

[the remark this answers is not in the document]
breakerzirconia: I've understood the logic behind your proposal, but I'm also looking at the pseudocode and can't wrap my head around how to include `verifyReference` itself in the definition of `verifyReferenceRec`... `cachedMap` is used only to store data collected from verifying external links, and the location type of the `Reference` is deduced in `verifyReference`, so that we won't be needing to access the map or insert anything into it if the reference is not an external resource. The fact that the lookup and the insertion operations are performed in `verifyReferenceRec` throws me off a bit. 

Am I correct to assume that, since we're basically redefining the whole traversal logic of the list `[(FilePath, Reference)]` in terms of the recursive function `verifyReferenceRec`, and not via `forConcurrently`, it is also sufficient to redefine **`checkExternalResource`**, where the whole `MVar` business occured, _only_ in terms of the primitive `Map` functions (i.e. no asynchronous actions in the aforementioned function are needed)?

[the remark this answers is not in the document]
Martoon-00: Oh right, that's a fair concern.

How about this: decoupling the concurrency/caching logic from the main business logic still seems good unless it gets too complex. 

What if we make our new re-implementation of `forConcurrently` allow optional caching - the map is checked and updated only for external links? 

If we implement this via a generic method, it will have the following signature:

```hs
forConcurrentlyCaching :: Ord cacheKey => [a] -> (a -> NeedsCaching cacheKey) -> (a -> IO b) -> [b]

data NeedsCaching key
  = NoCaching
  | CacheUnderKey key
```

`NoCaching` will be used for local links, and `CacheUnderKey` for external links. Within `forConcurrentlyCaching` we will check and update the map only if the current entry requires caching.

How does that sound?

---

Regarding the second paragraph - yes, this another approach is a replacement for `forConcurrently`. `checkExternalResource` can be reverted back, I think it won't need to know about `Map`, we can just run it only when necessary (`verifyReferenceRec` or `forConcurrentlyCaching` will care about `Map`).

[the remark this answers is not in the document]
breakerzirconia: Yeah, sounds good! I've already started experimenting with the definition of `forConcurrentlyCaching` some time ago; hopefully, the function's behavior comes out as expected!

breakerzirconia:
Since I've accidentally merged the master branch into mine, the tip of my local branch was behind the remote one, so I had to use `push --force-with-lease`. I will make sure to not let this slip up with the merge happen again.

Martoon-00:
> Since I've accidentally merged the master branch into mine, the tip of my local branch was behind the remote one, so I had to use `push --force-with-lease`. I will make sure not to let such slip up with the merge happen again.

Nah, in such cases I think force-pusing is just fine. When rebasing, you have to force-push in either way :slightly_smiling_face:

src/Xrefcheck/Verify.hs
```diff
@@ -160,9 +190,13 @@ verifyRepo
...
 
-  withAsync (printer progressRef) $ \_ ->
-    fmap fold . forConcurrently toScan $ \(file, ref) ->
-      verifyReference config mode progressRef repoInfo' root file ref
+  accumulated <- withAsync (printer progressRef) $ \_ ->
+    forConcurrentlyCaching toScan (\(_, Reference{..}) -> case locationType rLink of
+      ExternalLoc -> CacheUnderKey rLink
+      _           -> NoCaching) (\(file, ref) -> do
+      VerifyResult results <- verifyReference config mode progressRef repoInfo' root file ref
+      return $ Left results)
```
Martoon-00, line 198: Looks like here you wrap into `Left` (and never into `Right`), and later evaluate `lefts` which brings us back.

At this point `Either`s are probably gone, no need to work with them.

src/Xrefcheck/Verify.hs
```diff
@@ -160,9 +190,13 @@ verifyRepo
 
   progressRef <- newIORef $ initVerifyProgress (map snd toScan)
 
-  withAsync (printer progressRef) $ \_ ->
-    fmap fold . forConcurrently toScan $ \(file, ref) ->
-      verifyReference config mode progressRef repoInfo' root file ref
+  accumulated <- withAsync (printer progressRef) $ \_ ->
+    forConcurrentlyCaching toScan (\(_, Reference{..}) -> case locationType rLink of
+      ExternalLoc -> CacheUnderKey rLink
+      _           -> NoCaching) (\(file, ref) -> do
```
Martoon-00, line 196: Could you move this `case locationType rLink` to a separate (probably local) function? This way it should be simpler to read and update later.

And there will be no need for large parentheses.

src/Xrefcheck/Verify.hs
```diff
@@ -191,17 +191,19 @@ verifyRepo
...
   accumulated <- withAsync (printer progressRef) $ \_ ->
-    forConcurrentlyCaching toScan (\(_, Reference{..}) -> case locationType rLink of
-      ExternalLoc -> CacheUnderKey rLink
-      _           -> NoCaching) (\(file, ref) -> do
+    forConcurrentlyCaching toScan ifExternalThenCache $ \(file, ref) -> do
       VerifyResult results <- verifyReference config mode progressRef repoInfo' root file ref
-      return $ Left results)
-  return . foldMap VerifyResult $ lefts accumulated
+      return results
+  return . foldMap VerifyResult $ accumulated
```
Martoon-00, line 197: And last minor nitpick: wrapping/unwrapping `VerifyResult` now also looks redundant.

Martoon-00 (approved):
Looks good to me now, thanks for addressing my comments :+1: 

If this is tested manually on a repository with several broken links, I'm fine with merging this.

(maybe we should also wait for @dcastro's review).

src/Xrefcheck/Verify.hs
```diff
@@ -138,6 +139,35 @@ instance Buildable VerifyError where
         [h] -> ",\n   did you mean " +| h |+ "?\n"
         hs  -> ", did you mean:\n" +| blockListF' "    -" build hs
 
+data NeedsCaching key
+  = NoCaching
+  | CacheUnderKey key
+
+forConcurrentlyCaching
```
dcastro, line 146: Could you please add a haddock here describing what the function does and what we use it for?

src/Xrefcheck/Verify.hs
```diff
@@ -138,6 +139,35 @@ instance Buildable VerifyError where
...
+      NoCaching -> do
+        withAsync (action x) $ \b ->
+          go (b : acc) cached xs needsCashing action
+      CacheUnderKey cacheKey -> do
+        case M.lookup cacheKey cached of
+          Nothing -> do
+            withAsync (action x) $ \b ->
+              go (b : acc) (M.insert cacheKey b cached) xs needsCashing action
+          Just b -> go (b : acc) cached xs needsCashing action
+    go acc _ [] _ _ = for acc wait <&> reverse
```
dcastro, line 169: Slight improvement suggestion: `needsCashing` and `action` args are not modified in the `go` loop, so they can be "hoisted" to the top-level function, instead of being passed around from one iteration to the next.

```suggestion
forConcurrentlyCaching
  :: forall a b cacheKey
   . Ord cacheKey
  => [a] -> (a -> NeedsCaching cacheKey) -> (a -> IO b) -> IO [b]
forConcurrentlyCaching list needsCashing action = go [] M.empty list
  where
    go
      :: Ord cacheKey
      => [Async b]
      -> M.Map cacheKey (Async b)
      -> [a]
      -> IO [b]
    go acc cached (x : xs) = case needsCashing x of
      NoCaching -> do
        withAsync (action x) $ \b ->
          go (b : acc) cached xs
      CacheUnderKey cacheKey -> do
        case M.lookup cacheKey cached of
          Nothing -> do
            withAsync (action x) $ \b ->
              go (b : acc) (M.insert cacheKey b cached) xs
          Just b -> go (b : acc) cached xs
    go acc _ [] = for acc wait <&> reverse
```

dcastro (commented):
Solid piece of work, well done :D Just a couple of final small suggestions and we're good to go.

src/Xrefcheck/Verify.hs
breakerzirconia: Will do.
I've also been thinking about the fact that `NeedsCaching` is equivalent to `Maybe` and perhaps, even though it is descriptive, is redundant and could be replaced with `Maybe`. @Martoon-00, what do you think?

dcastro: > even though it is descriptive

Personally, I think this is a good enough reason to keep it. It improves readability for maintainers in the future.

dcastro (approved):

breakerzirconia:
Oh, I think I've somehow copied the commits from `origin/master` onto my issue branch. I'll try to figure out how to revert it and put them onto my local `master` branch instead.

Martoon-00:
Okay, the last thing: in the PR description you checked the tick next to `My commits comply with the policy used in Serokell.` point, however, there is a small issue - the rule says that commit descriptions must contain no more than 80 chars at each line. Could you wrap the text there to fit into this restriction?

Issue #49, opened 2020-08-20
[BUG] Urls with `[` and `]` fail to parse


Martoon-00:
# Description

When a link contains braces, xrefcheck reports that as bad url.

# To Reproduce

Have the following link: 

This results in

```
  ➥  In file ./code/indigo/tutorial/docs/index.md
     bad reference (external) at src:66:12-120:
       - text: "new GitLab issue"
       - link: https://gitlab.com/morley-framework/morley/-/issues/new?issue[title]=Indigo%20website:%20
       - anchor: -
     ⛂  Invalid url
```

# Expected behavior

Everything passes.

# Environment

- Ubuntu
- 5ba3605ac7c5fc5da4f0474bcc04855a61265714

Martoon-00:
Apparently the problem appeared after the recent bump of `req` package (which took place along with LTS bump). Now it accepts a URI, not a URL, and specification says

> Of the ASCII character set, the characters : / ? # [ ] @ are reserved for use as delimiters of the generic URI components and must be percent-encoded.

On the other hand, browsers take `[` and `]` just fine, and we want to stay close to their behaviour.

So the easiest way to resolve this seems to be uri-encoding the link before parsing it to URI, `uri-encode` library seems to rescue us here.

kirelagin:
So, this problem is actually a bit tricky. I don’t want to go into all the details, but, basically, this part of RFC 3986 seems to be relevant:

>   URI producing applications should percent-encode data octets that
>   correspond to characters in the reserved set unless these characters
>   are specifically allowed by the URI scheme to represent data in that
>   component.  If a reserved character is found in a URI component and
>   no delimiting role is known for that character, then it must be
>   interpreted as representing the data octet corresponding to that
>   character's encoding in US-ASCII.

This means that an application that produces a URI with unencoded square brackets in the query component is kind of in compliance with the RFC, because it says “should”. However, since there is no delimiting role for square brackets in the query component, any application that _parses_ such a URI _MUST_ accept it.

I guess there is some room for interpretation of this paragraph, since, while the RFC gives a grammar for characters allowed within the query component, it also says

> The query component is indicated by the first question mark ("?") character and terminated by a number sign ("#") character or by the end of the URI.

which one might interpret as saying that anything that is not a `#` is not a delimiter.

[...]

PR #91, opened 2021-10-31
[#49] Provide error messages for illegal square brackets in URIs

breakerzirconia:
## Description

Problem: According to RFC 3986 section 2.2, square brackets are reserved
characters and should be percent-encoded in the URIs; however, when
they are used outside their reserved purpose in the respective context,
they MUST be percent-encoded. Currently, xrefcheck does not provide a
verbose error message when such character is detected in the URI.

Solution: Provide a better means to indicate that such a character has
been detected by supplying an error message describing the location
of the said character and its percent-encoded octet.

## Related issue(s)

#49 

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

- [ ] My code complies with the [style guide](/docs/code-style.md).

### DIFF
--- a/src/Xrefcheck/Util.hs
+++ b/src/Xrefcheck/Util.hs
@@ -11,6 +11,8 @@ module Xrefcheck.Util
   , postfixFields
   , (-:)
   , aesonConfigOption
+  , pairedEnclosingGlyphs
+  , octet
   ) where
 
 import Universum
@@ -18,6 +20,10 @@ import Universum
 import Control.Lens (LensRules, lensField, lensRules, mappingNamer)
 import Data.Aeson qualified as Aeson
 import Data.Aeson.Casing (aesonPrefix, camelCase)
+import Data.Char (toUpper)
+import qualified Data.Map as M
+import qualified Data.Text as T (map)
+import Data.Text.Format (hex)
 import Fmt (Builder, build, fmt, nameF)
 import System.Console.Pretty (Pretty (..), Style (Faint))
 
@@ -43,3 +49,18 @@ infixr 0 -:
 -- | Options that we use to derive JSON instances for config types.
 aesonConfigOption :: Aeson.Options
 aesonConfigOption = aesonPrefix camelCase
+
+pairedEnclosingGlyphs :: M.Map Char Char
+pairedEnclosingGlyphs = M.fromList
+  [ ('(', ')')
+  , (')', '(')
+  , ('[', ']')
+  , (']', '[')
+  , ('{', '}')
+  , ('}', '{')
+  , ('<', '>')
+  , ('>', '<')
+  ]
+
+octet :: Char -> Text
+octet = ("%" <>) . T.map toUpper . fmt @Text . hex . ord
--- a/src/Xrefcheck/Util.hs
+++ b/src/Xrefcheck/Util.hs
@@ -25,7 +25,7 @@ import Data.Text qualified as T
 import Data.Text.Format (hex)
 import Fmt (Builder, build, fmt, nameF)
 import System.Console.Pretty (Pretty (..), Style (..))
-import Text.Megaparsec (ParseError (..), bundleErrors)
+import Text.Megaparsec (ParseError (..), ParseErrorBundle (..), bundleErrors)
 import Text.URI (ParseException (..), URI (..), mkURI)
 
 instance Pretty Builder where
@@ -52,7 +52,7 @@ aesonConfigOption :: Aeson.Options
 aesonConfigOption = aesonPrefix camelCase
 
 -- | Identify the positions of all the non-percent-encoded reserved characters in the URI.
-searchReservedChars :: Text -> ([Int], Text)
+searchReservedChars :: Text -> ([Int], Text, Bool)
 searchReservedChars = withAccumulator 0 []
   where
     -- | At every iteration, call @mkURI@ to attempt the parsing of the link.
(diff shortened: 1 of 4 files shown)


src/Xrefcheck/Verify.hs
```diff
@@ -357,6 +362,24 @@ checkExternalResource VerifyConfig{..} link
...
         Left _ -> False
 
+    processBrackets se = case (fromException se :: Maybe ParseException) of
+      Nothing -> throwError $ ExternalResourceInvalidUri Nothing
+      Just pe ->
+        let peBundle = (\(ParseException bundle) -> bundle) pe
+            err :| _ = bundleErrors peBundle
+        in case err of
+          TrivialError position _ _ ->
+            let characterInQuestion = link `T.index` position
```
dcastro, line 372: Instead of inspecting megaparsec's exception to see if it fails on an "illegal character", wouldn't it be easier to check if the link contains an illegal character upfront (before attempting to use `mkURI`)?

src/Xrefcheck/Util.hs
```diff
@@ -41,3 +46,54 @@ infixr 0 -:
...
+        " if it is present in the URI: " <>
+        style Bold (color Magenta (octet paired))
+
+    applyPrettyAt :: Int -> Text -> Style -> Color -> Text
+    applyPrettyAt pos text sty col =
+      let s = splitAt pos $ toString text
+      in toText $ case s of
+        (prefix, []) -> prefix
+        ([], suffix@(h : t)) ->
+          if pos < 0
```
dcastro, line 81: In what situations can `pos` be negative?

src/Xrefcheck/Util.hs
```diff
@@ -41,3 +46,54 @@ infixr 0 -:
...
+    pinpointInvalidCharacter  =
+      [ "unexpected character " <> show characterInQuestion
+      , style Bold (color Magenta "|")
+      , style Bold (color Magenta "|") <> "  " <>
+        applyPrettyAt position link Bold Magenta
+      , style Bold (color Magenta "|") <> "  " <>
+        T.replicate position " " <> style Bold (color Magenta "^")
+      , show characterInQuestion <>
+        " is not allowed to be explcitily used in the URIs by the WHATWG URL standard"
+      , "Consider using its percent-encoded counterpart: " <>
```
dcastro, line 64: Perhaps instead of just telling the user which characters to replace and with what, we could just show them the URL with all the illegal characters replaced? And then they can just copy-paste it into their .md file.

src/Xrefcheck/Util.hs
```diff
@@ -41,3 +46,54 @@ infixr 0 -:
...
+        style Bold (color Magenta $ octet characterInQuestion)
+      ]
+
+    suggestPercentEncoding = maybeToList $ do
+      paired <- characterInQuestion `M.lookup` pairedEnclosingGlyphs
```
dcastro, line 69: In `illegalCharactersToDisplay` (in the `Verify` module) we're saying that the only characters we consider "illegal" and deserve special handling are `[` and `]`. But here we're saying `[ ] { } < >` are also illegal?

src/Xrefcheck/Verify.hs
```diff
@@ -359,6 +364,62 @@ checkExternalResource VerifyConfig{..} link
...
+          bracketIndices = searchBrackets link
+      in case initialMkURIRequest of
+        Just uri -> pure uri
+        Nothing -> throwError . ExternalResourceInvalidUri $
+          if null bracketIndices
+          then Nothing
+          else Just $ invalidURIVerbose bracketIndices
+
+    invalidURIVerbose :: [Int] -> Text
+    invalidURIVerbose bracketIndices = fmt . indentF 3 . unlinesF $
```
kirelagin, line 379: I’m confused. I was going to comment here that instead of referring to RFC, we should refer to WHATWG URL, and also instead of `bracketIndices` and `unexpected bracket` it should be more general for any invalid characters. But then, @dcastro’s screenshot in Slack shows this fixed. Is there some other version of this change?

kirelagin: Anyway, what I wanted to say is: let’s just get rid of the term URI and always use URL in all user-visible parts.

breakerzirconia: > @dcastro’s screenshot in Slack shows this fixed. Is there some other version of this change?

That was my mistake, I had forgotten to push the latest changes to this PR I had made long before @dcastro's review of the old version. The old version (@dcastro's screenshot) did indeed contain the WHATWG URL reference, which I replaced with the RFC reference in the recent `push`. 

> let’s just get rid of the term URI and always use URL in all user-visible parts

> instead of bracketIndices and unexpected bracket it should be more general for any invalid characters

Got it.

src/Xrefcheck/Verify.hs
dcastro: Ignore this comment - Whether a character is legal or illegal depends on in which part of the URL it is located. A character may be legal in the host string and illegal in the path. Letting `mkUri` do the heavy lifting here is indeed the right choice.

src/Xrefcheck/Verify.hs
kirelagin: Ok, I see. I thought that the WHATWG reference was actually better, since the RFC is pretty ambiguous on that matter and WHATWG URL, in essence, deprecates it anyway. So I suggest returning WHATWG URL.

breakerzirconia:
I've noticed an interesting behavior when parsing a link with `mkURI`.

`mkURI`, when throwing a `ParseException` with the provided position of the error, can specify the character at that position. For example:
```
ParseException (ParseErrorBundle {bundleErrors = TrivialError 13 (Just (Tokens ('[' :| ""))) ... })
```
If, however, a link includes a percent-encoded substring in the _hostname_, (like `https://examp%5Dle.com/path/to/smth`), `mkURI` fails, with the position of the error being **one more** than the position of the last character in the hostname. Moreover, the `ParseException` object omits the character at that position:
```
ParseException (ParseErrorBundle {bundleErrors = TrivialError 22 Nothing ... })
```

Should we print out a separate error message for this specific case? This, as far as I've been able to understand, is the case where the replacement link **cannot** be created for the user, because a reserved character is positioned in the hostname.

PR #109, opened 2022-07-07
[#106] Fix using ./ in paths

### LINKED ISSUE #106: [BUG] Using `./` in paths
# Description

Both the `--ignored` and `--root` options misbehave when the filepath argument begins with `./`.

The following command (when run against the project root) works fine:

```
$ xrefcheck --ignored tests/markdowns --root .

```

But these commands don't:

```
$ xrefcheck --ignored tests/markdowns --root ./
$ xrefcheck --ignored ./tests/markdowns --root .
```

Note: this bug _may or may not_ be related to #90. Once one of the two bugs is fixed, we should try to reproduce the other one and see if it has been fixed as well.

# To Reproduce

Run these commands against the xrefcheck project's root:

```
$ xrefcheck --ignored tests/markdowns --root ./

xrefcheck --ignored tests/markdowns --root ./

  ➥  In file ././.github/pull_request_template.md
     bad reference (absolute) at src:48:33-66:
       - text: "style guide"
       - link: /docs/code-style.md
       - anchor: -
     
     ⛀  File does not exist:
        .//docs/code-style.md
     
     
```

```
$ xrefcheck --ignored ./tests/markdowns --root .

Error when scanning ./tests/markdowns/with-annotations/unexpected_ignore_file.md: "ignore file" must be at the top of markdown or right after comments at the top(11:1-20)
```

# Expected behavior

Expected both commands to behave similarly to `xrefcheck --ignored tests/markdowns --root .` and not return any errors.

# Environment

- master, f69d74e13dc0b4123dfde520f274d13adf0ed423

s-and-witch:
## Description

We used strings equality on file path, but we can have different strings for the same path, for example `./` and `././`, so I added path normalisation at code where we compare them.

Happily, in 9.6 will be new opaque FilePath type, so nobody will use (==) from strings on them.

## Related issue(s)

- Fixed #106

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -9,7 +9,7 @@ steps:
   - command: nix-build ci.nix -A xrefcheck-lib-and-tests
     label: Library and tests
 
-  - command: nix run -f ci.nix pkgs.bats pkgs.diffutils xrefcheck-static -c bash -c "cd tests/golden && bats ./"
+  - command: nix run -f ci.nix pkgs.bats pkgs.diffutils xrefcheck-static -c bash -c "cd tests/golden/ && bats ./**"
     label: Golden tests (bats)
 
   - command: nix-build ci.nix -A xrefcheck-static
@@ -22,7 +22,7 @@ steps:
     artifact_paths:
       - "result/bin/*"
 
-  - command: nix run -f ci.nix xrefcheck-static -c xrefcheck --ignored tests/markdowns
+  - command: nix run -f ci.nix xrefcheck-static -c xrefcheck --ignored tests/markdowns --ignored tests/golden/
     label: Xrefcheck itself
 
   - label: lint
--- a/src/Xrefcheck/CLI.hs
+++ b/src/Xrefcheck/CLI.hs
@@ -27,13 +27,14 @@ import Data.Text qualified as T
 import Data.Version (showVersion)
 import Options.Applicative
   (Parser, ReadM, command, eitherReader, execParser, flag, flag', footerDoc, fullDesc, help, helper,
-  hsubparser, info, infoOption, long, metavar, option, progDesc, short, strOption, switch, value)
+  hsubparser, info, infoOption, long, metavar, option, progDesc, short, strOption, switch, value, OptionFields, Mod)
 import Options.Applicative.Help.Pretty (Doc, displayS, fill, fillSep, indent, renderPretty, text)
 
 import Paths_xrefcheck (version)
 import Xrefcheck.Config (VerifyConfig (..))
 import Xrefcheck.Core
 import Xrefcheck.Scan
+import Xrefcheck.Util (normaliseWithNoTrailing)
 
 modeReadM :: ReadM VerifyMode
 modeReadM = eitherReader $ \s ->
@@ -97,6 +98,9 @@ defaultConfigPaths = ["./xrefcheck.yaml", "./.xrefcheck.yaml"]
 -- and flavors, so we write a type alias here.
 type RepoType = Flavor
 
+filepathOption :: Mod OptionFields FilePath -> Parser FilePath
+filepathOption = fmap normaliseWithNoTrailing <$> strOption
+
 repoTypeReadM :: ReadM RepoType
 repoTypeReadM = eitherReader $ \name ->
   maybeToRight (failureText name) $ L.lookup (map C.toLower name) allRepoTypesNamed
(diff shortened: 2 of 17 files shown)


src/Xrefcheck/Scan.hs
```diff
@@ -67,30 +67,30 @@ specificFormatsSupport formats = \ext -> M.lookup ext formatsMap
...
-  let fileInfos = filter (\(path, _) -> not $ isIgnored path)
+  _ Tree.:/ repoTree <- liftIO $ Tree.readDirectoryWithL processFile root
+  let fileInfos = map (first normalise)
+        $ filter (\(path, _) -> not $ isIgnored path)
         $ dropSndMaybes . F.toList
         $ Tree.zipPaths . (locationOf root Tree.:/)
         $ filterExcludedDirs root repoTree
   return $ RepoInfo (M.fromList fileInfos)
   where
-    rootNE = if null root then "." else root
```
dcastro, line 85: Nice catch! :+1:

src/Xrefcheck/CLI.hs
```diff
@@ -97,6 +98,9 @@ defaultConfigPaths = ["./xrefcheck.yaml", "./.xrefcheck.yaml"]
 -- and flavors, so we write a type alias here.
 type RepoType = Flavor
 
+filepathOption :: Mod OptionFields FilePath -> Parser FilePath
+filepathOption = fmap normalise <$> strOption
```
dcastro, line 102: We should probably do this in _all_ places where we read filepaths from an external source / use hardcoded filepaths.

As far as I can tell, we're only missing 2 places: `readConfig` and `defConfig`. We should probably have a `normaliseFilePaths :: Config -> Config` function, and use it inside `readConfig` / `defConfig`.

dcastro (changes requested):

src/Xrefcheck/Scan.hs
```diff
@@ -67,30 +67,30 @@ specificFormatsSupport formats = \ext -> M.lookup ext formatsMap
 -- > locationOf "/"  == "/"
 locationOf :: FilePath -> FilePath
 locationOf fp
-  | fp == "" || fp == "." = ""
+  | fp == "." = ""
```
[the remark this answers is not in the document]
Player-205: BTW, we use locationOf in one place, on the root variable, that already normalised. I think, better to move this function into where block.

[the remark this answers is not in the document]
dcastro: Sure, sounds good

tests/golden/check-localhost.bats
```diff
@@ -88,3 +90,11 @@ prepare () {
 
   rm /tmp/check-localhost.test
 }
+
+@test "Config: missing, reduchant slashes" {
```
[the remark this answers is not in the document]
dcastro: Probably a bats file+folder for each individual option

.buildkite/pipeline.yml
```diff
@@ -9,8 +9,11 @@ steps:
   - command: nix-build ci.nix -A xrefcheck-lib-and-tests
     label: Library and tests
 
-  - command: nix run -f ci.nix pkgs.bats pkgs.diffutils xrefcheck-static -c bash -c "cd tests/golden && bats ./"
-    label: Golden tests (bats)
+  - command: nix run -f ci.nix pkgs.bats pkgs.diffutils xrefcheck-static -c bash -c "cd tests/golden/check-localhost && bats ./"
+    label: Golden tests for localhost (bats)
+
+  - command: nix run -f ci.nix pkgs.bats pkgs.diffutils xrefcheck-static -c bash -c "cd tests/golden/check-cli && bats ./"
+    label: Golden tests for CLI options (bats)
```
dcastro, line 16: Oh, we should just have one CI step for running all bats tests.

dcastro: You can do `cd tests/golden && bats ./**` to run all bats tests in all subdirs.

On the other hand, this does mean that the "working directory" for all tests will be `tests/golden`, so they'll probably start failing. You can do what we did in coffer:

* Create a helper file with a setup function that changes the working dir ([example](https://github.com/serokell/coffer/blob/main/tests/golden/helpers.bash))
* Add `load '../helpers` at the top of every bats file ([example](https://github.com/serokell/coffer/blob/main/tests/golden/delete-field-command/delete-field-command.bats))

dcastro (changes requested):

dcastro (approved):

PR #110, opened 2022-07-08
[#105] Add hlint and weeder support

### LINKED ISSUE #105: Enable weeder and hlint

We should enable `weeder` and `hlint` in our pipeline, like we've done in [other projects](https://github.com/serokell/segmented-cfmm/blob/53a806061c39fc6a46fe36314f72c07b86e48544/.buildkite/pipeline.yml#L65-L69).

This will let us know of unused dependencies and exports in xrefcheck.

For example, I noticed all our components [depend](https://github.com/serokell/xrefcheck/blob/f69d74e13dc0b4123dfde520f274d13adf0ed423/package.yaml#L80) on `HUnit` (including `xrefcheck:lib`), even though it's only really used in `xrefcheck:test:xrefcheck-tests`. Weeder would have caught this.

# Acceptance criteria

* weeder and hlint have been added to the pipeline.

Player-205:
## Description

I added hlint and weeder configuration files, fixed all useful hints of these tools and add them to the CI pipeline, so we can handle lints and redundant dependencies automatically.

## Related issue(s)

- Fixed  #105

### DIFF
--- a/xrefcheck.nix
+++ b/xrefcheck.nix
@@ -11,7 +11,9 @@ let
     src = pkgs.haskell-nix.haskellLib.cleanGit { src = ./.; };
     modules = [{
       packages.xrefcheck = {
-        ghcOptions = [ "-Werror" ];
+        ghcOptions =
+          [ "-Werror" ];
+
         components.tests = {
           links-tests = {
             build-tools = [ pkgs.vsftpd ];
@@ -40,4 +42,5 @@ let
       };
     }];
   };
-in project.xrefcheck
+in
+project.xrefcheck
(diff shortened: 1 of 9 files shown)


xrefcheck.nix
```diff
@@ -11,7 +11,20 @@ let
...
+            "-Werror"
+            "-Weverything"
+            "-Wno-missing-safe-haskell-mode"
+            "-Wno-unsafe"
+            "-Wno-missing-import-lists"
+            "-Wno-missing-local-signatures"
+            "-Wno-missing-export-lists"
+            "-Wno-all-missed-specialisations"
+            "-Wno-prepositive-qualified-module"
+            "-Wno-monomorphism-restriction"
```
dcastro, line 25: Please move the flags to `package.yaml` - we want them to be turned on at all times, not just in the CI.

The only flag we disable in dev environments is `-Werror`, for convenience.

dcastro: Oh and delete `-Wall`

PR #112, opened 2022-07-12
[#104] Add maxRetries to configurable options

### LINKED ISSUE #104: Make `maxRetries` configurable

In #88, we added support for handling 429 responses and retrying requests. We have a hardcoded limit of `maxRetries = 20`.

We should make this configurable.

# Acceptance criteria

`maxRetries` can be configured through a config file, and overridden via a CLI option.

Player-205:
## Description

I added new field to VerifyConfig, updated all configuration files in tests, added new CLI option to override config, if it's necessary.

## Related issue(s)

- Fixed #104

### DIFF
--- a/src-files/def-config.yaml
+++ b/src-files/def-config.yaml
@@ -48,6 +48,10 @@ verification:
   # wait this amount of time before attempting to verify the link again.
   defaultRetryAfter: 30s
 
+  # How many attempts to retry an external link after getting
+  # a "429 Too Many Requests" response.
+  maxRetries: 3
+
 # Parameters of scanners for various file types.
 scanners:
   markdown:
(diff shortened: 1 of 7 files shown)


src-files/def-config.yaml
```diff
@@ -48,6 +48,9 @@ verification:
   # wait this amount of time before attempting to verify the link again.
   defaultRetryAfter: 30s
 
+  # Count of attempts of getting access to remote data (links to the sites, etc)
```
dcastro, line 51: Let's be a bit more specific here:

```suggestion
  # How many attempts to retry an external link after getting
  # a "429 Too Many Requests" response.
```

dcastro: Please update the CLI option's help text as well.

src-files/def-config.yaml
```diff
@@ -48,6 +48,9 @@ verification:
...
+  maxRetries: 20
```
dcastro, line 52: The hardcoded value of 20 was very, very conservative. Now that this isn't hardcoded anymore, maybe 3 would be a more sensible default?

dcastro (changes requested):

dcastro (approved):

PR #116, opened 2022-07-14
[#49] Allow certain reserved characters in the URLs

### LINKED ISSUE #49: [BUG] Urls with `[` and `]` fail to parse

breakerzirconia:
## Description

Problem: The current version of xrefcheck doesn't allow the square
brackets and some other special characters, like the angle brackets and
the curly brackets, to be present in the URLs, even in the query
strings, as they need to be percent-encoded first.

Solution: Allow some of the reserved characters, like the brackets, to
be present in the query strings of the URLs.
There exist two main standards of URL parsing: RFC 3986 and the Web
Hypertext Application Technology Working Group's URL standard. Ideally,
we want to be able to parse the URLs in accordance with the latter
standard, because it provides a much less ambiguous set of rules for
percent-encoding special characters, and is essentially a living
standard that gets updated constantly.
We allow these characters to be present in the query strings by using
the `parseURI` function from the `uri-bytestring` library with
`laxURIParseOptions`.

## Related issue(s)

Fixes #49

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/src/Xrefcheck/Verify.hs
+++ b/src/Xrefcheck/Verify.hs
@@ -26,6 +26,9 @@ module Xrefcheck.Verify
   , verifyRepo
   , verifyReference
   , checkExternalResource
+
+    -- * URI parsing
+  , parseUri
   ) where
 
 import Universum
@@ -59,8 +62,9 @@ import System.FilePath (takeDirectory, (</>), normalise)
 import System.FilePath.Glob qualified as Glob
 import Text.ParserCombinators.ReadPrec qualified as ReadPrec (lift)
 import Text.Regex.TDFA.Text (Regex, regexec)
-import Text.URI (Authority (..), URI (..), mkURI)
+import Text.URI (Authority (..), URI (..), mkURIBs)
 import Time (RatioNat, Second, Time (..), ms, sec, threadDelay, timeout, (+:+), (-:-))
+import URI.ByteString qualified as URIBS
 
 import Data.Bits (toIntegralSized)
 import Xrefcheck.Config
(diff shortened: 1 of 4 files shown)


src/Xrefcheck/Verify.hs
```diff
@@ -423,7 +424,11 @@ checkExternalResource :: VerifyConfig -> Text -> IO (VerifyResult VerifyError)
 checkExternalResource VerifyConfig{..} link
   | skipCheck = return mempty
   | otherwise = fmap toVerifyRes $ runExceptT $ do
-      uri <- mkURI link & maybe (throwError ExternalResourceInvalidUri) pure
+      uri' <- URIBS.parseURI URIBS.laxURIParserOptions (encodeUtf8 link)
```
dcastro, line 427: Let's also add a test to make sure links with `[` and `]` are properly processed.

dcastro (changes requested):

tests/Test/Xrefcheck/URIParsingSpec.hs
```diff
@@ -0,0 +1,47 @@
...
+
+parse :: Text -> Either URIBS.URIParseError (URIBS.URI)
+parse link = URIBS.parseURI URIBS.laxURIParserOptions (encodeUtf8 link)
+
+spec :: Spec
+spec = do
+  describe "URI parsing should be successful" $ do
+    it "Without the special characters in the query strings" $
+      forM_ withoutSpecChars $ \link -> do
+        assertBool "Failed to parse the link" (isRight $ parse link)
```
dcastro, line 22: Well, looks like these tests are testing the local `parse` function... they're not testing xrefcheck.

If the code in `Verify` changes, this test will continue passing. We won't catch any possible bugs.

dcastro: I suggest extracting those two lines in `checkExternalResource` into a top-level function, e.g. `parseURL :: Text -> ExceptT VerifyError IO URI`, and then export that function so it can be tested here.

dcastro (changes requested):

tests/Test/Xrefcheck/URIParsingSpec.hs
dcastro, line 22: Also, let's not just check that the parse is successful (with `isRight`).

Let's check that it does indeed return the correct URI. I think you should be able to use the `uri` quasiquoter to express what the result should be: https://hackage.haskell.org/package/modern-uri-0.3.4.4/docs/Text-URI-QQ.html

tests/Test/Xrefcheck/URIParsingSpec.hs
```diff
@@ -0,0 +1,72 @@
...
+
+import Xrefcheck.Verify (parseUri, VerifyError (..))
+
+spec :: Spec
+spec = do
+  describe "URI parsing should be successful" $ do
+    it "Without the special characters in the query strings" do
+      parsed <- traverse parseUri' withoutSpecChars
+      assertBool "Failed to parse the link"
+        $ parsed == withoutSpecCharsExpected
```
dcastro, line 25: Instead of using `assertBool msg $ x == y`, let's use ``x `shouldBe` y``, as it gives us much better error messages (i.e. `assertBool` just prints a generic error message, `shouldBe` prints the expected and the actual values as well)

dcastro: In fact, I think these tests could be written more succinctly (and with less indirection, and more readable) like this:

```hs
    it "Without the special characters in the query strings" do
      parseUri' "https://example.com/?q=a&p=b#fragment" `shouldReturn`
        Right [uri|https://example.com/?q=a&p=b#fragment|]

      parseUri' "https://example.com/path/to/smth?q=a&p=b" `shouldReturn`
        Right [uri|https://example.com/path/to/smth?q=a&p=b|]
```

dcastro (changes requested):

dcastro (approved):

PR #127, opened 2022-08-01
[#99] Support `Retry-After` headers with dates

### LINKED ISSUE #99: Support `Retry-After` headers with dates

In #88, we added support for the `Retry-After` header.

The header's value can be either:
1. a date
2. or a non-negative integer representing a number of seconds ([source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After#directives)).

We added support for (2). We should now add support for (1) as well.

# Acceptance criteria

We support `Retry-After` values with a date.

nalkuatov:
This adds some bytes to parse `Retry-After` headers as dates for `429: Too many requests`

## Related issue(s)

Fixes #99 

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/src/Xrefcheck/Verify.hs
+++ b/src/Xrefcheck/Verify.hs
@@ -14,6 +14,7 @@ module Xrefcheck.Verify
   , verifyErrors
   , verifying
 
+  , RetryAfter (..)
   , WithReferenceLoc (..)
 
     -- * Concurrent traversal with caching
@@ -37,10 +38,12 @@ import Data.List qualified as L
 import Data.Map qualified as M
 import Data.Text qualified as T
 import Data.Text.Metrics (damerauLevenshteinNorm)
+import Data.Time (UTCTime, defaultTimeLocale, formatTime, readPTime, rfc822DateFormat)
 import Data.Time.Clock.POSIX (getPOSIXTime)
 import Data.Traversable (for)
-import Fmt (Buildable (..), blockListF', listF, (+|), (|+))
+import Fmt (Buildable (..), blockListF', listF, maybeF, nameF, (+|), (|+))
 import GHC.Exts qualified as Exts
+import GHC.Read (Read (readPrec))
 import Network.FTP.Client
   (FTPException (..), FTPResponse (..), ResponseStatus (..), login, nlst, size, withFTP, withFTPS)
 import Network.HTTP.Client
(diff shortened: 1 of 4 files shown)


src/Xrefcheck/Verify.hs
```diff
@@ -499,13 +525,8 @@ checkExternalResource VerifyConfig{..} link
...
       where
-        retryAfterInfo :: Response a -> Time Second
-        retryAfterInfo response =
-          fromMaybe vcDefaultRetryAfter $ do
-            retryAfterByteString <- L.lookup hRetryAfter $ responseHeaders response
-            retryAfterString <- rightToMaybe $ decodeUtf8Strict @String retryAfterByteString
-            retryAfter <- readMaybe @Natural retryAfterString
-            pure . sec $ fromIntegral @Natural @RatioNat retryAfter
+        retryAfterInfo :: Response a -> Maybe RetryAfter
+        retryAfterInfo = decodeThrow <=< L.lookup hRetryAfter . responseHeaders
```
dcastro, line 529: I don't think we want to use a YAML decoder for this, the header is in plain text.

src/Xrefcheck/Verify.hs
```diff
@@ -300,24 +321,29 @@ verifyReference
     retryVerification numberOfRetries resIO = do
       res@(VerifyResult ves) <- resIO
 
+      now <- getPOSIXTime <&> posixTimeToTimeSecond
+
+      let toSeconds = \case
+            Seconds s -> s
+            -- Calculate the seconds left until @Retry-After@ date
+            Date date -> utcTimeToTimeSecond date -:- now
```
dcastro, line 329: `a -:- b` underflows when `b < a` (see [here](https://hackage.haskell.org/package/o-clock-1.2.1.1/docs/Time-Timestamp.html#v:-45-:-45-))

```hs
λ> sec 3 -:- sec 2
1s
λ> sec 3 -:- sec 4
*** Exception: arithmetic underflow
```

This might happen when the clocks of the client and server are not synced up (and the retry-after timestamp is actually in the past, from the client's point of view). Or when the server means to tell the server to wait 1 second, but there's a stutter in the network and the response takes more than 1 second to arrive.

We should behave sensibly in this scenario (i.e. wait 0 seconds). Let's also write a test for this scenario.

dcastro (changes requested):

src/Xrefcheck/Verify.hs
nalkuatov: Yup, makes a great sense

Sereja313 (approved):

dcastro (approved):
LGTM :+1: :smile: please rebase, squash all fixups, and merge ^^

dcastro:
Oh, and please add an entry to the changelog before merging!

PR #132, opened 2022-09-06
[#125] Display URL parsing errors

### LINKED ISSUE #125: Display URL parsing errors

As of #116, we use a 2-step process to parse a URL: we use `parseURI` and then `mkURIBs`.

Both of these functions can fail. At the moment, we're ignoring their errors and simply throwing a `ExternalResourceInvalidUri`, and then displaying a generic error message to the user.

Instead, we should catch those errors and, to our best effort, tell the user _why_ the URL was invalid.

# Acceptance criteria

The errors from `parseURI` and `mkURIBs` are caught and we use them to tell the user why the URL was invalid.

Sereja313:
## Description

Problem: We use a 2-step process to parse a URL: we use `parseURI` and then `mkURIBs`. Both of these functions can fail. At the moment, we're ignoring their errors and simply throwing a `ExternalResourceInvalidUri`, and then displaying a generic error message to the user.

Solution: Catch errors from `parseUri` and `mkURIBs` and use them to tell user why the URL was invalid.

## Related issue(s)

Fixes #125 

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

- [ ] My code complies with the [style guide](/docs/code-style.md).


src/Xrefcheck/Verify.hs
```diff
@@ -150,8 +152,11 @@ instance Buildable VerifyError where
       "   Use of such anchors is discouraged because referenced object\n\
       \   can change silently whereas the document containing it evolves.\n"
 
-    ExternalResourceInvalidUri ->
-      "⛂  Invalid URI\n"
+    ExternalResourceInvalidUri err ->
+      "⛂  Invalid URI (" +| err |+ ")\n"
+
+    ExternalResourceUriConversionError err ->
+      "⛂  Invalid URI (" +| displayException err |+ ")\n"
```
dcastro, line 159: It seems `displayException @ParserExceptionBs` is [implemented](https://hackage.haskell.org/package/modern-uri-0.3.4.4/docs/src/Text.URI.Types.html#line-255) via `errorBundlePretty`, which is likely to be a multi-line string.

In that case, instead of surrounding the message in `(` `)`, I think it would be better to display it in its own line, with indentation:

```suggestion
      unlinesF
        [ "⛂  Invalid URI"
        , indentF 4 $ build $ displayException err
        ]
```

src/Xrefcheck/Orphans.hs
```diff
@@ -45,3 +46,20 @@ instance Buildable FTPException where
   build (BogusResponseFormatException e) = build e
 
 deriving stock instance Eq FTPException
+
+instance Buildable URIParseError where
+  build = \case
+    MalformedScheme e ->  build e
+    MalformedUserInfo -> "Malformed user info"
+    MalformedQuery -> "Malformed query"
+    MalformedFragment -> "Malformed fargment"
```
dcastro, line 55:
```suggestion
    MalformedFragment -> "Malformed fragment"
```

dcastro (approved):

PR #141, opened 2022-09-13
[#56] Dump all the errors from different files

### LINKED ISSUE #56: Dump all the errors from different files

As we have right in-place annotations now (`<!-- xrefcheck: ignore <mode>-->`), there can be broken markdowns.
For example:
1. `<!-- xrefcheck: ignore file-->` located in the "middle" of the file (not right after the comments at the top) causes `Error when scanning .\markdowns\with-annotations\no_paragraph.md: expected a PARAGRAPH after "ignore paragraph", but found HEADING (9:1-17)`.
2. A typo like `<!-- xrefcheck: ignore ile -->` causes `Error when scanning .\markdowns\with-annotations\unexpected_ignore_file.md: unrecognised option " "ignore ile" perhaps you meant <"ignore link"|"ignore paragraph"|"ignore file"> ( 9:1 - 8:0)`
And so on.

Currently, `xrefcheck` fails immediately after the first observed error because `die` is used right in `markdownScanner`:
```
markdownScanner path = do
    errOrInfo <- parseFileInfo . decodeUtf8 <$> BSL.readFile path
    case errOrInfo of
        Left errTxt -> do
            die $ "Error when scanning " <> path <> ": " <> T.unpack errTxt
        Right fileInfo -> return fileInfo
```
And propagated on a higher level, to `Main`.
What we want is dumping all the errors from different markdowns and then print them as a final `xrefcheck`'s result together with the broken links.
A possible solution is to add a custom exception type for scanner errors, this way exceptions can be reasonably caught outside if necessary (not in inner components).

# Acceptance criteria

1. No `die` in inner components.
2. All the errors are printed as a result together with the broken links.

Sereja313:
## Description

Problem: Currently, xrefcheck fails immediately after the first observed error because `die` is used right in `markdownScanner` What we want is dumping all the errors from different markdowns and then print them as a final xrefcheck's result together with the broken links. Also, despite the fact that in the `makeError` function we have 4 error messages, 2 of them are not reported, and the test case that should check this only checks that at least one of the four files throws an error.
https://github.com/serokell/xrefcheck/blob/1b0b8d5fdbae53438df051bf3d57f91df9e9c125/tests/Test/Xrefcheck/IgnoreAnnotationsSpec.hs#L18-L20
Solution: Make xrefcheck to report all errors. Add `ScanError` type and propagate errors to report all of them, rather than failing immediately after the first error is detected.
## Related issue(s)

Fixes #56

- Tests
        silently reappearing again.

- Documentation
    - [README](/README.md)
    - Haddock

- Public contracts
        and

### DIFF
--- a/src/Xrefcheck/Scanners/Markdown.hs
+++ b/src/Xrefcheck/Scanners/Markdown.hs
@@ -9,16 +9,18 @@
 
 module Xrefcheck.Scanners.Markdown
   ( MarkdownConfig (..)
+  , IgnoreMode (..)
   , defGithubMdConfig
   , markdownScanner
   , markdownSupport
   , parseFileInfo
+  , makeError
   ) where
 
 import Universum
 
 import CMarkGFM (Node (..), NodeType (..), PosInfo (..), commonmarkToNode)
-import Control.Monad.Except (MonadError, throwError)
+import Control.Monad.Trans.Writer.CPS (Writer, runWriter, tell)
 import Data.Aeson.TH (deriveFromJSON)
 import Data.ByteString.Lazy qualified as BSL
 import Data.DList qualified as DList
@@ -53,7 +55,7 @@ toPosition = Position . \case
     | startLine == endLine -> Just $
         startLine |+ ":" +| startColumn |+ "-" +| endColumn |+ ""
     | otherwise -> Just $
-        " " +|
+        "" +|
         startLine |+ ":" +| startColumn |+ " - " +|
         endLine |+ ":" +| endColumn |+ ""
 
@@ -76,38 +78,107 @@ data IgnoreMode
   | None
   deriving stock (Eq)
 
+-- | Bind `IgnoreMode` to its `PosInfo` so that we can tell where the
+-- corresponding annotation was declared.
+data Ignore = Ignore IgnoreMode (Maybe PosInfo)
+
+type ScannerM a = StateT Ignore (Writer [ScanError]) a
+
+-- | Empty `Ignore` state
+ignoreNone :: Ignore
+ignoreNone = Ignore None Nothing
+
 -- | A fold over a `Node`.
 cataNode :: (Maybe PosInfo -> NodeType -> [c] -> c) -> Node -> c
 cataNode f (Node pos ty subs) = f pos ty (cataNode f <$> subs)
 
 -- | Remove nodes with accordance with global `MarkdownConfig` and local
 --   overrides.
-removeIgnored :: Node -> Either Text Node
-removeIgnored = runIdentity . runExceptT . flip evalStateT None . cataNode remove
+removeIgnored :: FilePath -> Node -> Writer [ScanError] Node
+removeIgnored fp = withIgnoreMode . cataNode remove
   where
     remove
-      :: (MonadError Text m, MonadState IgnoreMode m)
-      => Maybe PosInfo
+      :: Maybe PosInfo
       -> NodeType
-      -> [m Node]
-      -> m Node
+      -> [ScannerM Node]
+      -> ScannerM Node
     remove pos ty subs = do
-      mode <- get
+      Ignore mode modePos <- get
+      let node = Node pos ty []
       case (mode, ty) of
-        (Paragraph, PARAGRAPH) -> put None $> defNode
-        (Paragraph, x)         -> throwError (makeError mode (prettyType x) pos)
-        (File, _)              -> throwError (makeError mode "" pos)
-        (Link, LINK {})        -> put None $> defNode
-        (Link, _)              -> Node pos ty <$> sequence subs
-        (None, _) -> do
-          case getIgnoreMode (Node pos ty []) of
-            Just mode' -> put mode' $> defNode
+        -- We expect to find a paragraph immediately after the
+        -- `ignore paragraph` annotanion. If the paragraph is not
+        -- found we should report an error.
+        (Paragraph, PARAGRAPH) -> put ignoreNone $> defNode
+        (Paragraph, x)         -> do
+          lift . tell . makeError modePos fp mode $ prettyType x
+          put ignoreNone
+          Node pos ty <$> sequence subs
+
+        -- We don't expect to find an `ignore file` annotation here,
+        -- since that annotation should be at the top of the file and
+        -- the file should already be ignored when `checkIgnoreFile` is called.
+        -- We should report an error if we find it anyway.
+        (File, _)              -> do
+          lift . tell $ makeError modePos fp mode ""
+          put ignoreNone
+          Node pos ty <$> sequence subs
+
+        -- When we find an `ignore link` annotation, we skip nodes until
+        -- we find a link and ignore it, or we find another ignore annotation,
+        -- then we should report an error and set new `Ignore` state.
+        (Link, LINK {})        -> put ignoreNone $> defNode
+        (Link, _)            ->
+          case getIgnoreMode node of
+            Just mode'  -> do
+              lift . tell $ makeError modePos fp mode ""
+              handleMode node mode'
+            Nothing     -> Node pos ty <$> sequence subs
+
+        -- When no `Ignore` state is set check next node for annotation,
+        -- if found then set it as new `IgnoreMode` otherwise skip node.
+        (None, _)              ->
+          case getIgnoreMode node of
+            Just mode' -> handleMode node mode'
             Nothing    -> Node pos ty <$> sequence subs
+
+    handleMode
+      :: Node
+      -> IgnoreMode
+      -> ScannerM Node
+    handleMode node = \case
+      -- Report unknown `IgnoreMode`.
+      None   -> do
+        let unrecognised = fromMaybe ""
+              $ safeHead . drop 1 . words =<< getXrefcheckContent node
+        lift . tell $ makeError (getPosition node) fp None unrecognised
+        put ignoreNone $> defNode
+      -- Set new `Ignore` state.
+      mode'  -> put (Ignore mode' $ getPosition node) $> defNode
 
     prettyType :: NodeType -> Text
     prettyType ty =
       let mType = safeHead $ words $ show ty
       in fromMaybe "" mType
+
+    withIgnoreMode
+      :: ScannerM Node
+      -> Writer [ScanError] Node
+    withIgnoreMode action = action `runStateT` ignoreNone >>= \case
+      -- We expect `IgnoreMode` to be `None` when we reach EOF,
+      -- otherwise that means there was an annotation that didn't match
+      -- any node, so we have to report that.
+      (node, Ignore None _) -> pure node
+      (node, (Ignore mode pos))
+        | mode == Paragraph -> do
+            tell $ makeError pos fp mode "EOF"
+            pure node
+        -- Link and File scan errors do not require extra text info
+        -- to make error description.
+        | otherwise      -> do
+            tell $ makeError pos fp mode ""
+            pure node
+
 
 -- | Custom `foldMap` for source tree.
 foldNode :: (Monoid a, Monad m) => (Node -> m a) -> Node -> m a
@@ -116,24 +187,20 @@ foldNode action node@(Node _ _ subs) = do
   b <- concatForM subs (foldNode action)
   return (a <> b)
 
+type ExtractorM a = ReaderT MarkdownConfig (Writer [ScanError]) a
+
 -- | Extract information from source tree.
 nodeExtractInfo
-  :: forall m
-  .  ( MonadError Text m
-     , MonadReader MarkdownConfig m
-     )
-  => Node
-  -> m FileInfo
-nodeExtractInfo input@(Node _ _ nSubs) = do
+  :: FilePath
+  -> Node
+  -> ExtractorM FileInfo
+nodeExtractInfo fp input@(Node _ _ nSubs) = do
   if checkIgnoreFile nSubs
   then return def
-  else case removeIgnored input of
-    Left err -> throwError err
-    Right relevant ->
-      diffToFileInfo <$> foldNode extractor relevant
+  else diffToFileInfo <$> (foldNode extractor =<< lift (removeIgnored fp input))
 
   where
-    extractor :: Node -> m FileInfoDiff
+    extractor :: Node -> ExtractorM FileInfoDiff
     extractor node@(Node pos ty _) =
       case ty of
         HTML_BLOCK _ -> do
@@ -182,6 +249,8 @@ nodeExtractInfo input@(Node _ _ nSubs) = do
 
         _ -> return mempty
 
+-- | Check if there is `ignore file` at the beginning of the file,
+-- ignoring preceding comments if there are any.
 checkIgnoreFile :: [Node] -> Bool
 checkIgnoreFile nodes =
   let isSimpleComment :: Node -> Bool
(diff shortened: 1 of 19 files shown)


src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -82,33 +84,73 @@ cataNode f (Node pos ty subs) = f pos ty (cataNode f <$> subs)
...
 
+    handleMode
+      :: (MonadWriter [ScanError] m, MonadState IgnoreMode m)
+      => Node
+      -> IgnoreMode
+      -> m Node
+    handleMode node@(Node pos _ _)= \case
+      None   -> do
+        let unrecognised = fromMaybe ""
+              $ safeHead . drop 1 . words =<< getXrefcheckContent node
```
dcastro, line 129: It seems this is duplicating some of the work in `getIgnoreMode`...

Thinking about it a bit more, I think the issue is that we're using `None` to represent two distinct things:

1. In ``action `runStateT` None`` (line 147), `None` means no "ignore mode" is currently set
2. In `getIgnoreMode / textToMode`, `None` means we found an unrecognised "ignore mode"

I think we should add another `IgnoreMode` constructor like `UnknownMode Text` for the 2nd use case.
Then, here in `handleMode`, you could simply pattern match on `UnknownMode` to obtain the name of the unrecognised mode, no need to parse it again.

dcastro (changes requested):

src/Xrefcheck/Scanners/Markdown.hs
dcastro: Actually, adding another constructor might not be the right thing to do either. It wouldn't make sense to do `put (UnknownMode "x")`.

Instead, I think it would make more sense to change `getIgnoreMode`. Instead of returning `Maybe IgnoreMode`, it should return something like:

```hs
data GetIgnoreMode
  = NoMode
  | ValidMode IgnoreMode
  | InvalidMode Text
```

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -70,7 +72,7 @@ nodeExtractText = T.strip . mconcat . map extractText . nodeFlatten
     nodeFlatten (Node _pos ty subs) = ty : concatMap nodeFlatten subs
 
 data IgnoreMode
-  = Link
+  = Link (Maybe PosInfo)
```
dcastro, line 75: Let's add some haddocks explaining that `PosInfo` here represents the position where the `ignore link` annotation was declared.

tests/golden/check-scan-errors/expected.gold
```diff
@@ -0,0 +1,74 @@
...
+     scan error at src:17:1 - 16:0:
+
+     ⛀  Expected a LINK after "ignore link" annotation
+
+
+
+  ➥  In file ./check-scan-errors.md
+     scan error at src:21:1 - 20:0:
+
+     ⛀  Unrecognised option "unrecognised-annotation" perhaps you meant <"ignore link"|"ignore paragraph"|"ignore file">
```
dcastro, line 34: Hm, it seems the position info of these 4 errors is not entirely consistent:

* The `Expected a LINK` error points to line 17, where the annotation is declared.
* The `Unrecognised option` error points to line 21, where the annotation is declared.

This behaviour seems correct to me.
However:

* The `Annotation "ignore file" must` error points to line 11, a couple of lines below where the annotation is declared.
* The `Expected a PARAGRAPH` error points to line 15, a couple of lines below where the annotation is declared.


Can you please make these last two errors behave like the first two?

dcastro (changes requested):

dcastro:
> Also, despite the fact that in the `makeError` function we have 4 error messages, 2 of them are not reported, and the test case that should check this only checks that at least one of the four files throws an error.

Great catch! :clap: Thank you for fixing this and adding these tests, it really helped uncover lots of issues with the existing code, well done!

tests/golden/check-scan-errors/expected.gold
dcastro: Oh just noticed another odd thing:

```
  ➥  In file ./check-scan-errors.md
     scan error at src:17:1 - 16:0:

     ⛀  Expected a LINK after "ignore link" annotation
```

It's saying the error starts in line 17/column 1 and ends in line 16 / column 0? :thinking: I think it should say `src:17:1-31`

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -82,33 +84,73 @@ cataNode f (Node pos ty subs) = f pos ty (cataNode f <$> subs)
...
+          tell . scanError pos $ makeError mode ""
+          put None
+          Node pos ty <$> sequence subs
+        (Link _, LINK {})      -> put None $> defNode
+        (Link pos', _)         ->
+          case getIgnoreMode node of
+            Just mode'  -> do
+              tell . scanError pos' $ makeError mode ""
+              handleMode node mode'
+            Nothing     -> Node pos ty <$> sequence subs
```
dcastro, line 115: Can you please explain this `(Link pos', _)` case?

If the mode is currently `Link` and we find a non-LINK node, shouldn't we unconditionally report a scan error? Why do we check `getIgnoreMode` and why does the `Nothing` branch not report an error?

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -82,33 +84,73 @@ cataNode f (Node pos ty subs) = f pos ty (cataNode f <$> subs)
...
     prettyType ty =
       let mType = safeHead $ words $ show ty
       in fromMaybe "" mType
 
+    withIgnoreMode
+      :: (MonadWriter [ScanError] m)
+      => StateT IgnoreMode m Node
+      -> m Node
+    withIgnoreMode action = action `runStateT` None >>= \case
+      (node, link@(Link pos)) -> do
```
dcastro, line 148: Same here: why does the `Link` case get special treatment?

If the scanning `action` ends with a `Paragraph`/`File` mode, we don't report an error (presumably because it's already been reported _during_ the scanning). But if it ends with a `Link` mode, we report an error - why? Shouldn't it already have been reported during the scanning `action`?

src/Xrefcheck/Scanners/Markdown.hs
Sereja313: One of the examples in the `README` shows that the link does not have to follow the annotation
https://github.com/serokell/xrefcheck/blob/c4f0bfb9fc771d706902b9741f3b3aaca5ac47c7/README.md?plain=1#L102-L106
Same for tests 
https://github.com/serokell/xrefcheck/blob/f4eb187f31918367f71b741a3066f0632070d394/tests/markdowns/with-annotations/ignore_link.md?plain=1#L34-L38
I thought it was on purpose... :thinking:  I tried to keep the old behavior. Do I need to change this?

src/Xrefcheck/Scanners/Markdown.hs
Sereja313: For the reasons mentioned above, I thought the link might be deeply nested in other nodes, so we had to report that we reached EOF without finding any links. In fact, we probably need to keep this even if we change the behavior of the `ignore link` annotation so that we can report that there is no link after the annotation if the annotation is used at the end of file. Also, right now, if we use `ignore paragraph` on the last line of the file, nothing is reported.

src/Xrefcheck/Scanners/Markdown.hs
dcastro: > Here we have text node before the link

> I tried to keep the old behavior

Oh you're absolutely right, I forgot about that!

Ok, can you sprinkle some comments in this function to explain all these decisions? E.g.
* In the `Paragraph` case, we should explain that "When we find an `ignore paragraph` annotation, we expect to find a paragraph immediately after"
* In the `Link` case, explain that when we find an `ignore link` annotation, we skip nodes until (1) we find a link, (2) find another `ignore` annotation, in which case we throw an error, or (3) reach the end of the file, in which case we also throw an error.
* We don't expect to find an `ignore file` annotation at all while traversing the nodes. We've **already** used `checkIgnoreFile` to extract those annotations located at the top of the file, and handle them. If we do find one anyway, that means it's **not** at the top of the file, in which case we throw an error.

You don't necessarily have to use my words, these are just examples of things I think are not completely obvious from a first glance and would benefit from better docs. Feel free to add more notes as well if you think they could be useful.

Also, please add some haddocks to `checkIgnoreFile` and `getIgnoreMode` while we're here :D

src/Xrefcheck/Scanners/Markdown.hs
dcastro: Okay that makes sense!

> Also, right now, if we use ignore paragraph on the last line of the file, nothing is reported, we probably need to report that as well.

Yup, that's true. More generally, the ignore mode must be set to `None` when we reach the end of the file.

dcastro: A non-`None` mode implies that there was an annotation in the file that didn't match any node.

src/Xrefcheck/Scanners/Markdown.hs
dcastro: One more thing: with the change I proposed above, I think the `None` constructor no longer makes sense.

I.e., it wouldn't make sense for `getIgnoreMode` to return `ValidMode None`. That would be an invalid state, and according to the "make illegal states unrepresentable" mantra, the types shouldn't allow this to even be expressed.

I think we should do this:
* `IgnoreMode` should be just `Link`, `Paragraph` and `File`
* `ignoreMode` returns `NotAnAnnotation`, `ValidMode IgnoreMode` or `InvalidMode Text`
* In the `remove` helper function, we change from `MonadState IgnoreMode` to `MonadState (Maybe IgnoreMode)`, where `Nothing` represents no mode is currently set.

This is turning into a lot of work, so let's not do this in this PR, leave this part as it is. Can you please write a new follow-up issue for refactoring `IgnoreMode`?

tests/golden/check-scan-errors/expected.gold
Sereja313: Oh that's very weird. 
Also, even though both `ingore file` annotations are exactly the same, we get different lengths for these annotations
https://github.com/serokell/xrefcheck/blob/1295eecc205c7bdcbae4a07629887bfcc5484a1a/tests/golden/check-scan-errors/expected.gold#L3-L13
There seems to be a bug in `commonmarkToNode` from the `cmark-gfm` package.

Sereja313: It looks like this bug only affects the positions of one line `HTML_BLOCK`s.

Sereja313: > Also, even though both ignore file annotations are exactly the same, we get different lengths for these annotations

My bad. This actually is not related to ` commonmarkToNode` bug. This is because we report position of the next node instead of the position of the annotation.

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -76,64 +78,129 @@ data IgnoreMode
...
       let mType = safeHead $ words $ show ty
       in fromMaybe "" mType
 
+    withIgnoreMode
+      :: ScannerM Node
+      -> Writer [ScanError] Node
+    withIgnoreMode action = action `runStateT` ignoreNone >>= \case
+      -- We expect `IgnoreMode` to be `None` when we reach EOF,
+      -- otherwise that means there was an annotation that didn't match
+      -- any node, so we have to report that.
```
dcastro, line 170: We should probably test this scenario (`ignore link` and `ignore paragraph` at the end of the file)

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -247,23 +307,33 @@ getCommentContent :: Node -> Maybe Text
...
   let notStripped = T.stripPrefix "xrefcheck:" . T.strip =<<
         getCommentContent node
   in T.strip <$> notStripped
 
+-- | Get correct possition of an annotation node. There is a bug in
+-- `commonmarkToNode` from the `cmark-gfm` package that affects one line
+-- `HTML_BLOCK` nodes, those node have wrong end line and end column positions.
+-- As our annotations are always oneliners, we can fix this by simply setting
+-- end line equals to start line and calculating end column from start column
+-- and annotation length.
```
dcastro, line 327: Can you please open an issue in their repo? Preferably with a small reproducible example, like something they could paste into their GHCI.

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -247,23 +307,33 @@ getCommentContent :: Node -> Maybe Text
...
+getPosition :: Node -> Maybe PosInfo
+getPosition node@(Node pos _ _) = do
+      annLength <- length . T.strip <$> getHTMLText node
+      PosInfo sl sc _ _ <- pos
+      pure $ PosInfo sl sc sl (sc + annLength - 1)
```
dcastro, line 332:
```suggestion
  annLength <- length . T.strip <$> getHTMLText node
  PosInfo sl sc _ _ <- pos
  pure $ PosInfo sl sc sl (sc + annLength - 1)
```

src/Xrefcheck/Scanners/Markdown.hs
dcastro: Great job!

Sorokin-Anton (approved):

src/Xrefcheck/Scanners/Markdown.hs
Sereja313: [Done](https://github.com/kivikakk/cmark-gfm-hs/issues/23)

tests/Test/Xrefcheck/IgnoreAnnotationsSpec.hs
```diff
@@ -7,17 +7,41 @@ module Test.Xrefcheck.IgnoreAnnotationsSpec where
...
+      getErrs file `shouldReturn`
+        makeError (Just $ PosInfo 7 1 7 46) file None "unrecognised-option"
+    it "Check if ignore link annotation produce error when placed at the end of file" do
+      let file = "tests/markdowns/with-annotations/no_link_eof.md"
+      getErrs file `shouldReturn`
+        makeError (Just $ PosInfo 9 1 9 31) file Link ""
+    it "Check if ignore paragraph annotation produce error when placed at the end of file" do
+      let file = "tests/markdowns/with-annotations/no_paragraph_eof.md"
+      getErrs file `shouldReturn`
+        makeError (Just $ PosInfo 9 1 9 36) file Paragraph "EOF"
```
dcastro, line 44: I think it would be even better if these 2 new tests were written in bats instead, because then we can see whether the error message printed to stdout is sensible.

dcastro: Tried this manually, it seems there's a double space in `"but found^^EOF"`.

dcastro (approved):
Apart from that single comment above, LGTM, so I'm pre-approving :smile: 

Please merge after handling that comment, and I'll cut the 0.2.1 release :tada:

Sereja313:
@dcastro looks like `readDirectoryWithL` scans files in a different order on different machines(test in [bf5501e](https://github.com/serokell/xrefcheck/pull/141/commits/bf5501edf2b1e8db0c8ea846673d6e27efd1dff6) works for me but doesn't work in CI and vice versa for [069434e](https://github.com/serokell/xrefcheck/pull/141/commits/069434e6e064202bd52fa6d6ad2e1c6f5dbd64de)).
Do I need to sort scan errors (or perhaps file paths in a directory tree before scanning) to make tests reproducible?

dcastro:
@Sereja313 Sounds like a good idea!

> sort scan errors (or perhaps file paths in a directory tree before scanning) 

Whichever way you prefer

Sereja313:
Done :tada:

dcastro:
Great work @Sereja313 :smile: :tada:

PR #163, opened 2022-09-24
[#162] Do not cancel the progress bar thread

### LINKED ISSUE #162: [BUG] Jumbled up stdout/stderr
# Description

Every now and then the output of `xrefcheck` will appear a confusing mess.
Sometimes it may even clear up the entire content of the terminal.

### Example 1

* A part of the progress bar was not erased, see the "external ..." text in yellow at the bottom.
* The "All repository links are valid" message is not in white as it should be.

image

### Example 2

* The actual repository data was printed and then quickly erased from the terminal
* The "A" in "All repository links are valid" is missing

image

# To Reproduce

The easiest way to reproduce this is to run xrefcheck on a loop until we start seeing the behaviour described above.

```
$ mkdir demo && cd demo
$ echo "[a](a.md)" > a.md
$ while xrefcheck -v; do ; done
```

# Expected behavior

Expected xrefcheck to _always_ print this:

image

# Environment

* master 4cf1f562f74f8668edbf62da3a4f9c3dc2c28728
* Konsole + oh-my-zsh

dcastro:
## Description

Problem: The output of xrefcheck sometimes appears as a confusing mess, see here for details and examples: https://github.com/serokell/xrefcheck/issues/162

The culprit seems to be the `withAsync` used in `verifyRepo`. This spawns a thread that prints and refreshes the progress bar, while the main thread coordinates the verification of references.

https://github.com/serokell/xrefcheck/blob/4cf1f562f74f8668edbf62da3a4f9c3dc2c28728/src/Xrefcheck/Verify.hs#L268-L270

The problem here is, [as the docs of `withAsync` explain](https://hackage.haskell.org/package/async-2.2.4/docs/Control-Concurrent-Async.html#v:withAsync), when the main thread finishes/throws, the spawned thread will be cancelled with `uninterruptibleCancel`.

```hs
withAsync action inner = mask $ \restore -> do
  a <- async (restore action)
  restore (inner a) `finally` uninterruptibleCancel a
```

If the thread is in the middle of updating the progress bar (printing return carriages, writing over an existing line, printing control sequences to change the text color, etc), it'll be abruptly interrupted and will not be given the chance to finish cleanly.

Solution: replace `withAsync` with `loopAsyncUntil`. This lets the printer thread finish what it's currently doing.

----

Additionally, there was a `threadDelay (ms 100)` in `Xrefcheck.Progress.allowRewrite` whose purpose is not clear. 
I really don't what issue it's supposed to fix and how it fixes it.

This seems like it may have been an earlier attempt to fix the issue in #162, so I'm deleting this delay.

## Related issue(s)

Fixes #162

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

#### ✓ Release Checklist

- [ ] (After merging) I created a new entry in the [releases](https://github.com/serokell/xrefcheck/releases) page,
      with a summary of all user-facing changes.
    *  I made sure a tag was created using the format `vX.Y.Z`


dcastro:
@Sereja313 @Sorokin-Anton Please try to reproduce this issue on your machines using the instructions in #162, let me know if (a) you can reproduce it and (b) whether this PR fixes it for you as well.

Sereja313:
I was able to reproduce the issue. PR fixes the problem for me.

Sereja313 (approved):

dcastro:
Thank you! Pushed a fixup that simplifies `loopAsyncUntil` a bit using `whenM`

Sorokin-Anton:
In my vscode terminal there was text coloring issues (for example when making your test about #162), and they are fixed by this PR

Sorokin-Anton (approved):

Martoon-00:
Just some thoughts re-iterating this:

Maybe wrapping the printing action into `uninterruptibleMask` would be a simpler fix for this issue?
This would work as all `throwTo`-like functions (and so `uninterruptibleCancel`) wait for the exception to be delivered.

Despite all concerns with "uninterruptible", this seems to be exactly what we want: in no circumstance abort the printing in the middle, better abandon waiting for the progress bar thread than stop it prematurely. And this is as important as the cleanup phase in `withAsync` that is run with "uninterruptible".

---

And one additional pros of this: an interrupt sent by the user won't abort the printing action in the middle, which we probably want along with the original issue. 

This problem holds even after #209 because there is a short period of time when the action within `loopAsyncUntil` is not wrapped into `handleAsync \UserInterrupt -> ...`. So in case of exception, the main thread may finish in the middle of the progress bar rendering, leaving the terminal in an inconsistent state again.

Issue #165, opened 2022-09-25
Test all variants of local references


dcastro:
# Clarification and motivation

Let's extend the bats tests a bit to test some basic local references:
* "self links" to an anchor within the same file: `#anchor`
* "relative links" with and without anchors: `dir/file.md`, `dir/file.md#anchor`, `./dir/file.md`, `dir/./file.md`, `../dir/file.md`
* "absolute links" with and without anchors: `/dir/file.md`, `dir/file.md#anchor`, `/dir/./file.md`
* Test all of this when the root directory matches the current directory (i.e. the default `--root .`) **and** when it doesn't (i.e. `--root somedir`)

The tests in `tests/golden/check-anchors` already cover the specific case of relative links with anchors when `--root .`, but I think those tests should be replaced with a more complete suite for _all_ kinds of local references

# Acceptance criteria

* We have bats tests for all variants of local references

dcastro:
I forgot to mention: we should also test (relative / absolute) links to:
* directories
* non-markdown files

dcastro:
One more: links with trailing path separator, e.g. a file `a.md/` and a directory `dir/`

PR #174, opened 2022-09-26
[#139] [#168] [#195] Ignore build-related files

### LINKED ISSUE #139: Ignore build-related files

At the moment, we're using the `ignored` option for mainly 2 purposes:

1. to ignore all files in the `.git` folder (`.git/**/*`)
2. to ignore all build-related temporary files (the default config ignores `.stack-work/**/*`)

Regarding point (2): 
* It would be useful to ignore not only `.stack-work` but also `dist-newstyle`.
* `.stack-work` is specific to Haskell projects. Other languages/build tools will have their own directories they want to ignore (e.g. `node_modules`).
* We can't possibly list all build-related directories related to all languages/tools. Which means that, with this approach, there'll always be a bias towards the languages/tools **we** use ourselves.

~A more robust alternative might be to introduce a new option, `useGitIgnore`, which ignores all files ignored by `.gitignore`.~
A more robust alternative might be to ignore all files implicitly ignored by git.

Internally, we could use the ~[`git check-ignore <filename>`](https://git-scm.com/docs/git-check-ignore)~ [`git ls-files`](https://git-scm.com/docs/git-ls-files) command to implement this.

~This option's default value should be `true`.~

As part of this issue, let's also try and refactor `gatherRepoInfo` so that it uses `git ls-files` instead of `Tree.readDirectoryWithL`.

# Acceptance criteria

* ~By default,~ we automatically ignore all files ignored by git
* ~The user can optionally disable this behaviour.~
* ~The option's description~ The readme should mention this requires having git >= 2.18.0 in the PATH.

### LINKED ISSUE #168: [BUG] `ignored` option does not pretend the file doesn't exist
# Description

( Original conversation, see the comments in https://github.com/serokell/xrefcheck/issues/118 )

The help text for the `--ignored` option in `xrefcheck --help` says the following:

> Files which we pretend do not exist.

However, this doesn't seem to be true. We do **not** pretend they don't exist. If a link references that file, we do in fact check that it exists.

# To Reproduce

```
mkdir demo && cd demo
echo "[](b.md)" > a.md
touch b.md
```

```
$ xrefcheck --ignored b.md
```

# Expected behavior

Expected `xrefcheck` to pretend `b.md` doesn't exist, and report an error in the reference from the `a.md` file.

# Environment

* 955e0ea9cf389e782cf30faf8f613756ea22cd0c

### LINKED ISSUE #195: [BUG] Trailing slashes in filepaths
# Description

At the moment, we're rejecting relative/absolute filepaths that end with a trailing slash.

```
[a](README.md/)
```

Technically, this is indeed an invalid filepath:

```
$ touch a.md 
$ stat a.md/
stat: cannot stat 'a.md/': Not a directory
```

However, both github and gitlab correctly link `README.md/` to `https://github.com/serokell/xrefcheck/blob/master/README.md`, so we should consider it valid.

We already have tests in place for this (see #194), but at the moment we expect those links to be invalid. As part of this issue, let's change the expectation of those tests.

# To Reproduce

```
$ mkdir demo && cd demo
$ echo "[a](a.md/)" > a.md
$ xrefcheck

                                                                                                                                                    

  ➥  In file a.md
     bad reference (relative) at src:1:1-10:
       - text: "a"
       - link: a.md/
       - anchor: -
     
     ⛀  File does not exist:
        a.md/
     
     
```

# Expected behavior

Expected the links to be considered valid.

Sereja313:
## Description

Problem: At the moment, we're using the ignored option for mainly 2 purposes: 1) to ignore all files in the `.git` folder (`.git/**/*`) to ignore all build-related temporary files (the default config ignores `.stack-work/**/*`). A more robust alternative might be to ignore all files implicitly ignored by git.

Solution: Use `git ls-files` to ignore all files implicitly ignored by git.
## Related issue(s)

Fixes #139
Fixes #168
Fixes #195 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

#### ✓ Release Checklist


Sorokin-Anton:
Am I right that now we check only staged files? And what about links to unstaged ones - will we report them? Another comment -  I am not sure that xrefcheck should work only with Git repos (and before this commit we can run this in any folder), so I think there should be CLI opt and config flag to disable "git ls-files" usage

Sorokin-Anton:
> We really support only git repositories

@dcastro what do you think about making an option to disable "git ls-files" call? For example, someone can use xrefcheck on directory which is served by FTP server. I think it's better to give not perfect reports (when talking about rendering GitHub/GitLab anchors) then don't give any

src/Xrefcheck/Scan.hs
```diff
@@ -89,6 +89,20 @@ specificFormatsSupport formats = \ext -> M.lookup ext formatsMap
...
+-- | Process files that are tracked by git and not ignored by the config.
+readDirectoryWith
+  :: TraversalConfig
+  -> (FilePath -> IO a)
+  -> FilePath
+  -> IO [(FilePath, a)]
+readDirectoryWith config scanner root =
+  traverse scanFile
+  . filter (not . isIgnored)
+  . L.lines =<< readProcess "git" ["ls-files", root] ""
```
Sorokin-Anton, line 101: I am not sure this works as you expect

`~/serokell` is a folder where I clone different repositories

```
anton@RedmiBook:~/serokell$ git ls-files xrefcheck
fatal: not a git repository (or any of the parent directories): .git
anton@RedmiBook:~/serokell$ git ls-files ~/serokell/xrefcheck
fatal: not a git repository (or any of the parent directories): .git
```

Sorokin-Anton: And if I do `cd xrefcheck && git ls-files`, everything is fine

Sereja313:
Looks like #168 is no longer an issue. Added test.

Sorokin-Anton (approved):
IDK how we can add tests on CI which checks new behavior, by the way

Sereja313:
Added tests

src/Xrefcheck/Verify.hs
```diff
@@ -408,18 +410,17 @@ verifyReference
...
     checkReferredFileExists file = do
-      let fileExists = readingSystem $ doesFileExist file
-      let dirExists = readingSystem $ doesDirectoryExist file
+      let fileExists =
+            any (equalFilePath $ normaliseWithNoTrailing file) $ M.keys repoInfo
+          isVirtual = matchesGlobPatterns root vcVirtualFiles file
 
-      let isVirtual = matchesGlobPatterns root vcVirtualFiles file
-
-      unless (fileExists || dirExists || isVirtual) $
```
dcastro, line 422: Hold on, we removed the `dirExists` check...

Won't this break links to directories? That reminds me, do we even have tests for links to directories?

/cc @Sorokin-Anton I think I forgot about these in #165 :sweat: I've reopened the issue, can you please have a look at [this comment](https://github.com/serokell/xrefcheck/issues/165#issuecomment-1276549285)?

src/Xrefcheck/Verify.hs
```diff
@@ -408,18 +410,17 @@ verifyReference
...
+      case join $ M.lookup referredFile repoInfo of
         Nothing -> pass  -- no support for such file, can do nothing
         Just referredFileInfo -> whenJust mAnchor $
           checkAnchor referredFile (_fiAnchors referredFileInfo)
 
     checkReferredFileExists file = do
-      let fileExists = readingSystem $ doesFileExist file
-      let dirExists = readingSystem $ doesDirectoryExist file
+      let fileExists =
+            any (equalFilePath $ normaliseWithNoTrailing file) $ M.keys repoInfo
```
dcastro, line 420: This feels off...

In this line, we're using `Map.keys + any + equalFilePath + normaliseWithNoTrailing` to check if a filepath exists in the map. This implies that the `file` variable might not look exactly like the filepath in the `Map` (e.g. one might contain a leading `./`, the other might not), so we need to use `equalFilePath` to correct that behavior.

However, in line 413, we're using `M.lookup`, which implies that `file` variable _does_ look exactly like the filepaths in the `Map`.

Can you please rebase this on master? The tests in #189 have been merged to master, I have a suspicion some of them might fail.

.buildkite/pipeline.yml
```diff
@@ -28,7 +28,7 @@ steps:
     artifact_paths:
       - "result/bin/*"
 
-  - command: nix run -f ci.nix xrefcheck-static -c xrefcheck --ignored 'tests/**/*'
+  - command: nix run -f ci.nix xrefcheck-static -c xrefcheck --ignored 'tests/markdowns/**/*' --ignored 'tests/golden/**/*'
```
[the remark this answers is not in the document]
Sereja313: This PR also fixes the bug from #168, so we now report an error whenever we find a link to an ignored file. And we have a link to `tests/configs/github-config.yaml` in `CHANGES.md`
https://github.com/serokell/xrefcheck/blob/33fcfdb739bf43bd97734a99fea94b7de500cb40/CHANGES.md?plain=1#L81-L84

src/Xrefcheck/Verify.hs
Sereja313: Oh, indeed tests for local references fail. I will fix it, thanks!

src/Xrefcheck/Verify.hs
Sereja313: Oops, i forgot about directories. Thanks!

.buildkite/pipeline.yml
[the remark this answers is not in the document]
dcastro: Ah makes sense! Thank you for the explanation

tests/golden/check-ignored/check-ignored.bats
```diff
@@ -37,5 +37,47 @@ load '../helpers'
...
+  ⛀  Annotation "ignore file" must be at the top of markdown or right after comments at the top
+
+
+
+Scan errors dumped, 1 in total.
+EOF
+}
+
+@test "Ignore referenced file, check error" {
+  to_temp xrefcheck --ignored referenced-file.md
```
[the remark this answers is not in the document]
dcastro: Just realized this is a regression test for #168 :+1: Nice!

Sereja313:
Rebased on top of #194. Also #195 is no longer an issue.

src/Xrefcheck/Verify.hs
```diff
@@ -439,11 +461,17 @@ verifyReference
...
         throwError (LocalFileDoesNotExist file)
+      where
+        checkExists :: Bool
+        checkExists = any (liftA2 (||) filePathExists dirExists) $ M.keys repoInfo
+
+        filePathExists :: FilePath -> Bool
+        filePathExists = equalFilePath $ expandIndirections file
+
+        dirExists :: FilePath -> Bool
+        dirExists = any filePathExists . scanl (</>) "" . splitDirectories
```
dcastro, line 474: The name `dirExists` is a bit misleading...

It seems to suggest `dirExists x` will check if the directory `x` exists.

But in fact, it'll check if any of the parent directories of `x` match the free variable `file`.

dcastro: On a similar note: Instead of retrieving the parent dirs of every single file in `RepoInfo` every single time, what if we collected all directories beforehand?

E.g., change `scanRepo` to return this:

```
data RepoInfo = RepoInfo 
  { riFiles :: Map FilePath (Maybe FileInfo)
  , riDirectories :: Set FilePath
  }
```

I believe this would (1) save a lot of work and (2) make the code's intent much clearer.

src/Xrefcheck/Verify.hs
```diff
@@ -417,11 +420,30 @@ verifyReference
...
+    lookupFilePath :: FilePath -> [(FilePath, Maybe FileInfo)] -> Maybe FileInfo
+    lookupFilePath fp = snd <=< find (equalFilePath (expandIndirections fp) . fst)
+
+    -- expands ".." and "."
+    -- expandIndirections "a/b/../c"      = "a/c"
+    -- expandIndirections "a/b/c/../../d" = "a/d"
+    -- expandIndirections "../../a"       = "../../a"
+    -- expandIndirections "a/./b"         = "a/b"
+    -- expandIndirections "a/b/./../c"    = "a/c"
+    expandIndirections :: FilePath -> FilePath
```
dcastro, line 437: As far as I can tell, this is needed because otherwise `dir1/dir2/..` won't match with `dir1`, even tho it should, right?

It works, but it does feel a bit hackish :thinking: 

I think we should solve this problem by using canonical paths across the entire codebase though. Then, simply using `==` everywhere should work.

This will probably be a bigger effort, so I'll write a separate issue for this.

dcastro: Okay, created a follow-up issue here: https://github.com/serokell/xrefcheck/issues/197

Sereja313:
Also fixed a bug where the `--verbose` option displayed all tracked files, not just scanned ones.

src/Xrefcheck/Verify.hs
Sereja313: Yes, you got it right. Thanks for creating the issue.

src/Xrefcheck/Verify.hs
Sereja313: Fixed. Thanks!

dcastro (approved):

PR #209, opened 2022-10-27
[#89] Handle user interrupts

### LINKED ISSUE #89: Dump the processed errors upon triggering `UserInterrupt` (Ctrl-C)

Hitting Ctrl-C raises an `AsyncException` and terminates the further execution of the program, but since the current version of xrefcheck doesn't handle it, all the errors that have already been processed don't get displayed.

# Acceptance criteria

When the user presses Ctrl-C and force terminates the execution of xrefcheck, all the already processed errors, if present, should get printed out, or the usual `All repository links are valid.` otherwise.

Sorokin-Anton:
## Description

Currently, if xrefcheck is interrupted, it's not displaying any `VerifyError`s, even if it has already found some. When we found error, we are notifying user via progress bar, but he must wait checking of all references to see it
(and it can be long enough).

Solution: when xrefcheck is interrupted, display all errors we have already found and a number of
checked references.
To do this, we need to modify `forConcurrentlyCaching` behaviour on exceptions.

## Related issue(s)

Fixes #89

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/src/Xrefcheck/Verify.hs
+++ b/src/Xrefcheck/Verify.hs
@@ -29,12 +29,13 @@ module Xrefcheck.Verify
 
     -- * URI parsing
   , parseUri
+  , reportVerifyErrs
   ) where
 
 import Universum
 
-import Control.Concurrent.Async (async, wait, withAsync)
-import Control.Exception (throwIO)
+import Control.Concurrent.Async (AsyncCancelled (AsyncCancelled), async, cancel, wait, withAsync)
+import Control.Exception (AsyncException (..), handle, throwIO)
 import Control.Monad.Catch (handleJust)
 import Control.Monad.Except (MonadError (..))
 import Data.ByteString qualified as BS
@@ -45,7 +46,7 @@ import Data.Text.Metrics (damerauLevenshteinNorm)
 import Data.Time (UTCTime, defaultTimeLocale, formatTime, readPTime, rfc822DateFormat)
 import Data.Time.Clock.POSIX (getPOSIXTime)
 import Data.Traversable (for)
-import Fmt (Buildable (..), maybeF, nameF)
+import Fmt (Buildable (..), blockListF', fmt, fmtLn, maybeF, nameF, (+|), (|+))
 import GHC.Exts qualified as Exts
 import GHC.Read (Read (readPrec))
 import Network.FTP.Client
@@ -227,6 +228,17 @@ instance Given ColorMode => Buildable VerifyError where
       ⛂  #{err}
       |]
 
+reportVerifyErrs
+  :: Given ColorMode => NonEmpty (WithReferenceLoc VerifyError) -> IO ()
+reportVerifyErrs errs = fmt
+  [int||
+  === Invalid references found ===
+
+  #{interpolateIndentF 2 (interpolateBlockListF' "➥ " build errs)}
+  Invalid references dumped, #{length errs} in total.
+  |]
+
+
 data RetryAfter = Date UTCTime | Seconds (Time Second)
   deriving stock (Show, Eq)
 
@@ -260,22 +272,39 @@ data NeedsCaching key
 -- then the @action@ will not be executed, and the value is added to the accumulator list.
 -- After the whole list has been traversed, the accumulator is traversed once again to ensure
 -- every asynchronous action is completed.
+-- If interrupted by AsyncException, returns this exception and list of already calcualted results
+-- (their subset can be arbitrary). Computations that were not ended till this moment are cancelled.
 forConcurrentlyCaching
   :: Ord cacheKey
-  => [a] -> (a -> NeedsCaching cacheKey) -> (a -> IO b) -> IO [b]
-forConcurrentlyCaching list needsCaching action = go [] M.empty list
+  => [a] -> (a -> NeedsCaching cacheKey) -> (a -> IO b) -> IO (Either (AsyncException, [b]) [b])
+forConcurrentlyCaching list needsCaching action = do
+  res <- go [] M.empty list
+  case res of
+    Right r -> return $ Right $ map (fromMaybe (error "impossible")) r
+      -- impossible, since no actions were cancelled
+
+    Left (exception, partialResult) -> return $  Left (exception, catMaybes partialResult)
   where
-    go acc cached (x : xs) = case needsCaching x of
-      NoCaching -> do
-        withAsync (action x) $ \b ->
-          go (b : acc) cached xs
-      CacheUnderKey cacheKey -> do
-        case M.lookup cacheKey cached of
-          Nothing -> do
-            withAsync (action x) $ \b ->
-              go (b : acc) (M.insert cacheKey b cached) xs
-          Just b -> go (b : acc) cached xs
-    go acc _ [] = for acc wait <&> reverse
+    interruptibleAction a = handle (\AsyncCancelled -> pure Nothing) (Just <$> action a)
+
+    go acc cached items = handle
+      (\exception -> Left . (exception,) <$> for acc (\f -> cancel f >> wait f))
+      -- If action was already completed, then @cancel@ will have no effect, and we
+      -- will get a result from @cancel f >> wait f@. Otherwise interrupted action
+      -- will return us @Nothing@.
+      case items of
+        [] -> Right . reverse <$> for acc wait
+
+        (x : xs) -> case needsCaching x of
+          NoCaching -> do
+            withAsync (interruptibleAction x) $ \b ->
+              go (b : acc) cached xs
+          CacheUnderKey cacheKey -> do
+            case M.lookup cacheKey cached of
+              Nothing -> do
+                withAsync (interruptibleAction x) $ \b ->
+                  go (b : acc) (M.insert cacheKey b cached) xs
+              Just b -> go (b : acc) cached xs
 
 verifyRepo
   :: Given ColorMode
@@ -306,7 +335,17 @@ verifyRepo
   accumulated <- loopAsyncUntil (printer progressRef) do
     forConcurrentlyCaching toScan ifExternalThenCache $ \(file, ref) ->
       verifyReference config mode progressRef repoInfo' root file ref
-  return $ fold accumulated
+  case accumulated of
+    Right res -> return $ fold res
+    Left (exception, partialRes) -> do
+      let errs = verifyErrors (fold partialRes)
+          total = length toScan
+          checked = length partialRes
+      whenJust errs $ reportVerifyErrs
+      fmtLn $ "\nInterrupted (" <> show exception <> "), checked "
+        +| checked |+ " out of " +| total |+ " references."
+      exitFailure
+
   where
     printer :: IORef VerifyProgress -> IO ()
     printer progressRef = do
@@ -681,12 +720,12 @@ checkExternalResource Config{..} link
       -> Bool
       -> ExceptT VerifyError IO ()
     makeFtpRequest host port path secure = handler host port $
-      \handle response -> do
+      \ftpHandle response -> do
         -- check connection status
         when (frStatus response /= Success) $
           throwError $ ExternalFtpResourceUnavailable response
         -- anonymous login
-        loginResp <- login handle "anonymous" ""
+        loginResp <- login ftpHandle "anonymous" ""
         -- check login status
         when (frStatus loginResp /= Success) $
           if ncIgnoreAuthFailures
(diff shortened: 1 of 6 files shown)


src/Xrefcheck/Verify.hs
```diff
@@ -231,22 +274,39 @@ data NeedsCaching key
...
-        case M.lookup cacheKey cached of
-          Nothing -> do
-            withAsync (action x) $ \b ->
-              go (b : acc) (M.insert cacheKey b cached) xs
-          Just b -> go (b : acc) cached xs
-    go acc _ [] = for acc wait <&> reverse
+    interruptibleAction a = handle (\AsyncCancelled -> pure Nothing) (Just <$> action a)
+
+    go acc cached items = handle
+      (\exception -> Left . (exception,) <$> for acc (\f -> cancel f >> wait f))
```
dcastro, line 293: I think this would be easier to read if we (1) avoided point-free notation, (2) used `do` blocks, and (3) renamed `f` to `asyncAction`.

```suggestion
      (\exception -> do
          partialResults <- for acc \asyncAction -> do
            cancel asyncAction
            wait asyncAction
          pure $ Left (exception, partialResults)
      )
```

dcastro, line 293: Should we be more conservative and handle `UserInterrupt` specifically? Instead of all `AsyncException`s.

dcastro, line 293: Let's also add a comment explaining what's going on:

> When a user hits Ctrl+C, a `UserInterrupt :: AsyncException` is thrown onto the main thread.
> We catch it here, cancel all child threads, and collect their results.

src/Xrefcheck/Verify.hs
```diff
@@ -231,22 +274,39 @@ data NeedsCaching key
...
+-- (their subset can be arbitrary). Computations that were not ended till this moment are cancelled.
 forConcurrentlyCaching
   :: Ord cacheKey
-  => [a] -> (a -> NeedsCaching cacheKey) -> (a -> IO b) -> IO [b]
-forConcurrentlyCaching list needsCaching action = go [] M.empty list
+  => [a] -> (a -> NeedsCaching cacheKey) -> (a -> IO b) -> IO (Either (AsyncException, [b]) [b])
+forConcurrentlyCaching list needsCaching action = do
+  res <- go [] M.empty list
+  case res of
+    Right r -> return $ Right $ map (fromMaybe (error "impossible")) r
```
dcastro, line 285: I think we can avoid this `error "impossible"` case if we use [`poll`](https://hackage.haskell.org/package/async-2.2.4/docs/Control-Concurrent-Async.html#v:poll) (which returns a `(Maybe (Either SomeException a))`), instead of `wait` (which returns `a`).

Consider:

```diff
 forConcurrentlyCaching
   :: forall a b cacheKey. Ord cacheKey
   => [a] -> (a -> NeedsCaching cacheKey) -> (a -> IO b) -> IO (Either (AsyncException, [b]) [b])
 forConcurrentlyCaching list needsCaching action = do
-  res <- go [] M.empty list
-  case res of
-    Right r -> return $ Right $ map (fromMaybe (error "impossible")) r
-      -- impossible, since no actions were cancelled
-
-    Left (exception, partialResult) -> return $  Left (exception, catMaybes partialResult)
+  go [] M.empty list
   where
-    interruptibleAction :: a -> IO (Maybe b)
-    interruptibleAction a = handle (\AsyncCancelled -> pure Nothing) (Just <$> action a)
-
     go
-      :: [Async (Maybe b)]
-      -> Map cacheKey (Async (Maybe b))
+      :: [Async b]
+      -> Map cacheKey (Async b)
       -> [a]
-      -> IO (Either (AsyncException, [Maybe b]) [Maybe b])
+      -> IO (Either (AsyncException, [b]) [b])
     go acc cached items = handle
       (\exception -> do
           partialResults <- for acc \asyncAction -> do
             cancel asyncAction
-            wait asyncAction
-          pure $ Left (exception, partialResults)
+            poll asyncAction <&> \case
+              Nothing -> Nothing
+              Just (Right a) -> Just a
+              Just (Left _ex) -> Nothing
+          pure $ Left (exception, catMaybes partialResults)
       )
       -- If action was already completed, then @cancel@ will have no effect, and we
       -- will get a result from @cancel f >> wait f@. Otherwise interrupted action
       -- will return us @Nothing@.
       case items of
         [] -> Right . reverse <$> for acc wait
 
         (x : xs) -> case needsCaching x of
           NoCaching -> do
-            withAsync (interruptibleAction x) $ \b ->
+            withAsync (action x) $ \b ->
               go (b : acc) cached xs
           CacheUnderKey cacheKey -> do
             case M.lookup cacheKey cached of
               Nothing -> do
-                withAsync (interruptibleAction x) $ \b ->
+                withAsync (action x) $ \b ->
                   go (b : acc) (M.insert cacheKey b cached) xs
               Just b -> go (b : acc) cached xs

```

src/Xrefcheck/Verify.hs
```diff
@@ -277,7 +337,17 @@ verifyRepo
   accumulated <- loopAsyncUntil (printer progressRef) do
     forConcurrentlyCaching toScan ifExternalThenCache $ \(file, ref) ->
       verifyReference config mode progressRef repoInfo' root file ref
-  return $ fold accumulated
+  case accumulated of
+    Right res -> return $ fold res
+    Left (exception, partialRes) -> do
+      let errs = verifyErrors (fold partialRes)
```
dcastro, line 343: Let's add some docs here for context.

```suggestion
      -- The user has hit Ctrl+C; display any verification errors we managed to find and exit.
      let errs = verifyErrors (fold partialRes)
```

src/Xrefcheck/Verify.hs
```diff
@@ -231,22 +274,39 @@ data NeedsCaching key
...
-      CacheUnderKey cacheKey -> do
-        case M.lookup cacheKey cached of
-          Nothing -> do
-            withAsync (action x) $ \b ->
-              go (b : acc) (M.insert cacheKey b cached) xs
-          Just b -> go (b : acc) cached xs
-    go acc _ [] = for acc wait <&> reverse
+    interruptibleAction a = handle (\AsyncCancelled -> pure Nothing) (Just <$> action a)
+
+    go acc cached items = handle
```
dcastro, line 292: Hm it's a bit weird to have an exception handler in a recursive function like `go`... If the list `[a]` has 10 items, we'll end up with 10 exception handles stacked on top of each other.

I think a better place to do this would be in the `case items of [] ->` branch, where we wait for all children threads to complete:

```hs
      case items of
        [] ->
          -- Wait for all children threads to complete.
          --
          -- If, while the threads are running, the user hits Ctrl+C,
          -- a `UserInterrupt :: AsyncException` will be thrown onto the main thread.
          -- We catch it here, cancel all child threads, and return the results of only the threads that finished successfully.
          (Right . reverse <$> for acc wait)
          `catch`
          \exception -> do
            partialResults <- for acc \asyncAction -> do
              cancel asyncAction
              poll asyncAction <&> \case
                Nothing -> Nothing
                Just (Right a) -> Just a
                Just (Left _ex) -> Nothing
            pure $ Left (exception, catMaybes partialResults)
`

Comment (Sorokin-Anton) [reply src/Xrefcheck/Verify.hs:292]: in this case I was able to quickly hit Ctrl-C before items list was empty. Even when progress bar indicates some errors. This is very strange because spawning of async threads is fast. I will test this again, and if in normal scenario it works well, I agree that this looks better

Comment (Sorokin-Anton) [reply src/Xrefcheck/Verify.hs:292]: Yes, this is totaly not working for me, e.g. when I interrupt xrefheck during it checks external links on xrefcheck repo (and waiting 429's timeout with blue progress bar). Maybe we need to mask exceptions in other cases, I'll try to figure this out

Comment (Sorokin-Anton) [reply src/Xrefcheck/Verify.hs:292]: Everything except this is done

Comment (Sorokin-Anton) [reply src/Xrefcheck/Verify.hs]: Last  fixup uses approach from comment

File: src/Xrefcheck/Verify.hs [hunk 5]
```diff
@@ -5,6 +5,7 @@
 
 {-# LANGUAGE DeriveFunctor         #-}
 {-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE ScopedTypeVariables #-}
```
dcastro, line 8: This is already in `default-extensions`, so it shouldn't be needed here ^^

src/Xrefcheck/Verify.hs
```diff
@@ -231,22 +275,50 @@ data NeedsCaching key
...
+                          poll asyncAction <&> \case
+                            Just (Right a) -> Just a
+                            Just (Left _ex) -> Nothing
+                            Nothing -> Nothing
+                        pure $ Left (exception, catMaybes partialResults)
+                )
+                $ Right . reverse <$> for acc wait
+      -- If action was already completed, then @cancel@ will have no effect, and we
+      -- will get result from @cancel f >> pool f@. Otherwise action will be interrupted,
+      -- so pool will return @Left (SomeException AsyncCancelled)@
```
dcastro, line 321:
```suggestion
      -- will get result from @cancel f >> poll f@. Otherwise action will be interrupted,
      -- so poll will return @Left (SomeException AsyncCancelled)@
```

src/Xrefcheck/Verify.hs
dcastro: Good job, well done!

Instead of replacing `handle` from `universum` (which is actually re-exported from `safe-exceptions`) with `handle` from `exceptions`, I think it would be more idiomatic to replace it with `handleAsync` from `safe-exceptions`.

dcastro: Good job, well done!

Instead of replacing `handle` from `universum` (which is actually re-exported from `safe-exceptions`) with `handle` from `exceptions`, I think it would be more idiomatic to replace it with `handleAsync` from `safe-exceptions`.

dcastro: Also, please don't forget to add this comment near `handle`:

```hs
          -- Wait for all children threads to complete.
          --
          -- If, while the threads are running, the user hits Ctrl+C,
          -- a `UserInterrupt :: AsyncException` will be thrown onto the main thread.
          -- We catch it here, cancel all child threads,
          -- and return the results of only the threads that finished successfully.
```

src/Xrefcheck/Verify.hs
```diff
@@ -231,22 +275,50 @@ data NeedsCaching key
...
+                  go (b : acc) (M.insert cacheKey b cached) xs
+              Just b -> go (b : acc) cached xs
+
+        [] -> handleJust
+                (\case
+                  UserInterrupt -> Just UserInterrupt
+                  _ -> Nothing
+                )
+                (\exception -> do
+                        partialResults <- for acc \asyncAction -> do
```
dcastro, line 310: Excessive indentation in this `do` block

src/Xrefcheck/Verify.hs
```diff
@@ -231,22 +275,50 @@ data NeedsCaching key
...
+              go (b : acc) cached xs
+          CacheUnderKey cacheKey -> do
+            case M.lookup cacheKey cached of
+              Nothing -> do
+                withAsync (action x) $ \b ->
+                  go (b : acc) (M.insert cacheKey b cached) xs
+              Just b -> go (b : acc) cached xs
+
+        [] -> handleJust
+                (\case
```
dcastro, line 305: Excessive indentation here too, it should be:

```hs
[] ->
  handleJust
    (\case ...
```

Or

```hs
[] -> handleJust
  (\case ...
```

src/Xrefcheck/Verify.hs
Sorokin-Anton: `handleAsync` is not allowing us to handle only part of Async exceptions (because there is no `handleAsyncJust`), but I don't see a problem with handling all things from `AsyncException` type. Also to avoid depending both `exceptions` and `safe-exceptions`, I changed `handleJust` here
```
      mkURIBs (URIBS.serializeURIRef' uri')
           -- Ideally, this exception should never be thrown, as the URI
           -- already *percent-encoded* with 'parseURI' from 'uri-bytestring'
           -- and 'mkURIBs' is only used to convert to 'URI' type from
           -- 'modern-uri' package.
           & handleJust (fromException @ParseExceptionBs)
           (throwError . ExternalResourceUriConversionError)
```
to one from `safe-exceptions` (it was the only use of exceptions package before)

src/Xrefcheck/Verify.hs
Sorokin-Anton: both looks a bit strange for me, but ok

src/Xrefcheck/Verify.hs
Sorokin-Anton: done

src/Xrefcheck/Verify.hs
dcastro: > `handleAsync` is not allowing us to handle only part of `Async` exceptions

That's fine, we can just re-throw the other constructors (that's what `handleJust` and `catchJust` do [under the hood anyway](https://hackage.haskell.org/package/base-4.17.0.0/docs/src/Control.Exception.Base.html#catchJust)).

So:

```hs
handleAsync
  (\case
    UserInterrupt x -> ...
    otherAsyncEx -> throwM otherAsyncEx
  )
```

> Also to avoid depending both exceptions and safe-exceptions, I changed handleJust here

Good idea!

dcastro (approved):

PR #212, opened 2022-11-01
[#208] Remove exessive newlines from output

Sorokin-Anton:
## Description

Problem: xrefcheck's output contains many redundant newlines, so it takes more display space than it could.
For list of places where such newlines appear, see #208

Solution: don't print redundant newlines, so output is more compact. Since tests are newline-sensetive, I've checked that now there are no extra  (e.g. 2 ajacent) blank lines in `expected` part of tests.

## Related issue(s)

Fixes #208

## :white_check_mark: Checklist for your Pull Request

Ideally a PR has all of the checkmarks set.

If something in this list is irrelevant to your PR, you should still set this
checkmark indicating that you are sure it is dealt with (be that by irrelevance).

#### Related changes (conditional)

- Tests
  - [x] If I added new functionality, I added tests covering it.
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [x] I checked whether I should update the docs and did so if necessary:
    - [README](https://github.com/serokell/xrefcheck/tree/master/README.md)
    - Haddock

- Public contracts
  - [x] Any modifications of public contracts comply with the [Evolution
  of Public Contracts](https://www.notion.so/serokell/Evolution-of-Public-Contracts-2a3bf7971abe4806a24f63c84e7076c5) policy.
  - [ ] I added an entry to the [changelog](https://github.com/serokell/xrefcheck/tree/master/CHANGES.md) if my changes are visible to the users
        and
  - [x] provided a migration guide for breaking changes if possible

#### Stylistic guide (mandatory)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [x] My code complies with the [style guide](https://github.com/serokell/style/blob/master/haskell.md).

PR #214, opened 2022-11-01
[#213] Do not print trailing whitespaces

### LINKED ISSUE #213: Do not print trailing whitespaces
Currently xrefcheck's output contain many trailing whitespaces, since `Fmt.indentF 2 "a\n\nb\n" == "  a\n  \n  b\n"`, so e.g. when we redirect xrefcheck output to `expected.gold` for bats tests, we need to trim whitespaces manually (since we don't want them in commited files).
I suggest use analogs of `Fmt.identF`, `Fmt.blockListF` and `Fmt.blockListF'` that are not producing lines containing only whitespaces, and fix other reasons that can cause appearing of trailing whitespaces.

# Acceptance criteria
There is no trailing whitespaces in xrefcheck output. Bats tests are space-sensetive

Sorokin-Anton:
## Description

Problem: bats tests are not space sensetive
Solution: remove trailing spaces from xrefcheck output (see next problems), remove `--ignore-trailing-space` from `assert_diff`

Problem: there are lines containing only spaces in xrefcheck's output, because `Fmt.indentF` "indents" empty lines too.
Solution: add `Xrefcheck.Util.indentF_` function that is not indenting empty lines. Same for `Fmt.blockListF`

Problem: when there is a current file link `[a](#b)`, it is printed like
```
- text: "a"
- link: (trailing space here)
- anchor: b
```
Solution: as in case when there is no anchor, print `link: -` instead

## Related issue(s)

Fixes #213 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -110,3 +110,15 @@
     name: "Avoid style function that ignore ColorMode"
     lhs: 'System.Console.Pretty.style'
     rhs: 'Xrefcheck.Util.Colorize.styleIfNeeded'
+- warn:
+    name: "Avoid functions that generate extra trailing newlines/whitespaces"
+    lhs: 'Fmt.indentF'
+    rhs: 'Xrefcheck.Util.Interpolate.interpolateIndentF'
+- warn:
+    name: "Avoid functions that generate extra trailing newlines/whitespaces"
+    lhs: 'Fmt.blockListF'
+    rhs: 'Xrefcheck.Util.Interpolate.interpolateBlockListF'
+- warn:
+    name: "Avoid functions that generate extra trailing newlines/whitespaces"
+    lhs: "Fmt.blockListF'"
+    rhs: "Xrefcheck.Util.Interpolate.interpolateBlockListF'"
--- a/package.yaml
+++ b/package.yaml
@@ -146,7 +146,6 @@ tests:
       - xrefcheck
       - bytestring
       - directory
-      - fmt
       - http-types
       - o-clock
       - regex-tdfa
@@ -159,6 +158,7 @@ tests:
       - uri-bytestring
       - yaml
       - reflection
+      - nyan-interpolation
 
   ftp-tests:
     main: Main.hs
--- a/src/Xrefcheck/Util/Interpolate.hs
+++ b/src/Xrefcheck/Util/Interpolate.hs
@@ -1,0 +1,81 @@
+{- SPDX-FileCopyrightText: 2018-2019 Serokell <https://serokell.io>
+ -
+ - SPDX-License-Identifier: MPL-2.0
+ -}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
+module Xrefcheck.Util.Interpolate
+  ( -- $notes
+    interpolateIndentF
+  , interpolateBlockListF
+  , interpolateBlockListF'
+  )
+  where
+
+import Universum
+
+import Data.Text.Lazy qualified as TL
+import Data.Text.Lazy.Builder (fromLazyText, toLazyText)
+import Fmt (Buildable, Builder, blockListF, blockListF', indentF)
+
+{- $notes
+The `blockListF` and `indentF` frunctions from @fmt@ add a trailing newline, which makes them unsuitable for string interpolation.
+Consider this case:
+> [int||
+> aaa
+> #{indentF 2 "bbb"}
+> ccc
+> |]
+One would reasonably expect this to produce:
+> aaa
+>   bbb
+> ccc
+But, in reality, it produces:
+> aaa
+>   bbb
+>
+> ccc
+This module introduces versions of these functions that do not produce a trailing newline
+and can therefore be safely used in string interpolation.
+-}
+
+{-# HLINT ignore "Avoid functions that generate extra trailing newlines/whitespaces" #-}
+
+-- | Like @Fmt.indentF@, but strips trailing spaces and does not add a trailing newline.
+--
+-- >>> import Fmt
+-- >>> indentF 2 "a\n\nb"
+-- "  a\n  \n  b\n"
+--
+-- >>> interpolateIndentF 2 "a\n\nb"
+-- "  a\n\n  b"
+interpolateIndentF :: HasCallStack => Int -> Builder -> Builder
+interpolateIndentF n b = (case TL.last (toLazyText b) of
+  '\n' -> id
+  _ ->  stripLastNewline) $ stripTrailingSpaces $ indentF n b
+  -- strips newline added by indentF
+
+-- | Like @Fmt.blockListF'@, but strips trailing spaces and does not add a trailing newline.
+interpolateBlockListF' :: HasCallStack => Text -> (a -> Builder) -> NonEmpty a -> Builder
+interpolateBlockListF' =  stripLastNewline . stripTrailingSpaces ... blockListF'
+
+-- | Like @Fmt.blockListF@, but strips trailing spaces and does not add a trailing newline.
+interpolateBlockListF :: HasCallStack => Buildable a => NonEmpty a -> Builder
+interpolateBlockListF = stripLastNewline . stripTrailingSpaces . blockListF
+
+-- remove trailing whitespace from all lines.
+-- Note: output always ends with newline (adds trailing newline if there wasn't one).
+stripTrailingSpaces :: Builder -> Builder
+stripTrailingSpaces
+  = fromLazyText
+  . TL.unlines
+  . map (TL.stripEnd)
+  . TL.lines
+  . toLazyText
+
+stripLastNewline :: HasCallStack => Builder -> Builder
+stripLastNewline
+  = fromLazyText
+  . fromMaybe (error "stripLastNewline: expected newline to strip")
+  . TL.stripSuffix "\n"
+  . toLazyText
--- a/src/Xrefcheck/Verify.hs
+++ b/src/Xrefcheck/Verify.hs
@@ -45,7 +45,7 @@ import Data.Text.Metrics (damerauLevenshteinNorm)
 import Data.Time (UTCTime, defaultTimeLocale, formatTime, readPTime, rfc822DateFormat)
 import Data.Time.Clock.POSIX (getPOSIXTime)
 import Data.Traversable (for)
-import Fmt (Buildable (..), indentF, listF, maybeF, nameF, blockListF)
+import Fmt (Buildable (..), maybeF, nameF)
 import GHC.Exts qualified as Exts
 import GHC.Read (Read (readPrec))
 import Network.FTP.Client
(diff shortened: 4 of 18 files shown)


Sorokin-Anton:
Some changes that I maybe should do here, if this approach is ok and we want those changes at all:
- add hlint rule for `Fmt.indentF` that suggests not to use functions generating lines of spaces
- more docs/meaningful names to new functions?
- currently `indentF_` is made "from scratch" inspired by `Fmt.identF` source, and `blockListF'_` depends on `Fmt.blockListF'`, and simply deletes lines containing only spaces from it's result.  Maybe it's better to use one approach (which one)?

Sereja313 (approved):

package.yaml
```diff
@@ -39,6 +39,7 @@ default-extensions:
   - NamedFieldPuns
```
dcastro, line 39: Continuing the discussion from [here](https://github.com/serokell/xrefcheck/pull/212#discussion_r1013042345)

(This topic is not 100% strictly related the problem this PR aims to solve, but it is related to `blockListF` and `indentF`. So I think this should be solved in a separate commit in this PR, or in a separate PR altogether)


`indentF` adds a newline at the end (unless the text to be indented already ends in a newline):

```hs
$ indentF 2 "a"
"  a\n"

$ indentF 2 "a\n"
"  a\n"
```

Same thing with `blockListF`:
```
$ blockListF ["a","b"]
"- a\n- b\n"

$ blockListF ["a\n","b\n"]
"- a\n- b\n"
```

One could argue that, in general, adding a `\n` at the end is good.

However, I would argue this makes these functions unsuitable to be used in combination with string interpolation.

Consider this case:
```
λ> :{ 
[int||
aaa
#{indentF 2 "bbb"}
ccc
|]
:}
```

I would have expected the result to be

```
aaa
  bbb
ccc
```

But in reality it's:

```
aaa
  bbb

ccc
```

At the moment, to work around this, we're currently adding a `\` after `indentF` / `blockListF`. The backslash gets rid of one of the `\n` characters.

```hs
λ> :{ 
[int||
aaa
#{indentF 2 "bbb"}\
ccc
|]
:}
```

```
aaa
  bbb
ccc
```

But this is error-prone, it's very easy to forget to add a `\`.

I think we should have versions of `indentF` and `blockListF` that do not append a trailing newline, which would make them suitable to be used with string interpolation. Suggestion: we could call them `interpolateIndentF` and `interpolateBlockList`.

dcastro, line 39: > - add hlint rule for `Fmt.indentF` that suggests not to use functions generating lines of spaces

Agreed, good idea!

> - more docs/meaningful names to new functions?

:+1: 

> - currently `indentF_` is made "from scratch" inspired by `Fmt.identF` source, and `blockListF'_` depends on `Fmt.blockListF'`, and simply deletes lines containing only spaces from it's result.  Maybe it's better to use one approach (which one)?

I prefer the approach you used with `blockListF_`. Run the underlying Fmt function, iterate through all the lines, and replace all-white-space lines with `""`.

In fact, this could even be abstracted into some function of type `Builder -> Builder`, that we then reuse in both `indentF_` and `blockListF'_`

dcastro: An even more robust approach could be to just remove all trailing whitespace from every line (using `Text.stripEnd`)

Sorokin-Anton: Is it enough to have only `interpolateIndentF` and `interpolateBlockList` functions that are same time not producing trailing whitespaces nor trailing newlines?

dcastro: > Is it enough to have only interpolateIndentF and interpolateBlockListF functions that are same time not producing trailing whitespaces nor trailing newlines?

Yeah, I think so.

I think we're only using it in the context of interpolation anyway.

> I would also prefer interpolateBlockListF to take non-empty instead of lists, btw

How would you handle empty lists (e.g., printing `_fiAnchors` in `Buildable FileInfo`)?

Sorokin-Anton: Something like
```
case nonEmpty _fiAnchors of
  Just ne -> ...
  Nothing -> "-"
```

dcastro: Okay makes sense - different places will want to handle empty lists differently.

----

Alternatively, our version of `blockListF` could keep accepting normal lists, but also take an additional `Builder` parameter  to display when the list is empty.

Sorokin-Anton: Done. Choosed version with NonEmty, since there was nontrivial "null checks" before printing blockLists. Also changed reports a bit - you  can see changes in bats tests. + `xrefcheck -v` is more informative when there is no scannable files:
Was
```
=== Repository data ===
  []
```

Now
```
=== Repository data ===
  No scannable files found.
```

.hlint.yaml
```diff
@@ -110,3 +110,15 @@
...
     lhs: 'System.Console.Pretty.style'
     rhs: 'Xrefcheck.Util.Colorize.styleIfNeeded'
+- warn:
+    name: "Avoid functions that generate extra trailing newlines/whitespaces"
+    lhs: 'Fmt.indentF'
+    rhs: 'Xrefcheck.Util.Interpolate.interpolateIdentF'
+- warn:
+    name: "Avoid functions that generate extra trailing newlines/whitespaces"
+    lhs: 'Fmt.blockListF'
+    rhs: 'Xrefcheck.Util.Interpolate.blockListF'
```
dcastro, line 120:
```suggestion
    rhs: 'Xrefcheck.Util.Interpolate.interpolateBlockListF'
```

.hlint.yaml
```diff
@@ -110,3 +110,15 @@
...
+- warn:
+    name: "Avoid functions that generate extra trailing newlines/whitespaces"
+    lhs: "Fmt.blockListF'"
+    rhs: "Xrefcheck.Util.Interpolate.blockListF'"
```
dcastro, line 124:
```suggestion
    rhs: "Xrefcheck.Util.Interpolate.interpolateBlockListF'"
```

src/Xrefcheck/Util/Interpolate.hs
```diff
@@ -0,0 +1,54 @@
...
+-- Note: output always ends with newline (adds trailing newline if there wasn't one).
+stripTrailingSpaces :: Builder -> Builder
+stripTrailingSpaces
+  = fromLazyText
+  . TL.unlines
+  . map (TL.stripEnd)
+  . TL.lines
+  . toLazyText
+
+stripLastNewline :: HasCallStack => Builder -> Builder
```
dcastro, line 49: We should also add `HasCallStack` to `interpolateIndentF`, `interpolateBlockListF` and `interpolateBlockListF'`

src/Xrefcheck/Util/Interpolate.hs
```diff
@@ -0,0 +1,54 @@
+{- SPDX-FileCopyrightText: 2018-2019 Serokell <https://serokell.io>
+ -
+ - SPDX-License-Identifier: MPL-2.0
+ -}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
+module Xrefcheck.Util.Interpolate where
```
dcastro, line 7: Let's add an explicit export list. `stripTrailingSpaces` and `stripLastNewline` are not part of the public API of this module.

(And also a link to the notes I mentioned in my other comment)

```suggestion
module Xrefcheck.Util.Interpolate
  ( -- $notes
    interpolateIndentF
  , interpolateBlockListF
  , interpolateBlockListF'
  )
  where
```

src/Xrefcheck/Util/Interpolate.hs
```diff
@@ -0,0 +1,54 @@
...
+
+import Universum
+
+import Data.Text.Lazy qualified as TL
+import Data.Text.Lazy.Builder (fromLazyText, toLazyText)
+import Fmt (Buildable, Builder, blockListF, blockListF', indentF)
+
+{-# HLINT ignore "Avoid functions that generate extra trailing newlines/whitespaces" #-}
+
+-- | Like @Fmt.indentF@, but strips trailing spaces and do not adds newline.
```
dcastro, line 17:
```suggestion
-- | Like @Fmt.indentF@, but strips trailing spaces and does not add a trailing newline.
```

src/Xrefcheck/Util/Interpolate.hs
```diff
@@ -0,0 +1,54 @@
...
+--
+-- >>> interpolateIndentF 2 "a\n\nb"
+-- "  a\n\n  b"
+interpolateIndentF :: Int -> Builder -> Builder
+interpolateIndentF n b = (case TL.last (toLazyText b) of
+  '\n' -> id
+  _ ->  stripLastNewline) $ stripTrailingSpaces $ indentF n b
+  -- strips newline added by indentF
+
+-- | Like @Fmt.blockListF'@, but strips trailing spaces and do not adds newline
```
dcastro, line 31:
```suggestion
-- | Like @Fmt.blockListF'@, but strips trailing spaces and does not add a trailing newline.
```

src/Xrefcheck/Util/Interpolate.hs
```diff
@@ -0,0 +1,54 @@
...
+interpolateBlockListF' :: Text -> (a -> Builder) -> NonEmpty a -> Builder
+interpolateBlockListF' =  stripLastNewline . stripTrailingSpaces ... blockListF'
+
+-- | Like @Fmt.blockListF@, but strips trailing spaces and do not adds newline
```
dcastro, line 35:
```suggestion
-- | Like @Fmt.blockListF@, but strips trailing spaces and does not add a trailing newline.
```

src/Xrefcheck/Util/Interpolate.hs
```diff
@@ -0,0 +1,54 @@
...
+module Xrefcheck.Util.Interpolate where
+
+import Universum
+
+import Data.Text.Lazy qualified as TL
+import Data.Text.Lazy.Builder (fromLazyText, toLazyText)
+import Fmt (Buildable, Builder, blockListF, blockListF', indentF)
+
+{-# HLINT ignore "Avoid functions that generate extra trailing newlines/whitespaces" #-}
+
```
dcastro, line 16: Let's add some docs to explain why we needed to write these functions in the first place. It'll help future readers understand the context.

```suggestion

{- $notes

The `blockListF` and `indentF` frunctions from @fmt@ add a trailing newline, which makes them unsuitable for string interpolation.

Consider this case:

> [int||
> aaa
> #{indentF 2 "bbb"}
> ccc
> |]

One would reasonably expect this to produce:

> aaa
>   bbb
> ccc

But, in reality, it produces:

> aaa
>   bbb
>
> ccc

This module introduces versions of these functions that do not produce a trailing newline
and can therefore be safely used in string interpolation.
-}

```

src/Xrefcheck/Verify.hs
```diff
@@ -113,10 +114,11 @@ data WithReferenceLoc a = WithReferenceLoc
   }
```
dcastro, line 115: While we're here, let's also delete the `Buildable (VerifyResult e)` instance in line 91, I don't think it's used anywhere.

package.yaml
dcastro: Great work, well done! :+1:

.hlint.yaml
Sorokin-Anton: Oh, thanks

src/Xrefcheck/Verify.hs
Sorokin-Anton: Done

dcastro (approved):

PR #224, opened 2022-11-29
[#202] Remove poorly supported unicode symbols from the program output

### LINKED ISSUE #202: Avoid exotic unicode symbols

We're using the following characters in xrefcheck's output:

1. ➥ (U+27A5): indicates a scanning/verification error.
2. ⛀ (U+26C0): indicates a verification error for a local link
3. ⛂ (U+26C2): indicates a verification error for an external link
4. ⮚ (U+2B9A): indicates a file in the repository (when xrefcheck is invoked with `-v`)

There are some problems with this:
  * character (4) does not come a [wide font support](https://www.fileformat.info/info/unicode/char/2b9a/fontsupport.htm). Neither Ubuntu nor macOS come with any of the supported fonts, so the character won't be correcly rendered.
  * characters (2) and (3) have a better [font support](https://www.fileformat.info/info/unicode/char/26C0/fontsupport.htm), but still not good:
    * They're displayed correctly in Ubuntu. AFAIK, Ubuntu comes with the FreeSerif font.
    * But they're not displayed correctly in macOS, [it doesn't come](https://support.apple.com/en-us/HT212587) with any of the supported fonts.

(I haven't checked Windows support, we'll likely have issues there too.)

Given these problems, I think it's better to simply avoid using exotic unicode characters.
Let's use other visual cues instead (coloring, indentation, etc).

* I don't think we need characters (2) and (3). They're mainly used as a visual cue to distinguish between local and external links; but this distinction is already communicated via the `Buildable LocationType` in color.
* Character (4) doesn't seem to be needed either: we already use colors and indentation for this.
* Character (1) seems to have good support: Ubuntu via FreeSerif and macOS via Arial Unicode MS. I'm pretty sure it should also work on Windows out of the box, @Sorokin-Anton can you please double-check this?

# Acceptance criteria

* We're no longer using characters 2/3/4 as visual cues
* If character 1 is supported on Windows out of the box, we leave it alone. Otherwise, let's decide what do about it.

aeqz:
## Description

Problem: We are using unicode symbols as visual clues in the program output that are not commonly supported and are therefore not always displayed as intended.

Solution: Remove the usage of these symbols, as the program output is already using other visual clues and the result will remain understandable for the user. I also changed the expected output for tests and updated the sample output in the repository `README.md`.

## Related issue(s)

Fixes #202 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and


aeqz:
CI is failing because the new image file does not contain the appropriate copyright and licensing information. How should I address that? Maybe we can keep the file out of the repository as it was with the previous one.

Sorokin-Anton:
@aeqz looks good to me, approved (but wait for @martoon too). Answering to your question, you should add your file to `.reuse/dep5`

Sorokin-Anton (approved):

aeqz:
> Answering to your question, you should add your file to `.reuse/dep5`
@Sorokin-Anton thank you. Done!

tests/golden/check-local-refs/expected2.gold
```diff
@@ -139,7 +139,7 @@
        - link: /b/../../b.md
        - anchor: -
 
-     ⛀  Link targets a local file outside repository:
+     Link targets a local file outside repository:
         dir1/b/../../b.md
```
Martoon-00, line 143: Seems like we ended up with 3-spaces indentation at lines like this one.

Probably this made sense before because `⛀` was a wide character, but now this catches an eye. What do you think?

Martoon-00 (commented):
Looks good :+1: 

By the way, how did you construct a setup for the new image?

tests/golden/check-local-refs/expected2.gold
aeqz: Right. I didn't notice it. I can replace all these 3-space usages by 2-spaces.

Martoon-00: Nice, please do this replacement.

aeqz:
@Martoon-00 I reverse-engineered the old image and created a simple project that produces that output :sunglasses:. It is a cool exercise to get started with this tool.

aeqz:
@Martoon-00 what about including the project for the output sample under the `docs` directory? It would look like in this file tree:

```sh
docs
  output-sample
    docs
      installation.md
    README.md
    output-sample.png
```

I have tried to run `xrefcheck` with this setup and it worked fine.

aeqz:
@Martoon-00 I have squashed the commits and added the sample output sources. Let me know if it is fine and I will merge this.

Martoon-00:
> I have tried to run xrefcheck with this setup (within docs/output-sample) and it worked fine.

For me, when I run `xrefcheck` in `master`, there are plenty of errors that, I suppose, should stay there even when `xrefcheck` is run from a nested folder :thinking: 

Do you see other errors along with the errors from `docs` folder?

If yes, I'm just afraid that one day the output may get interleaved, and getting a concise screenshot may become difficult.

aeqz:
@Martoon-00 yes, running xrefcheck from the repository root reports errors, mainly from test files that contain wrong references on purpose. There is a CI step that runs xrefcheck on itself and the directories containing those files are ignored:

```sh
xrefcheck --ignore 'tests/markdowns/**/*' --ignore 'tests/golden/**/*' --ignore 'docs/output-sample/**/*'
```

In my case, if I run xrefcheck from `docs/output-sample`, I only see the errors which appear in the screenshot. Perhaps this is an unintended feature of xrefcheck?

Martoon-00:
Ah I see, right, I completely forgot about that extra folders can be excluded.

Nice, let's really put that folder in `master` then :+1:

Martoon-00 (approved):
Ah okay, this ↑ is already done what I have missed.

All looks good then, please rebase, and let's merge.

Issue #211, opened 2022-10-31
Allow uppercase in anchors


Sorokin-Anton:
# Clarification and motivation
Currently, if a file `a.md` has heading `H1`, reference `[...](a.md#H1)` is reported as 
```
  ➥  In file a.md
     bad reference (current file) at src:3:8-21:
       - text: "..."
       - link: 
       - anchor: H1
     
     ⛀  Anchor 'H1' is not present, did you mean:
         - h1 (header I) at src:11:1-4
```
But GitHub renderer supports both `h1` and `H1` as anchors here, tested [there](https://github.com/Sorokin-Anton/anchors-test)
that anchors are case insensitive. I think we should allow this, but first we need to check what happens on `GitLab`. If anchors are case-insensitive on GitLab too, we can simply convert anchors to lowercase before checking. Otherwise we should do this only if `GitHub` flavor is specified
# Acceptance criteria
We consider anchors case-insensitively on flavors that support uppercase in anchors.

dcastro:
Also, let's make sure "case-insensitivity" works well together with our features for detecting:
1. ambiguous anchors
2. similar anchors

aeqz:
> If anchors are case-insensitive on GitLab too

I have checked it and anchors seem to be case-sensitive on GitLab.

Regarding other anchor features I think that, even if the flavour is case-sensitive, the similar anchors detection should always be case-insensitive. Do you agree?

Martoon-00:
I can confirm that GitHub is case-insensitive now, at least in my browser. So let's go with the flavor-depending fix.

> Regarding other anchor features I think that, even if the flavour is case-sensitive, the similar anchors detection should always be case-insensitive

Yeah, that's a good note, having it case-insensitive unconditionally would be very cool.

PR #230, opened 2022-12-02
[#197] Canonicalize filepaths

### LINKED ISSUE #197: Refactor filepath equality checks

In the past, we've had many bugs related to path equality. A few examples (there are probably more):
* https://github.com/serokell/xrefcheck/issues/106
* https://github.com/serokell/xrefcheck/issues/195
* https://github.com/serokell/xrefcheck/issues/120
* https://github.com/serokell/xrefcheck/commit/77cba6ccf2d637325eae25585b11fedc37f69a98

We've solved them mainly by sprinkling the codebase with `normalise`, [`normaliseWithNoTrailing`](https://github.com/serokell/xrefcheck/blob/33fcfdb739bf43bd97734a99fea94b7de500cb40/src/Xrefcheck/Util.hs#L55-L56), [`expandIndirections`](https://github.com/serokell/xrefcheck/pull/174), [`dropTrailingPathSeparator`](https://github.com/serokell/xrefcheck/blob/33fcfdb739bf43bd97734a99fea94b7de500cb40/src/Xrefcheck/Scan.hs#L139-L151), [`canonizeLocalRef`](https://github.com/serokell/xrefcheck/blob/33fcfdb739bf43bd97734a99fea94b7de500cb40/src/Xrefcheck/Core.hs#L301-L306) etc.

But this approach is **extremely** error-prone.

A more principled approach might be to canonicalize filepaths using [`canonicalizePath`](https://hackage.haskell.org/package/directory-1.3.8.0/docs/System-Directory.html#v:canonicalizePath).

We should canonicalize:
  * All filepaths returned by `git ls-files`
  * All local references

Here's a suggestion (haven't actually tried it):

```diff
-newtype RepoInfo = RepoInfo (Map FilePath (Maybe FileInfo))
+newtype RepoInfo = RepoInfo (Map CanonicalPath (Maybe FileInfo))

data FileInfo = FileInfo
  { _fiReferences :: [Reference]
  , _fiAnchors    :: [Anchor]
  } 

data Reference = Reference
  { rName   :: Text
  , rLink   :: Text
  , rAnchor :: Maybe Text
  , rPos    :: Position
+ , rInfo   :: ReferenceInfo
  }


+ -- We probably won't need the type `LocationType` after this.
+data ReferenceInfo
+  = RICurrentFile
+  | RIOtherFile CanonicalPath
+  | RIExternal
+  | RIOtherProtocol

+newtype CanonicalPath = UnsafeCanonicalPath { unCanonicalPath :: FilePath }
+ 
+canonicalizePath :: FilePath -> IO CanonicalPath
+canonicalizePath = UnsafeCanonicalPath . Directory.canonicalizePath
```

Then we should be able to simply use the `Eq` and `Ord` instances as usual for file equality.

Let's review the uses of `normalise` and similar functions, and try to get rid of as many as possible.

# Acceptance criteria

* We're canonicalizing paths _at the boundaries_ (i.e. after reading filepaths from `git ls-files` and from scanning markdown files)
* We're avoiding manually "massaging" paths using `normalise/dropTrailingPathSeparator/etc` throughout the codebase

aeqz:
## Description

Problem: The current usage of filepaths is error-prone and can be simplified.

Solution: Canonicalize filepaths at the boundaries, so their management will be safer and will simplify the codebase. This has the side effect of changing the program output regarding file paths, so we should also probably decide how these will be shown after the refactor.

## Related issue(s)

Fixes #197 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and


src/Xrefcheck/Core.hs
```diff
@@ -70,15 +68,89 @@ instance Given ColorMode => Buildable Position where
...
+    hasUrlProtocol = "://" `T.isInfixOf` T.take 10 link
+    hasProtocol    = ":"   `T.isInfixOf` T.take 10 link
+
+newtype CanonicalPath = UnsafeCanonicalPath
+  { unCanonicalPath :: FilePath
+  } deriving newtype (Show, Eq, Ord)
+
+canonicalizePath :: FilePath -> IO CanonicalPath
+canonicalizePath = (UnsafeCanonicalPath . expandIndirections <$>) . Directory.canonicalizePath
+  where
```
aeqz, line 115: `System.Directory.canonicalizePath` does not handle indirections, including "." and "..", when the file does not exist, so I kept this code to expand them anyways in order to not modify the program behaviour. Otherwise, some files that pointed outside the repo will be reported as "File does not exist" instead of "File is outside the repo".

src/Xrefcheck/Core.hs
```diff
@@ -116,12 +188,10 @@ instance Monoid FileInfoDiff where
 data FileInfo = FileInfo
   { _fiReferences :: [Reference]
   , _fiAnchors    :: [Anchor]
+  , _fiPath       :: FilePath
```
aeqz, line 191: For the moment I stored also the original FilePath here to be used in the program output, but this will be revisited after the discussion.

src/Xrefcheck/Scan.hs
```diff
@@ -170,88 +168,84 @@ readDirectoryWith mode config scanner root =
...
 
-    -- Strip leading "." and trailing "/"
-    location :: FilePath
-    location =
-      if root `equalFilePath` "."
-        then ""
-        else dropTrailingPathSeparator root
-
 scanRepo
   :: MonadIO m
```
aeqz, line 182: For the moment I have kept the `FilePath` parameters of `IO` actions as they are instead of changing them to `CanonicalPath`. They can be canonicalised inside the `IO` action and the exported interface remains the same, so I am not forced to modify the tests meanwhile.

src/Xrefcheck/Verify.hs
```diff
@@ -511,88 +523,52 @@ verifyReference
...
-    isVirtual = matchesGlobPatterns root (ecIgnoreLocalRefsTo cExclusions)
+    isVirtual canonicalRoot = matchesGlobPatterns canonicalRoot (ecIgnoreLocalRefsTo cExclusions)
 
-    checkRef mAnchor referredFile = verifying $
-      unless (isVirtual referredFile) do
-        checkReferredFileIsInsideRepo referredFile
-        mFileStatus <- tryGetFileStatus referredFile
+    checkRef mAnchor referredFile canonicalRoot = verifying $
+      unless (isVirtual canonicalRoot referredFile) do
+        referredFileRelative <- case stripPrefix
```
aeqz, line 530: This allows to both check if the file is inside the repo and to obtain the relative path of the file inside the repo, so it can be used to print them nicely.

Martoon-00:
Cool, looking at your comments I like how you approach all this.

Gonna take a thorough look by Monday. Please switch to another ticket meanwhile, unless you have something else to do here before my review.

src/Xrefcheck/Scan.hs
```diff
@@ -170,88 +159,84 @@ readDirectoryWith mode config scanner root =
...
-    isDirectory = readingSystem . doesDirectoryExist
+  where
+    mscanner :: CanonicalPath -> Maybe ScanAction
+    mscanner = formatsSupport . takeExtension . unCanonicalPath
 
-    -- Get all directories from filepath.
-    getDirs :: FilePath -> [FilePath]
-    getDirs = scanl (</>) "" . splitDirectories . takeDirectory
+    getDirs :: CanonicalPath -> IO [CanonicalPath]
+    getDirs = mapM canonicalizePath
```
Martoon-00, line 220: Doing so can be very expensive, it will be run for every directory obtained from every file.

I guess in this case we can safely avoid performing actual canonicalization given that the incoming file is already canonicalized.

src/Xrefcheck/Verify.hs
```diff
@@ -511,88 +523,52 @@ verifyReference
...
+              (unCanonicalPath canonicalRoot <> [pathSeparator])
+              (unCanonicalPath referredFile) of
+                Just ps -> pure ps
+                _       -> throwError (LocalFileOutsideRepo $ unCanonicalPath referredFile)
```
Martoon-00, line 534: Btw, I believe that we can keep `CanonicalPath` type in `LocalFileOutsideRepo`, that would make the code clearer for a reader.

src/Xrefcheck/Verify.hs
Martoon-00: Interesting, let's go with it.

src/Xrefcheck/System.hs
```diff
@@ -61,25 +61,20 @@ mkGlobPattern path = do
...
-normaliseGlobPattern :: RelGlobPattern -> RelGlobPattern
-normaliseGlobPattern = RelGlobPattern . normaliseWithNoTrailing . coerce
-
-bindGlobPattern :: FilePath -> RelGlobPattern -> Glob.Pattern
-bindGlobPattern root (RelGlobPattern relPat) = readingSystem $ do
-  -- TODO [#26] try to avoid using canonicalization
-  absPat <- canonicalizePath (root </> relPat)
+bindGlobPattern :: CanonicalPath -> RelGlobPattern -> Glob.Pattern
+bindGlobPattern root (UnsafeRelGlobPattern relPat) = readingSystem $ do
+  absPat <- canonicalizePath $ unCanonicalPath root </> relPat
```
Martoon-00, line 66: Suggestion: let's extract this appending to a separate function next to `CanonicalPath`. As well as the [substraction](https://github.com/serokell/xrefcheck/pull/230/files#diff-a2387580ec67193e7aca27b551d125c87d933dd0b499daaddaba74e1e528cd71R531).

src/Xrefcheck/Core.hs
Martoon-00: Sounds good, I believe we do not effectively call `expandIndirections` more often than before.

Martoon-00 (commented):
Looks pretty good, better than what I expected to be done in this issue :+1: 

I'm sorry for the delayed review, hope it wasn't a blocker for you.

Martoon-00:
On the failing tests:

I believe that the switch from displaying absolute paths in most cases is very good.

However I wonder why in case of Windows (this CI is in GitHub actions) we get relative paths with `/`, not with `\`, despite everything is canonicalized. Maybe there is a quick answer to this? If not, no need to setup Windows locally.

Also, I find that in case of `File not found` error it may be beneficial to print an absolute path too (perhaps on a separate line), this way it may be simpler for the user to investigate the issue if it was not expected.

---

There are tests that seem broken now while they shouldn't, e.g. "24 Ignore file with broken xrefcheck annotation: config file".

And some diff seems to point to places that were invalid but now are valid, not sure how did this happen. Apparently, we had too many "Link targets a local file outside repository" errors, now it seems to work better.

Could you please regenerate the golden files and carefully commit only those changes where we had "Link targets a local file outside repository" illegally, if that does not take much effort? This should give us an opportunity to document the reasoning of this change (in the commit description) and should simplify sorting through the remaining failing tests.

src/Xrefcheck/Core.hs
```diff
@@ -77,8 +75,82 @@ data Reference = Reference
   , rAnchor :: Maybe Text
     -- ^ Section or custom anchor tag.
   , rPos    :: Position
+    -- ^ More info about the link.
+  , rInfo   :: ReferenceInfo
```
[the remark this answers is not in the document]
aeqz: I keep it because of the draft that is proposed in the issue, but in fact I thought that it was proposed in this way to avoid repeated calculations. This value can be always calculated again with just other stored data (rLink), would you prefer to not store it?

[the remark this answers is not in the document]
Martoon-00: Ah right, I completely forgot about the suggestion made in the issue, thanks for the clarification.

Good then, let's store it for sure.

src/Xrefcheck/Scan.hs
aeqz: Yes, I think that it could be done 'unsafely' as you said. By the way, I have noticed that this function is now including directories beyond the root due to the canonicalized input, so I will also fix it.

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -327,18 +326,22 @@ nodeExtractInfo fp input@(Node _ _ nSubs) = do
...
-                 t : ts -> (t, Just $ T.intercalate "#" ts)
-                 []     -> error "impossible"
-         return $ FileInfoDiff
-           (DList.singleton $ Reference {rName, rPos, rLink, rAnchor})
-           DList.empty
+      where
+      extractLink url = do
+        let rName     = nodeExtractText node
+            rPos      = toPosition pos
+            link      = if null url then rName else url
```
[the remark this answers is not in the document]
aeqz: I would have voted for aligning because I think that it is more readable, but I agree, the result is harder to maintain.

aeqz:
I have moved `CanonicalPath` related code under the `Xrefcheck.System` module, and now I will try to fix the issues related to path printing and unexpected golden test diffs.

src/Xrefcheck/Verify.hs
aeqz: I like the idea and the resulting code would be simpler, but I think that then we would have to hardcode the absolute paths that are reported in our CI tests. Currently, I have changed it to show the path relative to the repository root.

aeqz:
The golden test that is failing reports as "local file outside repository" a link that is not really outside the repository, but passes trough its root: the root directory is `dir1` and the link, relative from `dir1`, is `../dir1/file.md`.

After the refactor, its path is treated as `file.md`, so it does not report that it is outside the repository. If we want to preserve that behaviour, we should add an extra check that, before canonicalizing the file link path, follows its indirections and detects whether it passes through the root or not.

src/Xrefcheck/System.hs
```diff
@@ -36,21 +42,67 @@ readingSystem = unsafePerformIO
...
+getDirsBetweenRootAndFile :: CanonicalPath -> CanonicalPath -> [CanonicalPath]
+getDirsBetweenRootAndFile root@(UnsafeCanonicalPath rootPath) file =
+  case getRelativeChild root (takeDirectory file) of
+    Just path -> UnsafeCanonicalPath
+      <$> scanl (Posix.</>) rootPath (Posix.splitDirectories path)
+    Nothing -> []
+
+getRelativeChild :: CanonicalPath -> CanonicalPath -> Maybe FilePath
+getRelativeChild (UnsafeCanonicalPath root) (UnsafeCanonicalPath child) =
+  dropWhile Posix.isPathSeparator <$> stripPrefix root child
```
Martoon-00, line 83: Note here that we seem to need to use a cross-platform version (from `System.FilePath`), not the Posix one, since the canonicalized path already contains `\` on windows.

I suppose this is a problem coming from the previous fixes, we used a big hammer and switched to `System.FilePath.Posix` more or less everywhere in the project, because markdown links are always separated with Posix'es `/` (when they point to a local place). This is an unexpected layer of problems, but now it may be the perfect to refactor this.

Ideally, to put a clear bound on where we expect a Posix-formatted filepath and where we expect a platform-dependent filepath, I think it would be cool to add another newtype for markdown links, and clarify that they are always like in Unix systems.

Then carefully pass through all our filepath-related utilities and update them respectively.

Does it make sense?

tests/Test/Xrefcheck/LocalSpec.hs
```diff
@@ -1,25 +0,0 @@
...
-test_local_refs_canonizing = testGroup "Local refs canonizing" $
-    [ testCase "Strips ./" $
-        canonizeLocalRef "./AnchorsSpec.hs" @?= "AnchorsSpec.hs"
-
-    , testCase "Strips ././" $
-        canonizeLocalRef "././AnchorsSpec.hs" @?= "AnchorsSpec.hs"
-
-    , testCase "Leaves plain other intact" $
-        canonizeLocalRef "../AnchorsSpec.hs" @?= "../AnchorsSpec.hs"
-    ]
```
Martoon-00, line 25: And it would be cool to eventually add unit tests on our new utility functions, at least on non-trivial ones, instead of these tests. I believe this will help us a lot in the future.

And unlike this module, at least part of those tests will have to be aware of the platform they are run in.

src/Xrefcheck/Verify.hs
```diff
@@ -511,88 +512,51 @@ verifyReference
       VerifyResult [ExternalHttpTooManyRequests retryAfter] -> retryAfter
       _ -> Nothing
 
-    isVirtual = matchesGlobPatterns root (ecIgnoreLocalRefsTo cExclusions)
+    isVirtual canonicalRoot = matchesGlobPatterns canonicalRoot (ecIgnoreLocalRefsTo cExclusions)
+
+    checkRef mAnchor canonicalRoot referredFile relativeToRoot = verifying $
```
Martoon-00, line 517: Could you document what exactly the `relativeToRoot` argument stands for?

src/Xrefcheck/Verify.hs
Martoon-00: Ah yes, given that we keep relative paths there now, my suggestion does not make sense anymore.

Martoon-00 (commented):
I like the changes to the golden tests that were applied, looks neat.

A few more comments.

src/Xrefcheck/System.hs
Martoon-00: I'm actually not 100% sure that I'm correct in my understanding, but we must end up with tests covering this logic, and I hope they will reveal what is correct.

Martoon-00:
> The golden test that is failing reports as "local file outside repository" a link that is not really outside the repository, but passes trough its root: the root directory is dir1 and the link, relative from dir1, is ../dir1/file.md.
>
> After the refactor, its path is treated as file.md, so it does not report that it is outside the repository. If we want to preserve that behaviour, we should add an extra check that, before canonicalizing the file link path, follows its indirections and detects whether it passes through the root or not.

Interesting. So we got more flexible after your changes.

I think this may be helpful in case the user has auto-generated documentation, so let's try to preserve this new feature of yours.

_But_, `canonicalizePath` seems to poke the filesystem taking the prefixes of the given filepath, and I believe we must avoid touching anything outside of the repository: some users may care about files' permissions and what is being touched by third party programs. So let's make sure to add the new check in a way that excludes `canonicalizePath`s calls over links that even "temporarily" exit the repository.

src/Xrefcheck/Verify.hs
aeqz: It is quite embarrassing, because its only purpose is to keep track of the link path relative to the repository root, before canonicalizing it, so it can be used in the error report if the link points out of the repository. Maybe this can be done in a neater way?

src/Xrefcheck/System.hs
aeqz: That makes sense! I was wondering why the Posix version was being used. I will try to use the Posix version only with the code that is related to Markdown links.

Martoon-00: Good :+1: :+1:

src/Xrefcheck/System.hs
```diff
@@ -36,21 +42,67 @@ readingSystem = unsafePerformIO
...
+
+getRelativeOrAbsoluteChild :: CanonicalPath -> CanonicalPath -> FilePath
+getRelativeOrAbsoluteChild root child@(UnsafeCanonicalPath c) =
+  fromMaybe c (getRelativeChild root child)
+
+(</) :: CanonicalPath -> FilePath -> IO CanonicalPath
+UnsafeCanonicalPath p </ f = canonicalizePath $ p Posix.</> f
+infixr 5 </
```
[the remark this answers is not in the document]
aeqz: This operator is being used by our code with a Markdown file path link (POSIX syntax) on the right hand side. I was thinking about replacing here the path separator with something like

```haskell
map (\c -> 
  if isPathSeparator c
  then pathSeparator
  else c
)
```

because, in `FilePath.Windows`, both '\\' and '/' hold for `isPathSeparator`, but the `pathSeparator` value is '\\'.

Then I realised that the resulting path goes through `canonicalizePath`, which uses `normalise` that already does that.

Does this mean that the Windows problems that you mentioned are already solved? Do you have any problem in mind that could appear with this behaviour?

src/Xrefcheck/System.hs
```diff
@@ -61,25 +113,20 @@ mkGlobPattern path = do
...
-normaliseGlobPattern :: RelGlobPattern -> RelGlobPattern
-normaliseGlobPattern = RelGlobPattern . normaliseWithNoTrailing . coerce
-
-bindGlobPattern :: FilePath -> RelGlobPattern -> Glob.Pattern
-bindGlobPattern root (RelGlobPattern relPat) = readingSystem $ do
-  -- TODO [#26] try to avoid using canonicalization
-  absPat <- canonicalizePath (root </> relPat)
+bindGlobPattern :: CanonicalPath -> RelGlobPattern -> Glob.Pattern
+bindGlobPattern root (UnsafeRelGlobPattern relPat) = readingSystem $ do
+  UnsafeCanonicalPath absPat <- root </ relPat
```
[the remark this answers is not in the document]
aeqz: If the decision is up to me, I don't really mind in using the `Unsafe` constructor within the same module that defines it for helper operations on that datatype. But I can change it if you prefer.

tests/Test/Xrefcheck/LocalSpec.hs
aeqz: Right! I will add tests for our `CanonicalPath` utility functions.

aeqz:
Another drawback that I have noticed with having file paths canonicalised is that, on case-insensitive systems, if the file `example/dir/file.md` exists, then `exAMple/DIR/FILe.md` is canonicalised as `example/dir/file.md`, so we cannot report it as invalid even if we wanted a case-sensitive setting such as in #211.

This is why I have not fixed the Golden tests yet: some case-related errors that are reported in CI are not reported in my computer.

src/Xrefcheck/System.hs
```diff
@@ -36,21 +43,78 @@ readingSystem = unsafePerformIO
...
+  } deriving newtype (Show, Eq, Ord)
+
+canonicalizePath :: FilePath -> IO CanonicalPath
+canonicalizePath = fmap canonicalize . Directory.makeAbsolute
+  where
+  canonicalize :: FilePath -> CanonicalPath
+  canonicalize = UnsafeCanonicalPath
+    . expandIndirections
+    . FP.normalise
+    . FP.dropTrailingPathSeparator
```
aeqz, line 64: What do you think about do not use `System.Directory.canonicalizePath` and instead implement a weaker version like this one? In this way, we have more control over its behaviour.

src/Xrefcheck/System.hs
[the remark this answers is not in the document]
Martoon-00: Okay, this perfectly makes sense to me.

src/Xrefcheck/System.hs
Martoon-00: Looks reasonable, we want to avoid the casing change that Windows does.

Could you explain how did you get to exactly this implementation? I want to be sure that we do not miss anything important that was done by `System.FilePath.canonicalizePath`.

src/Xrefcheck/Verify.hs
Martoon-00: Thanks :+1: 

Maybe call this argument like `shownFilepath`, since in `RIFileLocal` case it is not relative to the root?

While `relativeToRoot` naming in `case rInfo of` seems fine, it exactly indicates the meaning of the path it is being assigned.

src/Xrefcheck/System.hs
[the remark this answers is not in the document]
Martoon-00: In this case it is fine, yeah. Looks like all places where Markdown file path is used to produce `CanonicalPath` are safe simply because we use `canonicalizePath` so often to construct `CanonicalPath`s.

We have to be careful about the cases when we work with Markdown file path and produce another Markdown file path. How many such places do we have? If effectively none, if this `</` is the only place where we consume Markdown file path, then probably let's abandon my idea with newtype for such links and simply mention in haddock that the right hand side of `</` can accept filepath in any format.

tests/Test/Xrefcheck/CanonicalPathSpec.hs
```diff
@@ -0,0 +1,62 @@
...
+      getDirsBetweenRootAndFile root path @?= [],
+    testCase "Intermediate directories" $ do
+      path <- canonicalizePath "./example/dir/other"
+      current <- getCurrentDirectory >>= canonicalizePath
+      getRelativeOrAbsoluteChild current <$> getDirsBetweenRootAndFile current path @?=
+        [ "./"
+        , "example"
+        , "example/dir"
+        ]
+  ]
```
Martoon-00, line 62: Some minor suggestion: `test_*` function can also be of `IO TestTree` type, you could extract `getCurrentDirectory >>= canonicalizePath` there.

Also, would be nice to group test cases by what function you test (if any particular) using `testGroup`.

src/Xrefcheck/RepoInfo.hs
```diff
@@ -0,0 +1,119 @@
+{- SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io>
+ -
+ - SPDX-License-Identifier: MPL-2.0
+ -}
+
+{-# LANGUAGE GADTs #-}
+
+module Xrefcheck.RepoInfo
```
aeqz, line 8: I have solved the file system casing problem with:

- The weaker custom canonizalisation.
- Implementing `RepoInfo` in a way that it allows to search in a case-insensitive way.

This could have been implemented by carrying boolean values together with the `Map` keys telling if they are case-insensitive or not, but then we should have been taking into account cases that should not be possible, such as comparing a case-sensitive key with a case-insensitive one.

I have implemented it with a GADT instead, trying to hide the implementation and exposing a clear interface.

src/Xrefcheck/System.hs
aeqz: I used some of the transformations that are mentioned in the `System.Directory.canonicalizePath` [documentation](https://hackage.haskell.org/package/directory-1.3.8.0/docs/System-Directory.html#v:canonicalizePath). I hope to not have missed anything important, but at least it is hidden behind this smart constructor and we can modify it later.

src/Xrefcheck/Verify.hs
```diff
@@ -423,29 +421,37 @@ verifyReference
...
-          ExternalLoc       -> checkExternalResource config rLink
-          OtherLoc          -> verifying pass
+    = retryVerification 0 $
+        if shouldCheckLocType mode rInfo
+        then case rInfo of
+          RIFileLocal -> checkRef rAnchor (riRoot repoInfo) file ""
+          RIFileRelative ->
+            let relativeToRoot = getRelativeOrAbsoluteChild (riRoot repoInfo) (takeDirectory file)
+                  </> toString rLink
+             in if hasIndirectionThroughParent relativeToRoot
```
aeqz, line 441: These checks make links that pass through root considered as being outside the repository. Is this what you requested @Martoon-00?

aeqz:
I think that, upon review, this work is mostly clear except for Windows compatibility: now `CanonicalPath` is platform-dependent, so paths in Windows are represented with the `\\` separator and this breaks our test suites. This can be appreciated in the Windows CI pipeline, where tests are failing because of mismatching of `\\` vs `\` in expected outputs.

aeqz:
Windows CI is passing. What I did is to internally store the `FilePath` inside `CanonicalPath` with the platform-dependant version, but replacing path separators by POSIX ones when retrieving the `FilePath` again.

tests/golden/check-local-refs/expected2.gold
```diff
@@ -1,145 +1,127 @@
...
 
-  ➥  In file dir1/dir2/d2f1.md
-     bad reference (relative) at src:28:1-31:
-       - text: "bad-casing-file-rel"
-       - link: D2F2.md/
-       - anchor: -
-
-     File does not exist:
-       dir1/dir2/D2F2.md/
-
```
aeqz, line 28: This kind of output differences appear because the GitHub flavour is now case-insensitive.

src/Xrefcheck/Core.hs
```diff
@@ -77,8 +74,57 @@ data Reference = Reference
...
+    -- ^ Reference to a file relative to given one, e.g. @[b](folder/file#header)@
+  deriving stock (Show, Generic)
+
+-- | Compute the 'ReferenceInfo' corresponding to a given link.
+referenceInfo :: Text -> ReferenceInfo
+referenceInfo link = case toString link of
+  [] -> RIFileLocal
+  c : _             | isPathSeparator c -> RIFileAbsolute
+  '.' : c : _       | isPathSeparator c -> RIFileRelative
+  '.' : '.' : c : _ | isPathSeparator c -> RIFileRelative
```
YuriRomanowski, line 102: Don't you like `PathSep` pattern?

src/Xrefcheck/Core.hs
```diff
@@ -77,8 +74,57 @@ data Reference = Reference
...
+    hasProtocol = ":"   `T.isInfixOf` T.take 10 link
+
+-- | Whether this is a link to external resource.
+isExternal :: ReferenceInfo -> Bool
+isExternal = \case
+  RIFileLocal -> False
+  RIFileRelative -> False
+  RIFileAbsolute -> False
+  RIExternal -> True
+  RIOtherProtocol -> False
```
YuriRomanowski, line 117: Could you please keep only `RIExternal` pattern and use wildcard for others?

aeqz: Yeah! Unless you prefer the other style for any reason, I prefer to not use a wildcard in this cases to be forced to specify the value when adding a new constructor.

src/Xrefcheck/Core.hs
aeqz: As I have been trying different approaches during the refactor, at some point I removed it but now it maybe useful again. I will try to add it and see the result!

src/Xrefcheck/RepoInfo.hs
Martoon-00: First of all, have you checked how GitHub and GitLab handle the casing of filepaths? Does it for sure work similarly to anchors handling?

---

On the GADT / existential quantification, looks neat.

I suppose there were multiple ways to adopt our code here. The clue on which option to choose can give the following question: If we add more scanners (e.g. for Haskell sources), and there references have a different format, will it work?

Given that, we probably want to carry not a sum type with `RepoInfoData`s - either case-sensitive, or case-insensitive, rather multiple `RepoInfoData`, because in the future we can have several scanners that assume different filepath format and different comparison rules.

So I'd propose the following changes:

1. As you written, we will have `newtype`s over `FilePath` to indicate its format, casing, e.t.c.
     * Each such newtype will probably have to implement `Ord` instance + a typeclass providing a way to convert `FilePath` (platform-dependent path pointing to a previously scanned file in the filesystem) to this newtype. So we will probably do conversion not from Markdown to System filepath, but vice-versa (that's a debatable point!)
     
2. Let `Reference`, `FileInfo` probably some other types that use them be polymorphic over the format of filepath. We will let different scanners return different formats of links.
    * This will require `Reference` to keep a sum type for link (local file or external link), but this seems not that bad.
    * Our Markdown scanner will return case-sensitive or case-insensitive markdown links.

3. `RepoInfo` will keep a map from type to `RepoInfoData <this type>`.

    I.e. I should be able to write `lookupRepoInfoData (Proxy @CaseInsensitivePath)` and get `RepoInfoData CaseInsensitivePath`. We can use [dependent-map](https://hackage.haskell.org/package/dependent-map-0.4.0.0/docs/Data-Dependent-Map.html) or plain `Map TypeRep SomeRepoInfoData` here.

    Depending on the reference that we are currently checking, we will extract the map with the appropriate newtype.

Good that you extracted `RepoInfo` to a separate module, this will help us to introduce an appropriate interface in a better way.

aeqz: 🤦 I tried how anchors behave in both of them, but not filepaths. It turns out that GitHub, which is case-insensitive regarding anchors, is case-sensitive regarding filepaths.

I think that it is better to undo all this `RepoInfo` part for avoiding unnecessary complexity, right? It gets solved with just the weaker custom canonizalisation version.

Martoon-00: Yes, let's undo the `RepoInfo` extraction.

But please preserve it some way, probably: move the essence of those commits to a separate branch and create a Draft pull request, then drop a link to it in this thread. I believe this extraction will likely be useful in the future.

src/Xrefcheck/System.hs
```diff
@@ -36,21 +46,118 @@ readingSystem = unsafePerformIO
...
   Just (map C.toLower -> "true") -> True
-  _                              -> False
+  _ -> False
+
+-- | A FilePath that has been canonicalized. Should be created via
+-- 'canonicalizePath'. Currently, canonical paths have been made absolute,
+-- normalised regarding the running platform (e.g. Posix or Windows), with
+-- indirections syntractically expanded as much as possible and with no
+-- trailing path separator. All this results in a weaker version than that
+-- provided by 'System.Directory.canonicalizePath'.
```
Martoon-00, line 58: Good explanation :+1: 

On the formatting: it seems common to put a very brief explanation at the first line, and then elaborate at at a separate line. Sentences about different moments are also better put on separate lines.

Note, that to make two sentences appear at separate lines in the rendered Haddock documentation, you have to separate those with a blank line in the code, like:

```hs
-- A FilePath that has been canonicalized.
--
-- Should be created via 'canonicalizePath'.
```

src/Xrefcheck/RepoInfo.hs
aeqz: Done!

https://github.com/serokell/xrefcheck/pull/243

src/Xrefcheck/Verify.hs
Martoon-00: I wanted this to be checked upon `CanonicalPath` construction, but apparently this is now not justified, my previous concern about poking files outside of the repository does not hold for `makeAbsolute`. So nevermind my previous comment.

This looks good, however compared to how this check was performed previously, we lost the behaviour that this check is not performed if the file is considered virtual. I suspect we have to preserve this behaviour.

src/Xrefcheck/System.hs
Martoon-00: Okay, I've read the sources and can agree that we don't miss anything, except for symlinks resolution that `canonicalizePath` performed.

Not like we need it, we actually better go without it. But now symlinks are not transparent for us anymore, and we have to thoroughly verify how do we handle them. Please create a ticket for this under 3.0.0 milestone.

---

Hilarious actually, `makeAbsolute` and `normalize` seem completely unnecessary, they rather spoil the work of this function than help. But if we remove them, then we effectively use nothing smart from filepath utilities.

Looking forward to seeing the moment when we strip them completely (in a separate PR).

Martoon-00 (commented):
I believe we are approaching the final of this story (chapter 1 :smile_cat:), leaving few more comments.

I wanted to ask @YuriRomanowski for a second review after those comments are addressed, but looks like he already took a look, nice.

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -407,15 +411,17 @@ textToMode ("ignore" : [x])
...
-  $ nodeExtractInfo fp
+  $ nodeExtractInfo shownFilepath
   $ commonmarkToNode [optFootnotes] [extAutolink]
   $ toStrict input
 
 markdownScanner :: MarkdownConfig -> ScanAction
-markdownScanner config path = parseFileInfo config path . decodeUtf8 <$> BSL.readFile path
+markdownScanner config canonicalRoot canonicalFile =
+  parseFileInfo config (getRelativeOrAbsoluteChild canonicalRoot canonicalFile) . decodeUtf8
+    <$> BSL.readFile (unCanonicalPath canonicalFile)
```
YuriRomanowski, line 424: I think that the way of obtaining `shownName` value will not be different depending on the file format, it seems that it would make more sense just to pass `shownName` right away.

src/Xrefcheck/System.hs
```diff
@@ -36,21 +46,118 @@ readingSystem = unsafePerformIO
...
+
+-- | A 'FilePath' corresponding to the given 'CanonicalPath',
+-- always using POSIX path separators.
+unCanonicalPath :: CanonicalPath -> FilePath
+unCanonicalPath (UnsafeCanonicalPath path) =
+    fmap replaceSeparator path
+  where
+  replaceSeparator c
+    | FP.isPathSeparator c = FPP.pathSeparator
+    | otherwise = c
```
YuriRomanowski, line 70: Is there any other way to get a `FilePath` from a `CanonicalPath`? If there isn't, why not replace separators on `FilePath` creating?

src/Xrefcheck/Core.hs
```diff
@@ -77,8 +78,60 @@ data Reference = Reference
...
+  | RIFileLocal
+    -- ^ Reference to this file, e.g. @[a](#header)@
+  | RIFileAbsolute
+    -- ^ Reference to a file absolute to the root, e.g. @[c](/folder/file#header)@
+  | RIFileRelative
+    -- ^ Reference to a file relative to given one, e.g. @[b](folder/file#header)@
+  deriving stock (Show, Generic)
+
+pattern PathSep :: Char
+pattern PathSep <- (isPathSeparator -> True)
```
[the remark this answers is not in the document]
aeqz: In fact, it was not making any test fail because `isPathSeparator '/'` is `True` in both versions! (`FilePath.Posix` and `FilePath.Windows`)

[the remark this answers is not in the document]
Martoon-00: Ooh, really so!

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -256,13 +256,13 @@ removeIgnored fp = withIgnoreMode . cataNodeWithParentNodeInfo remove
...
-            tell . makeError pos fp $ ParagraphErr "EOF"
+            tell . makeError pos shownFilepath $ ParagraphErr "EOF"
             pure node
         IMSLink _ -> do
-            tell $ makeError pos fp LinkErr
+            tell $ makeError pos shownFilepath LinkErr
             pure node
         IMSAll -> do
-            tell $ makeError pos fp FileErr
+            tell $ makeError pos shownFilepath FileErr
```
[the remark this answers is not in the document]
YuriRomanowski: Unfortunately, I didn't fixed it there. But yes, we really don't need any paths here.

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -327,18 +327,22 @@ nodeExtractInfo fp input@(Node _ _ nSubs) = do
...
+        let (rLink, rAnchor) = case T.splitOn "#" link of
+              [t] -> (t, Nothing)
+              t : ts -> (t, Just $ T.intercalate "#" ts)
+              [] -> error "impossible"
+
+        let rInfo = referenceInfo rLink
+
+        return $ FileInfoDiff
+          (DList.singleton $ Reference {rName, rPos, rLink, rAnchor, rInfo})
+          DList.empty
```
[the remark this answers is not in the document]
aeqz: Right! Do you refer to the 2 space indentation for definitions within `where`? I will also fix it in other places that I wrote with this style.

src/Xrefcheck/System.hs
```diff
@@ -36,21 +46,118 @@ readingSystem = unsafePerformIO
...
-  Just "1"                       -> True
+  Just "1" -> True
   Just (map C.toLower -> "true") -> True
-  _                              -> False
+  _ -> False
+
+-- | A FilePath that has been canonicalized. Should be created via
+-- 'canonicalizePath'. Currently, canonical paths have been made absolute,
+-- normalised regarding the running platform (e.g. Posix or Windows), with
+-- indirections syntractically expanded as much as possible and with no
```
YuriRomanowski, line 56: Hm, what is "syntractically"? Is it a typo?

src/Xrefcheck/Scanners/Markdown.hs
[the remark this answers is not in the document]
aeqz: What do you prefer? I can try to remove them here.

src/Xrefcheck/System.hs
aeqz: I would prefer to keep the `CanonicalPath` contents with the platform dependant path separator, so they can be used as usual for file system related operations. I'm going to try to move this replacement to a more explicit part.

src/Xrefcheck/System.hs
```diff
@@ -36,21 +45,118 @@ readingSystem = unsafePerformIO
...
+getDirsBetweenRootAndFile :: CanonicalPath -> CanonicalPath -> [CanonicalPath]
+getDirsBetweenRootAndFile (UnsafeCanonicalPath rootPath) file =
+  case stripPrefix rootPath (unCanonicalPath (takeDirectory file)) of
+    Just path -> UnsafeCanonicalPath <$> scanl (FP.</>) rootPath directories
+      where
+        directories = FP.splitDirectories $ dropWhile FP.isPathSeparator path
+    Nothing -> []
+
+-- | Get a relative 'FilePath' from the second given path (child) with
+-- repect to the first one (root).
```
YuriRomanowski, line 106: *respect

src/Xrefcheck/System.hs
```diff
@@ -36,21 +45,118 @@ readingSystem = unsafePerformIO
...
+  where
+    replaceSeparator :: Char -> Char
+    replaceSeparator c
+      | FP.isPathSeparator c = FPP.pathSeparator
+      | otherwise = c
+
+    prepare :: FilePath -> FilePath
+    prepare path = case dropWhile FP.isPathSeparator path of
+      "" -> "."
+      other -> other
```
YuriRomanowski, line 124: IMO this name is confusing a bit, prepare for what? Wouldn't it be better to call it `dropPathSeparators`?

src/Xrefcheck/Scanners/Markdown.hs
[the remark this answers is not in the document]
YuriRomanowski: OK, so please remove it here

src/Xrefcheck/System.hs
aeqz: I will give it a proper name. It also handles specially the empty case, returning `"."` instead of `""`.

src/Xrefcheck/Scan.hs
```diff
@@ -67,35 +66,58 @@ data ExclusionConfig' f = ExclusionConfig
...
-  { sePosition    :: Position
-  , seFile        :: FilePath
+-- | A scan error indexed by different process stages.
+--
+-- Within 'Parse', 'seFile' has no information because the same
+-- file is being parsed.
+--
+-- Within 'Gather', 'seFile' stores the 'FilePath' corresponding
+-- to the file in where the error was found.
+data ScanError (a :: ScanStage) = ScanError
```
aeqz, line 90: Do you think that this 'Trees that grow' style fits in here?

src/Xrefcheck/Verify.hs
```diff
@@ -423,29 +421,31 @@ verifyReference
...
-          AbsoluteLoc       -> checkRef rAnchor (root <> toString rLink)
-          ExternalLoc       -> checkExternalResource config rLink
-          OtherLoc          -> verifying pass
+    = retryVerification 0 $
+        if shouldCheckLocType mode rInfo
+        then case rInfo of
+          RIFileLocal -> checkRef rAnchor riRoot file ""
+          RIFileRelative -> do
+            let relativeToRoot = getPosixRelativeOrAbsoluteChild riRoot (takeDirectory file)
+                  </> toString rLink
```
YuriRomanowski, line 440: Well, AFAIU, this is not always relative, from the function name. But even better it would be to call it `shownPath`, like inside the `checkRef` function, it seems that it doesn't really matter that it is relative.

tests/Test/Xrefcheck/CanonicalPathSpec.hs
```diff
@@ -0,0 +1,63 @@
...
+import Xrefcheck.System
+
+test_canonicalPath :: IO TestTree
+test_canonicalPath = do
+  current <- getCurrentDirectory >>= canonicalizePath
+  return $ testGroup "Canonical paths"
+    [ testGroup "Canonicalization"
+        [ testCase "Trailing separator" $ do
+            path <- canonicalizePath "./example/dir/"
+            getPosixRelativeOrAbsoluteChild current path @?= "example/dir",
```
YuriRomanowski, line 23: Please format it like this:
```
exceptions =
  [ InvalidStatusCode
  , MissingContentHeader
  , InternalServerError
  ]
```

YuriRomanowski (approved):

YuriRomanowski:
I'm perfectly fine with this PR, titanic work! I approved it.

aeqz:
I will wait also for @Martoon-00 's review. And then, get ready for conflicts in our currently open PRs 🙊

src/Xrefcheck/Scanners/Markdown.hs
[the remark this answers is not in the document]
Martoon-00: Yes, better to provide an indentation after `where` word.

src/Xrefcheck/System.hs
Martoon-00: For the record: https://github.com/serokell/xrefcheck/issues/244 has been created.

Martoon-00:
Oh, looks like `hasIndirectionThroughParent` is gone now, why so?

aeqz:
> Oh, looks like hasIndirectionThroughParent is gone now, why so?

I added `hasIndirectionThroughParent`, but then removed it because of the following comment:

> my previous concern about poking files outside of the repository does not hold for makeAbsolute. So nevermind my previous comment.

Do we need it for something that I have missed? Or should I keep it just in case it may be useful in the future?

src/Xrefcheck/Scan.hs
Martoon-00: Oh, I like your type families solution here, it makes particular sense. At your place, I would probably write it in the same way :)

However I think that two separate datatypes would work better here, and here's why: our core part serves as sort of API for potential scanners developers, and later there will be a lot of code that works with `ScanError 'Parse`.

And with the current approach, we win at the core side (fewer number of datatypes) and lose at the parsers' side (one has to always initialize that `seFile = ()`), and the latter is more significant for the code amount and API-wise.

Martoon-00:
> Do we need it for something that I have missed?

Ah, I didn't put my thought concrete.

My comment "about poking files outside of the repository" only affected where `hasIndirectionThroughParent` check had to be performed, that we shouldn't run `canonicalizePath` on paths that are yet unchecked on indirection.

But checking for indirections has to be there in either way, we cannot throw this check away altogether.

My subsequent comment mentioning `virtual` files meant that we probably want to check on indirections inside `checkRef` function, at the same position where this check was located before this PR.

aeqz:
@Martoon-00 right! I misunderstood it. Now the Golden tests output shows the same behaviour as before the refactor regarding both virtual files and links that pass through the repository root.

src/Xrefcheck/Scan.hs
aeqz: Would it be an option to use `mkParseScanError ` instead of initialising `seFile` to `()`? I can also give type aliases to `ScanError 'Parse` and `ScanError 'Gather` also.

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          17 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 5309f71ac2707cd646d2bae59b3d269f0ca362af lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 742aa00fe79c7efdbe4f70e8ec0d8e0e32b2dfb1 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 5e3baa6b7169fad8baef478f16e9e128a8253f52 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit bc9c5e4255c5a10d499a22e421a584a69ff340ca lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 3693423bb7d0cd3e10b18b9ae587b3f697762ff9 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit a07b344fdf2c1f44abfd085f8aed7af35dbcb538 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit e199b478ddf37351d970ed8de4e2aa1385a20e18 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 7110bae51d4eb0a43aca0ddc4b04912730d7670a lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit e188763bcec7f97bf5928d5b36f98c73bcbe0311 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit a7ba91c628706685c359b60c128d3db375a5711b lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 32d2180f5d1746cfc7872d7143cf6381b6a6de22 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit aee28d6ab7d00a29ef3f95914b6f4c41e67e0ffd lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit a5a1dcf574ab2b5ab646a38712e938fb76970009 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 6433d56bb8cd0092bec0af85dfd3e464b95cf6a5 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 6698151da85965a2798f5751b4513f688110b19b lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 86c160769b790be07868c5940967f0a0a8cf1171 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 788e1808554185037fa8d17e9297340d884c3c2c lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">In 7110bae51d4eb0a43aca0ddc4b04912730d7670a subject does not begin with an uppercase letter: <code>[#197] recover PathSep</code>.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please, only use letters, digits and dashes in the branch name.<br />
        Found: ["_"]</td>
    </tr>
  </tbody>
</table>

</p>

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

src/Xrefcheck/Scan.hs
Martoon-00: Oh right, we already use a smart constructor there, so avoiding `()` each time won't be a problem.

Okay, let's go with it.

Martoon-00:
Feel free to prettify the commits history, I'll make the final review pass then.

Martoon-00 (commented):
Thank you for your titanic work!

I believe we achieved quite a lot, both from the perspective of code cleanness and in how much better the output now looks like.

I'm leaving one more comment about commits history prettification, and going to approve after that.

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -327,18 +322,22 @@ nodeExtractInfo fp input@(Node _ _ nSubs) = do
...
+          let (rLink, rAnchor) = case T.splitOn "#" link of
+                [t] -> (t, Nothing)
+                t : ts -> (t, Just $ T.intercalate "#" ts)
+                [] -> error "impossible"
+
+          let rInfo = referenceInfo rLink
+
+          return $ FileInfoDiff
+            (DList.singleton $ Reference {rName, rPos, rLink, rAnchor, rInfo})
+            DList.empty
```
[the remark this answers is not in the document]
aeqz: I have separated it in between the CI fix commit and the refactor one.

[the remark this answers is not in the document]
Martoon-00: Nice :+1:

Martoon-00 (approved):

Martoon-00:
Hmhm, danger complains about `Style` tags.

But we can actually allow them. You can update `danger/helpers.rb` to allow both `Chore` and `Style` tags.

PR #234, opened 2022-12-07
[#217] Retry on response timeout

### LINKED ISSUE #217: Retry on response timeout

When we hammer a service like `github.com`, it _usually_ starts replying with `429 Too Many Requests`. `xrefcheck` knows how to handle that.

However, other times, we may also start getting `Reponse timeout`.

```

  ➥  In file README.md
     bad reference (external) at src:22:10-83:
       - text: "latest release"
       - link: https://github.com/tezos-commons/baseDAO/releases/latest
       - anchor: -
     
     ⛂  Response timeout
     
     
     
  ➥  In file README.md
     bad reference (external) at src:51:155-233:
       - text: "our tezos-packaging repository"
       - link: https://github.com/serokell/tezos-packaging
       - anchor: -
     
     ⛂  Response timeout
```

This happens often enough to make the CI fail at an annoying rate.

I think we should handle retries on timeout.

Let's also add a config option in `.xrefcheck.yaml` for specifying how many times to retry on timeouts. A default value of `1` seems sensible enough.

# Acceptance criteria

* By default, `xrefcheck` retries a link (once) when it times out.
* The number of retries is configurable in `.xrefcheck.yaml` (NB: it can be set to 0)

YuriRomanowski:
## Description

Problem: Currently, getting response timeout immediately results in fail, it's desired to have a possibility to configure retries on timeouts.

Solution: The new ExternalHttpTimeout error is added, which is treated in a similar way as the ExternalHttpTooManyRequests error. A new field is added to the config meaning how many timeouts are allowed. Default value equals to 1.

## Related issue(s)

Fixes #217 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

#### ✓ Release Checklist

### DIFF
--- a/src/Xrefcheck/Config.hs
+++ b/src/Xrefcheck/Config.hs
@@ -54,6 +54,30 @@ data NetworkingConfig' f = NetworkingConfig
   , ncMaxRetries                :: Field f Int
     -- ^ How many attempts to retry an external link after getting
     -- a "429 Too Many Requests" response.
+    -- Timeouts may also be accounted here, see the description
+    -- of `maxTimeoutRetries` field.
+    --
+    -- If a site once responded with 429 error code, subsequent
+    -- request timeouts will also be treated as hitting the site's
+    -- rate limiter and result in retry attempts, unless the
+    -- maximum retries number has been reached.
+    --
+    -- On other errors xrefcheck fails immediately, without retrying.
+  , ncMaxTimeoutRetries         :: Field f Int
+    -- ^ Querying a given domain that ever returned 429 before,
+    -- this defines how many timeouts are allowed during retries.
+    --
+    -- For such domains, timeouts likely mean hitting the rate limiter,
+    -- and so xrefcheck considers timeouts in the same way as 429 errors.
+    --
+    -- For other domains, a timeout results in a respective error, no retry
+    -- attempts will be performed. Use `externalRefCheckTimeout` option
+    -- to increase the time after which timeout is declared.
+    --
+    -- This option is similar to `maxRetries`, the difference is that
+    -- this `maxTimeoutRetries` option limits only the number of retries
+    -- caused by timeouts, and `maxRetries` limits the number of retries
+    -- caused both by 429s and timeouts.
   } deriving stock (Generic)
 
 -- | Type alias for ScannersConfig' with all required fields.
@@ -113,6 +137,7 @@ overrideConfig config
         , ncIgnoreAuthFailures        = overrideField ncIgnoreAuthFailures
         , ncDefaultRetryAfter         = overrideField ncDefaultRetryAfter
         , ncMaxRetries                = overrideField ncMaxRetries
+        , ncMaxTimeoutRetries         = overrideField ncMaxTimeoutRetries
         }
       where
         overrideField :: (forall f. NetworkingConfig' f -> Field f a) -> a
--- a/src/Xrefcheck/Config/Default.hs
+++ b/src/Xrefcheck/Config/Default.hs
@@ -54,7 +54,32 @@ networking:
 
   # How many attempts to retry an external link after getting
   # a "429 Too Many Requests" response.
+  # Timeouts may also be accounted here, see the description
+  # of `maxTimeoutRetries` field.
+
+  # If a site once responded with 429 error code, subsequent
+  # request timeouts will also be treated as hitting the site's
+  # rate limiter and result in retry attempts, unless the
+  # maximum retries number has been reached.
+  #
+  # On other errors xrefcheck fails immediately, without retrying.
   maxRetries: 3
+
+# Querying a given domain that ever returned 429 before,
+# this defines how many timeouts are allowed during retries.
+#
+# For such domains, timeouts likely mean hitting the rate limiter,
+# and so xrefcheck considers timeouts in the same way as 429 errors.
+#
+# For other domains, a timeout results in a respective error, no retry
+# attempts will be performed. Use `externalRefCheckTimeout` option
+# to increase the time after which timeout is declared.
+#
+# This option is similar to `maxRetries`, the difference is that
+# this `maxTimeoutRetries` option limits only the number of retries
+# caused by timeouts, and `maxRetries` limits the number of retries
+# caused both by 429s and timeouts.
+  maxTimeoutRetries: 1
 
 # Parameters of scanners for various file types.
 scanners:
--- a/src/Xrefcheck/Verify.hs
+++ b/src/Xrefcheck/Verify.hs
@@ -9,7 +9,6 @@
 module Xrefcheck.Verify
   ( -- * General verification
     VerifyResult (..)
-  , verifyOk
   , verifyErrors
   , verifying
 
@@ -21,6 +20,7 @@ module Xrefcheck.Verify
   , forConcurrentlyCaching
 
     -- * Cross-references validation
+  , DomainName (..)
   , VerifyError (..)
   , verifyRepo
   , verifyReference
@@ -43,6 +43,7 @@ import Data.List (lookup)
 import Data.List qualified as L
 import Data.Map qualified as M
 import Data.Reflection (Given)
+import Data.Set qualified as S
 import Data.Text (toCaseFold)
 import Data.Text.Metrics (damerauLevenshteinNorm)
 import Data.Time (UTCTime, defaultTimeLocale, formatTime, readPTime, rfc822DateFormat)
@@ -66,7 +67,7 @@ import System.FilePath.Posix ((</>))
 import Text.Interpolation.Nyan
 import Text.ParserCombinators.ReadPrec qualified as ReadPrec (lift)
 import Text.Regex.TDFA.Text (Regex, regexec)
-import Text.URI (Authority (..), ParseExceptionBs, URI (..), mkURIBs)
+import Text.URI (Authority (..), ParseExceptionBs, URI (..), mkURIBs, unRText)
 import Time (RatioNat, Second, Time (..), ms, sec, threadDelay, timeout, (+:+), (-:-))
 import URI.ByteString qualified as URIBS
 
@@ -91,9 +92,6 @@ newtype VerifyResult e = VerifyResult [e]
 
 deriving newtype instance Semigroup (VerifyResult e)
 deriving newtype instance Monoid (VerifyResult e)
-
-verifyOk :: VerifyResult e -> Bool
-verifyOk (VerifyResult errors) = null errors
 
 verifyErrors :: VerifyResult e -> Maybe (NonEmpty e)
 verifyErrors (VerifyResult errors) = nonEmpty errors
@@ -121,6 +119,13 @@ instance (Given ColorMode, Buildable a) => Buildable (WithReferenceLoc a) where
     #{wrlItem}
     |]
 
+-- | Contains a name of a domain, examples:
+-- @DomainName "github.com"@,
+-- @DomainName "localhost"@,
+-- @DomainName "192.168.0.104"@
+newtype DomainName = DomainName { unDomainName :: Text }
+  deriving stock (Show, Eq, Ord)
+
 data VerifyError
   = LocalFileDoesNotExist FilePath
   | LocalFileOutsideRepo FilePath
@@ -132,7 +137,8 @@ data VerifyError
   | ExternalResourceInvalidUrl (Maybe Text)
   | ExternalResourceUnknownProtocol
   | ExternalHttpResourceUnavailable Status
-  | ExternalHttpTooManyRequests (Maybe RetryAfter)
+  | ExternalHttpTooManyRequests (Maybe RetryAfter) (Maybe DomainName)
+  | ExternalHttpTimeout (Maybe DomainName)
   | ExternalFtpResourceUnavailable FTPResponse
   | ExternalFtpException FTPException
   | FtpEntryDoesNotExist FilePath
@@ -214,9 +220,14 @@ instance Given ColorMode => Buildable VerifyError where
       Resource unavailable (#{statusCode status} #{decodeUtf8 @Text (statusMessage status)})
       |]
 
-    ExternalHttpTooManyRequests retryAfter ->
+    ExternalHttpTooManyRequests retryAfter _ ->
       [int||
       Resource unavailable (429 Too Many Requests; retry after #{maybeF retryAfter})
+      |]
+
+    ExternalHttpTimeout _ ->
+      [int||
+      Response timeout
       |]
 
     ExternalFtpResourceUnavailable response ->
@@ -255,6 +266,21 @@ instance Given ColorMode => Buildable VerifyError where
         #{redirectedUrl}
       |]
 
+data RetryCounter = RetryCounter
+  { rcTotalRetries   :: Int
+  , rcTimeoutRetries :: Int
+  } deriving stock (Show)
+
+errorsOccured :: RetryCounter -> Bool
+errorsOccured rc = rcTotalRetries rc > 0
+
+incTotalCounter :: RetryCounter -> RetryCounter
+incTotalCounter rc = rc {rcTotalRetries = rcTotalRetries rc + 1}
+
+incTimeoutCounter :: RetryCounter -> RetryCounter
+incTimeoutCounter rc = rc {rcTimeoutRetries = rcTimeoutRetries rc + 1}
+
+
 reportVerifyErrs
   :: Given ColorMode => NonEmpty (WithReferenceLoc VerifyError) -> IO ()
 reportVerifyErrs errs = fmt
@@ -278,11 +304,6 @@ instance Buildable RetryAfter where
   build (Date d) = nameF "date" $
     fromString $ formatTime defaultTimeLocale rfc822DateFormat d
   build (Seconds s) = nameF "seconds" $ show s
-
--- | Determine whether the verification result contains a fixable error.
-isFixable :: VerifyError -> Bool
-isFixable (ExternalHttpTooManyRequests _) = True
-isFixable _ = False
 
 data NeedsCaching key
   = NoCaching
@@ -373,10 +394,10 @@ verifyRepo
                                  -- added to Git while gathering RepoInfo.
 
   progressRef <- newIORef $ initVerifyProgress (map snd toScan)
-
+  domainsReturned429Ref <- newIORef S.empty
   accumulated <- loopAsyncUntil (printer progressRef) do
     forConcurrentlyCaching toScan ifExternalThenCache $ \(file, ref) ->
-      verifyReference config mode progressRef repoInfo file ref
+      verifyReference config mode domainsReturned429Ref progressRef repoInfo file ref
   case accumulated of
     Right res -> return $ fold res
     Left (exception, partialRes) -> do
@@ -384,7 +405,7 @@ verifyRepo
       let errs = verifyErrors (fold partialRes)
           total = length toScan
           checked = length partialRes
-      whenJust errs $ reportVerifyErrs
+      whenJust errs reportVerifyErrs
       fmt [int|A|
           Interrupted (#s{exception}), checked #{checked} out of #{total} references.
           |]
@@ -419,6 +440,7 @@ shouldCheckLocType mode locType
 verifyReference
   :: Config
   -> VerifyMode
+  -> IORef (S.Set DomainName)
   -> IORef VerifyProgress
   -> RepoInfo
   -> CanonicalPath
@@ -427,109 +449,132 @@ verifyReference
 verifyReference
   config@Config{..}
   mode
+  domainsReturned429Ref
   progressRef
   repoInfo@RepoInfo{..}
   file
   ref@Reference{..}
-    = retryVerification 0 $
+    = fmap (fmap addReference . toVerifyRes) $
+      retryVerification (RetryCounter 0 0) $ runExceptT $
         if shouldCheckLocType mode rInfo
         then case rInfo of
           RIFileLocal -> checkRef rAnchor riRoot file ""
           RIFileRelative -> do
             let shownFilepath = getPosixRelativeOrAbsoluteChild riRoot (takeDirectory file)
                   </> toString rLink
-            canonicalPath <- takeDirectory file </ toString rLink
+            canonicalPath <- liftIO $ takeDirectory file </ toString rLink
             checkRef rAnchor riRoot canonicalPath shownFilepath
           RIFileAbsolute -> do
             let shownFilepath = dropWhile isPathSeparator (toString rLink)
-            canonicalPath <- riRoot </ shownFilepath
+            canonicalPath <- liftIO $ riRoot </ shownFilepath
             checkRef rAnchor riRoot canonicalPath shownFilepath
           RIExternal -> checkExternalResource config rLink
-          RIOtherProtocol -> verifying pass
-        else return mempty
+          RIOtherProtocol -> pass
+        else pass
   where
+    addReference :: VerifyError -> WithReferenceLoc VerifyError
+    addReference = WithReferenceLoc (getPosixRelativeOrAbsoluteChild riRoot file) ref
+
     retryVerification
-      :: Int
-      -> IO (VerifyResult VerifyError)
-      -> IO (VerifyResult $ WithReferenceLoc VerifyError)
-    retryVerification numberOfRetries resIO = do
-      res@(VerifyResult ves) <- resIO
-
+      :: RetryCounter
+      -> IO (Either VerifyError ())
+      -> IO (Either VerifyError ())
+    retryVerification rc resIO = do
+      let errorsOccured_ = errorsOccured rc
+      res <- resIO
       now <- getPOSIXTime <&> posixTimeToTimeSecond
+      case res of
+        -- Success
+        Right () -> do
+          let moveProgressOnSuccess =
+                if errorsOccured_
+                then decProgressFixableErrors
+                else incProgress
+          modifyProgressRef now Nothing moveProgressOnSuccess
+          pure res
+        Left err -> do
+          setOfReturned429 <- addDomainIf429 domainsReturned429Ref err
+          case decideWhetherToRetry setOfReturned429 rc err of
+            -- Unfixable
+            Nothing -> do
+              let moveProgressOnUnfixable = composeFuncList
+                    [ unlessFunc errorsOccured_ incProgress
+                    , if errorsOccured_
+                      then fixableToUnfixable
+                      else incProgressUnfixableErrors
+                    ]
+              modifyProgressRef now Nothing moveProgressOnUnfixable
+              pure res
+            -- Fixable, retry
+            Just (mbCurrentRetryAfter, counterModifier) -> do
+              let toSeconds = \case
+                    Seconds s -> s
+                    -- Calculates the seconds left until @Retry-After@ date.
+                    -- Defaults to 0 if the date has already passed.
+                    Date date | utcTimeToTimeSecond date >= now -> utcTimeToTimeSecond date -:- now
+                    _ -> sec 0
 
-      let toSeconds = \case
-            Seconds s -> s
-            -- Calculates the seconds left until @Retry-After@ date.
-            -- Defaults to 0 if the date has already passed.
-            Date date | utcTimeToTimeSecond date >= now -> utcTimeToTimeSecond date -:- now
-            _ -> sec 0
+              let currentRetryAfter = fromMaybe (ncDefaultRetryAfter cNetworking) $
+                    fmap toSeconds mbCurrentRetryAfter
 
-      let toRetry = any isFixable ves && numberOfRetries < ncMaxRetries cNetworking
-          currentRetryAfter = fromMaybe (ncDefaultRetryAfter cNetworking) $
-            extractRetryAfterInfo res <&> toSeconds
+              let moveProgressOnFixable = whenFunc (not errorsOccured_) $
+                    composeFuncList
+                    [ incProgressFixableErrors
+                    , incProgress
+                    ]
+              modifyProgressRef now (Just currentRetryAfter) moveProgressOnFixable
+              threadDelay currentRetryAfter
+              retryVerification
+                (counterModifier rc)
+                resIO
 
-      let moveProgress = alterOverallProgress numberOfRetries
-                       . alterProgressErrors res numberOfRetries
+    modifyProgressRef :: Time Second -> Maybe (Time Second) -> (Progress Int -> Progress Int) -> IO ()
+    modifyProgressRef now mbRetryAfter moveProgress = atomicModifyIORef' progressRef $ \VerifyProgress{..} ->
+      ( if isExternal rInfo
+        then VerifyProgress{ vrExternal =
+          let vrExternalAdvanced = moveProgress vrExternal
+          in case mbRetryAfter of
+             Just retryAfter -> case pTaskTimestamp vrExternal of
+                    Just (TaskTimestamp ttc start)
+                      | retryAfter +:+ now <= ttc +:+ start -> vrExternalAdvanced
+                    _ -> setTaskTimestamp retryAfter now vrExternalAdvanced
+             Nothing -> vrExternalAdvanced, .. }
+        else VerifyProgress{ vrLocal = moveProgress vrLocal, .. }
+      , ()
+      )
 
-      atomicModifyIORef' progressRef $ \VerifyProgress{..} ->
-        ( if isExternal rInfo
-          then VerifyProgress{ vrExternal =
-            let vrExternalAdvanced = moveProgress vrExternal
-            in if toRetry
-               then case pTaskTimestamp vrExternal of
-                      Just (TaskTimestamp ttc start)
-                        | currentRetryAfter +:+ now <= ttc +:+ start -> vrExternalAdvanced
-                      _ -> setTaskTimestamp currentRetryAfter now vrExternalAdvanced
-               else vrExternalAdvanced, .. }
-          else VerifyProgress{ vrLocal = moveProgress vrLocal, .. }
-        , ()
-        )
-      if toRetry
-      then do
-        threadDelay currentRetryAfter
-        retryVerification (numberOfRetries + 1) resIO
-      else return . (<$> res) $
-        WithReferenceLoc (getPosixRelativeOrAbsoluteChild riRoot file) ref
+    addDomainIf429 :: IORef (S.Set DomainName) -> VerifyError -> IO (S.Set DomainName)
+    addDomainIf429 setRef err = atomicModifyIORef' setRef $ \s ->
+      (\x -> (x, x)) $ case err of
+      ExternalHttpTooManyRequests _ mbDomain ->
+        maybe s (flip S.insert s) mbDomain
+      _ -> s
 
-    alterOverallProgress
-      :: (Num a)
-      => Int
-      -> Progress a
-      -> Progress a
-    alterOverallProgress retryNumber
-      | retryNumber > 0 = id
-      | otherwise = incProgress
+    decideWhetherToRetry
+      :: S.Set DomainName
+      -> RetryCounter
+      -> VerifyError
+      -> Maybe (Maybe RetryAfter, RetryCounter -> RetryCounter)
+    decideWhetherToRetry setOfReturned429 rc = \case
+      ExternalHttpTooManyRequests retryAfter _
+        | totalRetriesNotExceeded -> Just (retryAfter, incTotalCounter)
+      ExternalHttpTimeout (Just domain)
+        | totalRetriesNotExceeded && timeoutRetriesNotExceeded ->
+          -- If a given domain ever returned 429 error, we assume that getting timeout from
+          -- the domain can be considered as a 429-like error, and hence we retry.
+          -- If there was no 429 responses from this domain, then getting timeout from
+          -- it probably means that this site is not working at all.
+          -- Also, there always remains a possibility that we just didn't get the response
+          -- in time, but we can't avoid this case here, the only thing that can help
+          -- is to increase the allowed timeout in the config.
 
-    alterProgressErrors
-      :: (Num a)
-      => VerifyResult VerifyError
-      -> Int
-      -> Progress a
-      -> Progress a
-    alterProgressErrors res@(VerifyResult ves) retryNumber
-      | (ncMaxRetries cNetworking) == 0 =
-          if ok then id
-          else incProgressUnfixableErrors
-      | retryNumber == 0 =
-          if ok then id
-          else if fixable then incProgressFixableErrors
-          else incProgressUnfixableErrors
-      | retryNumber == (ncMaxRetries cNetworking) =
-          if ok then decProgressFixableErrors
-          else fixableToUnfixable
-      -- 0 < retryNumber < ncMaxRetries
-      | otherwise =
-          if ok then decProgressFixableErrors
-          else if fixable then id
-          else fixableToUnfixable
-      where
-        ok = verifyOk res
-        fixable = any isFixable ves
-
-    extractRetryAfterInfo :: VerifyResult VerifyError -> Maybe RetryAfter
-    extractRetryAfterInfo = \case
-      VerifyResult [ExternalHttpTooManyRequests retryAfter] -> retryAfter
+          if S.member domain setOfReturned429
+          then Just (Just (Seconds $ sec 0), incTimeoutCounter . incTotalCounter)
+          else Nothing
       _ -> Nothing
+      where
+        totalRetriesNotExceeded = rcTotalRetries rc < ncMaxRetries cNetworking
+        timeoutRetriesNotExceeded = rcTimeoutRetries rc < ncMaxTimeoutRetries cNetworking
 
     isVirtual canonicalRoot = matchesGlobPatterns canonicalRoot (ecIgnoreLocalRefsTo cExclusions)
 
@@ -540,7 +585,8 @@ verifyReference
     -- so it allows indirections and should be suitable for being shown to
     -- the user. Also, it will be considered as outside the repository if
     -- it is relative and its idirections pass through the repository root.
-    checkRef mAnchor canonicalRoot referredFile shownFilepath = verifying $
+    checkRef :: Maybe Text -> CanonicalPath -> CanonicalPath -> FilePath -> ExceptT VerifyError IO ()
+    checkRef mAnchor canonicalRoot referredFile shownFilepath =
       unless (isVirtual canonicalRoot referredFile) do
         when (hasIndirectionThroughParent shownFilepath) $
           throwError $ LocalFileOutsideRepo shownFilepath
@@ -642,10 +688,10 @@ parseUri link = do
            & handleJust (fromException @ParseExceptionBs)
            (throwError . ExternalResourceUriConversionError)
 
-checkExternalResource :: Config -> Text -> IO (VerifyResult VerifyError)
+checkExternalResource :: Config -> Text -> ExceptT VerifyError IO ()
 checkExternalResource Config{..} link
-  | isIgnored = return mempty
-  | otherwise = fmap toVerifyRes $ runExceptT $ do
+  | isIgnored = pass
+  | otherwise = do
       uri <- parseUri link
       case toString <$> uriScheme uri of
         Just "http" -> checkHttp uri
@@ -670,7 +716,7 @@ checkExternalResource Config{..} link
 
     checkHttp :: URI -> ExceptT VerifyError IO ()
     checkHttp uri = makeHttpRequest uri HEAD 0.3 `catchError` \case
-      e | isFixable e -> throwError e
+      e@(ExternalHttpTooManyRequests _ _) -> throwError e
       _ -> makeHttpRequest uri GET 0.7
 
     httpConfig :: HttpConfig
@@ -700,8 +746,12 @@ checkExternalResource Config{..} link
       let maxTime = Time @Second $ unTime ncExternalRefCheckTimeout * timeoutFrac
 
       mres <- liftIO (timeout maxTime $ void reqLink) `catch`
-        (either throwError (\() -> return (Just ())) . interpretErrors)
-      maybe (throwError $ ExternalResourceSomeError "Response timeout") pure mres
+        (either throwError (\() -> return (Just ())) . interpretErrors uri)
+      maybe (throwError $ ExternalHttpTimeout $ extractHost uri) pure mres
+
+    extractHost :: URI -> Maybe DomainName
+    extractHost =
+      either (const Nothing) (Just . DomainName . unRText . authHost) . uriAuthority
 
     isTemporaryRedirectCode :: Int -> Bool
     isTemporaryRedirectCode = flip elem [302, 303, 307]
@@ -719,7 +769,7 @@ checkExternalResource Config{..} link
       , (405 ==)  -- method mismatch
       ]
 
-    interpretErrors = \case
+    interpretErrors uri = \case
       JsonHttpException _ -> error "External link JSON parse exception"
       VanillaHttpException err -> case err of
         InvalidUrlException{} -> error "External link URL invalid exception"
@@ -733,7 +783,7 @@ checkExternalResource Config{..} link
             | isTemporaryRedirectCode code -> Right ()
             | isAllowedErrorCode code -> Right ()
             | otherwise -> case statusCode (responseStatus resp) of
-              429 -> Left . ExternalHttpTooManyRequests $ retryAfterInfo resp
+              429 -> Left $ ExternalHttpTooManyRequests (retryAfterInfo resp) (extractHost uri)
               _ -> Left . ExternalHttpResourceUnavailable $ responseStatus resp
             where
               code = statusCode $ responseStatus resp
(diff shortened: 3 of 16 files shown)


src/Xrefcheck/Config.hs
```diff
@@ -120,6 +123,7 @@ overrideConfig config
         , ncIgnoreAuthFailures        = overrideField ncIgnoreAuthFailures
         , ncDefaultRetryAfter         = overrideField ncDefaultRetryAfter
         , ncMaxRetries                = overrideField ncMaxRetries
```
YuriRomanowski, line 125: Should we rename this somehow? It can break old configs, is it important?

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +454,103 @@ verifyReference
...
-      -> IO (VerifyResult $ WithReferenceLoc VerifyError)
-    retryVerification numberOfRetries resIO = do
-      res@(VerifyResult ves) <- resIO
+    addReference :: VerifyError -> WithReferenceLoc VerifyError
+    addReference = WithReferenceLoc fileWithReference ref
 
+    retryVerification
+      :: RetryCounter
+      -> IO (Maybe VerifyError)
+      -> IO (Maybe VerifyError)
```
YuriRomanowski, line 466: In fact we get the maximum of one error from the checking functions, so using here newtype above the list was a bit confusing IMO. Also I refactored this function in order to isolate different cases from each other and make code more readable.

aeqz: I like the resulting code after the refactor! It seems easier to understand. However, let's wait for @Martoon-00 's review because I don't know som much this code and maybe he can catch anything that could have been unintentionally changed.

src/Xrefcheck/Verify.hs
```diff
@@ -706,6 +743,7 @@ checkExternalResource Config{..} link
       , (405 ==)  -- method mismatch
       ]
 
+    interpretErrors :: Network.HTTP.Req.HttpException -> Either VerifyError ()
```
aeqz, line 746: I think that this same `HttpException` is already imported in this file, so there is no need to use its full name.

tests/Test/Xrefcheck/TooManyRequestsSpec.hs
```diff
@@ -150,18 +150,45 @@ test_tooManyRequests = testGroup "429 response tests"
...
-        _ <- verifyLink "http://127.0.0.1:5000/429grandfinale"
+        _ <- verifyLinkDefault "http://127.0.0.1:5000/429grandfinale"
         infoReverseAccumulator <- readIORef infoReverseAccumulatorRef
         reverse infoReverseAccumulator @?=
           [ ("HEAD", tooManyRequests429)
           , ("HEAD", serviceUnavailable503)
           , ("GET", ok200)
           ]
+    , testCase "Returns an error in case of timeout with 0 maximum retries allowed" $ do
+      let prog = Progress{ pTotal = 1
```
aeqz, line 161: I would add a space between `Progress` and `{` in these `prog` bindings.

aeqz (changes requested):
Looks good! I have suggested some small style changes. Also, at first sight I don't know why one Golden test is succeeding in one of our CIs and failing in the other.

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +454,103 @@ verifyReference
...
+                    Seconds s -> s
+                    -- Calculates the seconds left until @Retry-After@ date.
+                    -- Defaults to 0 if the date has already passed.
+                    Date date | utcTimeToTimeSecond date >= now -> utcTimeToTimeSecond date -:- now
+                    _ -> sec 0
+
+              let toRetry = rcTooManyRequestsRetries < ncMaxRetries cNetworking
+                              && rcTimeoutRetries < ncMaxTimeoutRetries cNetworking
+                  currentRetryAfter = fromMaybe (ncDefaultRetryAfter cNetworking) $
+                    fmap toSeconds mbCurrentRetryAfter
```
Martoon-00, line 500: I suppose `mbCurrentRetryAfter` can be not a `Maybe` now? Eventually, it is always initialized with `Just`.

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +454,103 @@ verifyReference
...
+                        then incProgress
+                        else id
+              modifyProgressRef now (Just currentRetryAfter) moveProgress
+              if toRetry
+              then do
+                threadDelay currentRetryAfter
+                retryVerification
+                  (counterModifier rc)
+                  resIO
+              else pure res
```
Martoon-00, line 514: I love this refactoring :+1:   Having these `case`s simplifies the downstream conditions.

I would now also try to split this canvas of code, maybe by extracting the part not related to progress update to a separate function - if this turns out to be viable.

And a note on the future: it would be worth putting this refactoring to a separate commit before adding the timeouts handling, this would simplify the review. But for sure, no need to extract this now, how it currently is - is fine too.

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +454,103 @@ verifyReference
...
-          if ok then decProgressFixableErrors
-          else if fixable then id
-          else fixableToUnfixable
-      where
-        ok = verifyOk res
-        fixable = any isFixable ves
+      | otherwise = id
 
-    extractRetryAfterInfo :: VerifyResult VerifyError -> Maybe RetryAfter
+    extractRetryAfterInfo :: VerifyError -> Maybe (Maybe RetryAfter, RetryCounter -> RetryCounter, RetryCounter -> Int, NetworkingConfig -> Int)
```
Martoon-00, line 544: Interesting :yum:

Martoon-00 (changes requested):
Thanks for your work, I like how you approach this :+1: 

Left a few comments.

src/Xrefcheck/Config.hs
YuriRomanowski: Maybe `ncMaxTooManyRequestsRetries`, like this.
I think that in presence of `ncMaxTimeoutRetries` the name `ncMaxRetries` is confusing because from the lexical point of view its scope includes `ncMaxTimeoutRetries`, but in the code it's used only for 429 errors. So I think we should rename it.

src/Xrefcheck/Verify.hs
YuriRomanowski: > But I'm relatively skeptical about switching to `Maybe`, because in my understanding, the commonly taken semantics of `Maybe` is: `Just` for ok and `Nothing` for failure. To carry errors, `Either err ()` or some custom type suits more thinking
I think that we can slightly improve the code if we rename `res` to `err`, so we'll show that we expect an error, and `Nothing` will just denote that there was no errors. As for me, there is not really big difference between `Maybe err` and `Either err ()` because here we don't expect anything except unit.

src/Xrefcheck/Verify.hs
YuriRomanowski: Not really: our `ExternalHttpTooManyRequests` error can be shipped with `Nothing`, when required header is missing or its contents can't be parsed properly. In this case the default value from the config is used.

tests/Test/Xrefcheck/TooManyRequestsSpec.hs
```diff
@@ -150,18 +150,45 @@ test_tooManyRequests = testGroup "429 response tests"
...
+                          , pCurrent = 1
+                          , pErrorsUnfixable = 1
+                          , pErrorsFixable = 0
+                          , pTaskTimestamp = Nothing
+                          }
+      checkLinkAndProgressWithServer
+        (setOneSecondAllowedTimeout . (cNetworkingL . ncMaxTimeoutRetriesL .~ 0))
```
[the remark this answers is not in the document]
YuriRomanowski: I think the latter option looks much better if we move each new `& <...>` part to next line:
```
\c -> c
  & setOneSecondAllowedTimeout
  & cNetworkingL . ncMaxTimeoutRetriesL .~ 0
```

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +454,103 @@ verifyReference
...
-      else return $ fmap (WithReferenceLoc fileWithReference ref) res
-
-    alterOverallProgress
+      case res of
+        Nothing -> do
+          let moveProgressOnSuccess =
+                incProgress
+                  . if errorsOccured rc
+                    then decProgressFixableErrors
+                    else id
```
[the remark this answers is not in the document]
YuriRomanowski: I think there is no big difference between these two options, in any case we put some functions delimiter at the line start, and who cares if it a dot or a comma? :yum: The only difference here is the indentation, so I can just move `if` clause to the left, if you like.

src/Xrefcheck/Verify.hs
Martoon-00: Ah right, looks good then :+1:

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
Martoon-00: > The only difference here is the indentation, so I can just move if clause to the left, if you like.

Yeah, let's do so then, to indicate that these two parts are about the same type of transformation.

src/Xrefcheck/Verify.hs
YuriRomanowski: Sure, I'll then replace `Maybe e` with `Either e ()`

tests/Test/Xrefcheck/TooManyRequestsSpec.hs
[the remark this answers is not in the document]
YuriRomanowski: > This sounds good, though I wonder: why in your opinion it would look much better? With do-block this could look similar (without touching the surrounding code):
> 
> ```haskell
> do
>   setOneSecondAllowedTimeout
>   cNetworkingL . ncMaxTimeoutRetriesL .= 0
> ```

Well, true, I like this `do` block more, too.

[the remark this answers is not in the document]
Martoon-00: Oh, I see. Feel free to apply whatever then.

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +444,104 @@ verifyReference
...
-        threadDelay currentRetryAfter
-        retryVerification (numberOfRetries + 1) resIO
-      else return $ fmap (WithReferenceLoc fileWithReference ref) res
-
-    alterOverallProgress
+      case res of
+        Right () -> do
+          let moveProgressOnSuccess = composeFuncList
+                [ incProgress
+                , whenFunc (errorsOccured rc) decProgressFixableErrors
```
Martoon-00, line 464: Neat :+1:

src/Xrefcheck/Verify.hs
Martoon-00: Thanks :+1:

src/Xrefcheck/Config.hs
```diff
@@ -62,6 +62,9 @@ data NetworkingConfig' f = NetworkingConfig
   , ncMaxRetries                :: Field f Int
     -- ^ How many attempts to retry an external link after getting
     -- a "429 Too Many Requests" response.
+  , ncMaxTimeoutRetries         :: Field f Int
+    -- ^ How many attempts to retry an external link after disconnect
+    -- due to the timeout
```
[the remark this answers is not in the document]
Martoon-00: Nice, see you also updated few other places. Updating the user-facing config was really important.

src/Xrefcheck/Config.hs
Martoon-00: That makes sense. Let me think about this for a while more...

Martoon-00:
One more point I wanted to bring up: did you see my comment at the bottom of #217? Does it make sense?
(I actually had to mention about this comment explicitly from the very beginning.)

In my understanding, just adding a general retry on timeout is not much beneficial because we already allow tuning the time after which timeout is declared. This way we put an unpleasant dilemma in front of the user: whether to add a retry that would be ineffective in 99% of the cases but at least double the awaiting time?

src/Xrefcheck/Config.hs
YuriRomanowski: After your comment about retrying on timeouts only for domains that earlier returned 429, I decided that it's better to use `maxRetries` field for all the errors, and `maxTimeoutRetries` only for timeouts. That is, we stop retrying if either total errors get greater than `maxRetries` or the total timeouts errors get greater than `maxTimeoutRetries`. Does it make sense?

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +448,104 @@ verifyReference
...
+      VerifyError ->
+      Maybe (Maybe RetryAfter, RetryCounter -> RetryCounter)
+    extractRetryInfo setOfReturned429 rc = \case
+      ExternalHttpTooManyRequests retryAfter _
+        | totalRetriesNotExceeded -> Just (retryAfter, incTotalCounter)
+      ExternalHttpTimeout (Just domain)
+        | totalRetriesNotExceeded && timeoutRetriesNotExceeded ->
+          if S.member domain setOfReturned429
+          then Just $ (Just (Seconds $ sec 0), incTimeoutCounter . incTotalCounter)
+          else Nothing
```
Martoon-00, line 539: One more simplifying refactoring, awesome :+1:

src/Xrefcheck/Config.hs
Martoon-00: Sounds very sensible to me :+1:

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +448,104 @@ verifyReference
...
+                    Seconds s -> s
+                    -- Calculates the seconds left until @Retry-After@ date.
+                    -- Defaults to 0 if the date has already passed.
+                    Date date | utcTimeToTimeSecond date >= now -> utcTimeToTimeSecond date -:- now
+                    _ -> sec 0
+
+              let currentRetryAfter = fromMaybe (ncDefaultRetryAfter cNetworking) $
+                    fmap toSeconds mbCurrentRetryAfter
+
+              let moveProgressOnFixable = whenFunc (rcTotalRetries == 0) incProgressFixableErrors
```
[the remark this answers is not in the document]
YuriRomanowski: Well, now here this case is only for retrying, so at the moment this link is not verified, we should wait for the retries result and only after that increment total completed counter.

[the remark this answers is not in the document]
YuriRomanowski: My understanding is that we increment progress only if it's already clear that verification is definitely successful or definitely unsuccessful. If we are going to retry, we can't say that in advance. So we will increment progress only when we get no errors or an unfixable error, but we'll perform this increment in other cases.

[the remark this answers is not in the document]
Martoon-00: Where do you see that the semantics of that counter is this?

src/Xrefcheck/Verify.hs
```diff
@@ -117,6 +115,9 @@ instance (Given ColorMode, Buildable a) => Buildable (WithReferenceLoc a) where
     #{wrlItem}
     |]
 
+newtype DomainName = DomainName {unDomainName :: Text}
+  deriving stock (Show, Eq, Ord)
```
Martoon-00, line 119: Good :+1: 

Please also write a brief haddock description, maybe with an example (it is the simplest way to be concrete here I think). So that there would be no ambiguities in interpreting the semantics of this type.

And, spacing [does not match](https://github.com/serokell/style/blob/master/haskell.md#general-identifiers) the style guide :sweat_smile:

src/Xrefcheck/Verify.hs
```diff
@@ -236,6 +239,21 @@ instance Given ColorMode => Buildable VerifyError where
       #{err}
       |]
 
+data RetryCounter = RetryCounter
+  { rcTotalRetries :: Int,
+    rcTimeoutRetries         :: Int
+  }
+  deriving stock (Show)
```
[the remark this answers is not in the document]
Martoon-00: Now better, but the [position of `,`](https://github.com/serokell/style/blob/master/haskell.md#list-declarations) is also important.

src/Xrefcheck/Verify.hs
```diff
@@ -694,8 +730,12 @@ checkExternalResource Config{..} link
       let maxTime = Time @Second $ unTime ncExternalRefCheckTimeout * timeoutFrac
 
       mres <- liftIO (timeout maxTime $ void reqLink) `catch`
-        (either throwError (\() -> return (Just ())) . interpretErrors)
-      maybe (throwError $ ExternalResourceSomeError "Response timeout") pure mres
+        (either throwError (\() -> return (Just ())) . interpretErrors uri)
+      maybe (throwError $ ExternalHttpTimeout $ DomainName <$> extractHost uri) pure mres
```
Martoon-00, line 734: `extractHost` is probably also worth making return `DomainName`

tests/Test/Xrefcheck/TooManyRequestsSpec.hs
```diff
@@ -149,35 +158,118 @@ test_tooManyRequests = testGroup "429 response tests"
...
+                                     & cNetworkingL . ncMaxRetriesL .~ 1
+                                     & setAllowedTimeout
+                                 , vlteLink = "http://127.0.0.1:5000/timeout"
+                                 , vlteProgress = mkProgressWithOneTask True
+                                 , vlteExpectation = VerifyResult []
+                                 }
+            , VerifyLinkTestEntry{ vlteConfigModifier = \c -> c
+                                     & cNetworkingL . ncMaxTimeoutRetriesL .~ 0
+                                     & setAllowedTimeout
+                                 , vlteLink = "http://localhost:5000/timeout"
```
Martoon-00, line 205: `127.0.0.1:5000` vs `localhost:5000`? That's smart :trollface: 

I guess this test counts, cannot imagine what can practically go wrong.

tests/Test/Xrefcheck/TooManyRequestsSpec.hs
```diff
@@ -149,35 +158,118 @@ test_tooManyRequests = testGroup "429 response tests"
...
-    checkLinkAndProgressWithServer mock link progress vrExpectation =
+    setAllowedTimeout = cNetworkingL . ncExternalRefCheckTimeoutL .~ (sec 0.25)
+
+    mkProgressWithOneTask shouldSucceed =
+                  Progress{ pTotal = 1
+                          , pCurrent = 1
+                          , pErrorsUnfixable = if shouldSucceed then 0 else 1
+                          , pErrorsFixable = 0
+                          , pTaskTimestamp = Nothing
+                          }
```
Martoon-00, line 224: Btw, I would try to avoid code indented like this.

I know this codebase has such code and I suppose you likely just followed it, but for the new code let's try to put `{` at the separate line. E.g. something like this works

```hs
    mkProgressWithOneTask shouldSucceed =
      Progress
        { pTotal = 1
        , pCurrent = 1
        , pErrorsUnfixable = if shouldSucceed then 0 else 1
        , pErrorsFixable = 0
        , pTaskTimestamp = Nothing
        }
```

or this

```hs
   mkProgressWithOneTask shouldSucceed = Progress
     { pTotal = 1
     , ...
     }
```

This does not apply to one-liners ofc, there indentation is not a concern.

No need to refactor everything in the project according to this rule, maybe only in a separate PR. For now let's follow this rule only for the new code.

Martoon-00 (commented):
Thanks for addressing my previous comments. Leaving a few more.

tests/Test/Xrefcheck/TooManyRequestsSpec.hs
```diff
@@ -149,35 +158,118 @@ test_tooManyRequests = testGroup "429 response tests"
...
+                                 , vlteExpectation = VerifyResult []
+                                 }
+            , VerifyLinkTestEntry{ vlteConfigModifier = \c -> c
+                                     & cNetworkingL . ncMaxTimeoutRetriesL .~ 0
+                                     & setAllowedTimeout
+                                 , vlteLink = "http://localhost:5000/timeout"
+                                 , vlteProgress = mkProgressWithOneTask False
+                                 , vlteExpectation = VerifyResult
+                                   [ ExternalHttpTimeout (Just $ DomainName "localhost")
+                                   ]
```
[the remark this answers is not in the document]
YuriRomanowski: Earlier you commented:
> At least one test I would like to also see: involve two sites, and test that 429 in one and a timeout in the other is treated respectively; currently this is not covered at all.

Doesn't this mean what I've added? This test checks that returning 429 from one site doesn't affect treating timeouts for another site.

tests/Test/Xrefcheck/TooManyRequestsSpec.hs
```diff
@@ -149,35 +158,118 @@ test_tooManyRequests = testGroup "429 response tests"
...
+        (mockTimeout (sec 0.4) mockResponses)
+        "http://127.0.0.1:5000/timeout" prog $
+          VerifyResult $
+            [ExternalHttpTimeout $ Just (DomainName "127.0.0.1") | not shouldSucceed]
+
+    checkMultipleLinksWithServer :: IO () -> IORef (S.Set DomainName) -> [VerifyLinkTestEntry] -> IO ()
+    checkMultipleLinksWithServer mock setRef entries =
+      E.bracket (forkIO mock) killThread $ \_ -> do
+        forM_ entries $ \VerifyLinkTestEntry {..} ->
+          checkLinkAndProgress
```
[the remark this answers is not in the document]
YuriRomanowski: Yes, I though about this, using one server for all tests can also reduce time for tests to go, but I'd like to do it in a separate PR.

tests/Test/Xrefcheck/TooManyRequestsSpec.hs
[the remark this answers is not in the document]
Martoon-00: > This test checks that returning 429 from one site doesn't affect treating timeouts for another site

This is a thing that is also worth testing, but we had this behaviour before our new feature too - previously timeout always resulted in `Response timeout` error, and here you test that we get `Response timeout` in case of 429. This test does not check that the feature has changed something.

I rather meant to test the opposite thing: that 429 will prevent the next `Response timeout` if config allows it.

tests/Test/Xrefcheck/TooManyRequestsSpec.hs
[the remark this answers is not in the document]
Martoon-00: That would be cool :+1:

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +448,104 @@ verifyReference
...
+                    Just (TaskTimestamp ttc start)
+                      | retryAfter +:+ now <= ttc +:+ start -> vrExternalAdvanced
+                    _ -> setTaskTimestamp retryAfter now vrExternalAdvanced
+             Nothing -> vrExternalAdvanced, .. }
+        else VerifyProgress{ vrLocal = moveProgress vrLocal, .. }
+      , ()
+      )
+
+    addDomainIf429 :: IORef (S.Set Text) -> VerifyError -> IO (S.Set Text)
+    addDomainIf429 setRef err = atomicModifyIORef' setRef $ \s ->
```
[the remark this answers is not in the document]
Martoon-00: I'm not sure how much sense this concern has, but maybe it will be interesting to investigate.

The potential issue with `atomicModifyIORef'` is that it may cause high contention. If we get 1000 simultaneous responses with 429 and invoke this function repeatedly, it may end up with different calls hindering each other, because AFAIU `atomicModifyIORef` is implemented as compare-and-set, and each thread will get value, perform `S.insert` computation, only to find out that during that time another thread tried to insert something and thus the result is invalidated and computation must be repeated. If this is true, we may get quadratic execution time.

At least, I have been witnessing such issues with STM once.

Could you set up a simple experiment with some minimalistic example using `atomicModifyIORef'` and try?

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
Martoon-00: Thanks, looks correct to me now :+1:

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +455,107 @@ verifyReference
...
+                    Date date | utcTimeToTimeSecond date >= now -> utcTimeToTimeSecond date -:- now
+                    _ -> sec 0
+
+              let currentRetryAfter = fromMaybe (ncDefaultRetryAfter cNetworking) $
+                    fmap toSeconds mbCurrentRetryAfter
+
+              let moveProgressOnFixable = composeFuncList
+                    [ whenFunc (rcTotalRetries == 0) incProgressFixableErrors
+                    , whenFunc (rcTotalRetries == 0) incProgress
+                    ]
```
Martoon-00, line 508: And a couple of more things: I suppose, in all three branches, this can be composed in a more compact way. E.g. in the first case it is just `if .. then .. else`.

Also, this code will be more readable if we used consistently either `errorsOccured` or `rcTotalRetries == 0` in all the `case` branches.

Martoon-00: Thanks, that's exactly what I meant.

Martoon-00:
Okay, very good :+1: 

Please prettify the commits history now. Some commits are worth leaving separate, as the one where you eliminate redundant `$`s (but squash 002a08d9f022113d4d9508fe7c036073b2ada51d & 0e4a9de764287f4cd37a5be911ae75998430a79a). Other commits can probably be squashed into one, most of your commits just fix something in the very first commit.

Then please rebase on the current master.

And let's call @aeqz for another review after that.

YuriRomanowski:
It seems that with new settings some sites started to fail randomly, mostly I see https://github.com/serokell/xrefcheck/pull/88 failing often, but others too, should we add them to the config ignore list?

aeqz:
> It seems that with new settings some sites started to fail randomly, mostly I see https://github.com/serokell/xrefcheck/pull/88 failing often, but others too, should we add them to the config ignore list?

If I understand the issue correctly, this should not be happening: we are giving the requests more chance to succeed than before, so failure reports should not increase with this new feature. Is there any explanation for this? Otherwise, it would be worth to investigate.

YuriRomanowski:
> If I understand the issue correctly, this should not be happening: we are giving the requests more chance to succeed than before, so failure reports should not increase with this new feature. Is there any explanation for this? Otherwise, it would be worth to investigate.

Before this PR the timeout error wasn't fixable, so we called these sites with 0.3*<timeout config> using HEAD, then, in case of timeout, with 0.7*<timeout config> using GET. Currently timeout became fixable, and we are thrown to `retryVerification` after the first call. I think this may be the reason.

aeqz:
Hmm, maybe that was right when the only fixable error was 'Too many requests', but now that behaviour should be preserved only for that same error, like this:

```haskell
checkHttp :: URI -> ExceptT VerifyError IO ()
    checkHttp uri = makeHttpRequest uri HEAD 0.3 `catchError` \case
      ExternalHttpTooManyRequests _ _ -> throwError e
      _ -> makeHttpRequest uri GET 0.7
```

In a high level, by reading the issue, I expected this feature to decrease the number of reported errors, but not to increase them.

YuriRomanowski:
Yep, after trying with HEAD, I think we can just throw errors only on `TooManyRequests`, and try GET on timeout. Only after timeout on GET request a timeout error will be reported.

Martoon-00:
I tried locally, and get the same behaviour as you described: random timeouts after 3 seconds. So in the 3 seconds limit we just didn't manage to connect to the site, we got a timeout and instead of going to `GET` attempt as before we went to the retry logic and immediately failed there.

Not handling timeouts coming from `HEAD` is an option, but there is a better fix: currently `S.member domain setOfReturned429` seems to be just called in the wrong place, I think we should move it to `checkHttp`.

Martoon-00:
> By the way, I don't clearly understand the sense of 0.3/0.7 behaviour, do you have any ideas? If I were the user, I would expect the tool to call sites with exactly the same timeout as in the config.

I believe when configuring the timeout, the user more often thinks about "how much time will it take overall?" rather than "how much time do I give my PC to connect?". So such a distribution.

But maybe we could discuss an opportunity to remove the two attempts at all (later in Slack, currently let's apply the quickest fix), taking the recent information into account I suppose the current approach may be something redundant.

YuriRomanowski:
> Not handling timeouts coming from HEAD is an option, but there is a better fix: currently S.member domain setOfReturned429 seems to be just called in the wrong place, I think we should move it to checkHttp.

Could you please clarify this statement? Do you mean that the further behaviour after getting a timeout from HEAD request depends on presence of the domain in the `Set`? But I don't see what are the alternatives, it seems that in case of timeout the best is always to call GET. Or we have another option: if the domain is already in the set (so it returned 429), we can just skip HEAD request. If a domain returned 429, it both shouldn't like one extra request per try (I mean HEAD) and is unlikely to respond fast (as we expect in HEAD request). So what I think can be reasonable to implement:
1. In case of timeout in HEAD, we then always call GET
2. If the domain returned 429, we skip HEAD request for it
3. Other logic related to retries stays as it's already implemented

Does it make sense?

Martoon-00:
Skipping `HEAD` seems to be not justified in this case, because the primary motivation for it was not to simply gain a speed-up, but to avoid loading more data than necessary (AFAIU `GET` can return much more information) and to not use much of the user's traffic. This is debatable though.

> it seems that in case of timeout the best is always to call GET

Probably let's elaborate here. I see the following reasons why timeout can happen:
1. The site does not work.
2. We just didn't manage to get response in that short amount of time (`HEAD` request is especially subject to this) - this is the issue we witness now.
3. Timeout actually means 429 - this is what we resolved in this ticket.

In case we get a timeout and previously got 429 for the same domain, 1 can be treated as impossible, and both 2 and 3 can take place.

I suspect that your proposal aimed at improving the case 2 (there is no sense in retrying `HEAD` with that short duration), and my proposal aimed at improving the case 3 (in this case there is no sense in treating timeout differently from 429).

Aand, now I'd agree that how my proposal handles the 2nd case is much worse than how your proposal handles the 3rd case.

So I'd really agree on going with the code suggested [by @aeqz above](https://github.com/serokell/xrefcheck/pull/234#issuecomment-1353158327) unless we continue discussing other options. 

And I guess we really should document somewhere what a timeout can mean in our interpretation. Otherwise, understanding some decisions taken in code will soon become nearly impossible.

YuriRomanowski:
> I believe when configuring the timeout, the user more often thinks about "how much time will it take overall?" rather than "how much time do I give my PC to connect?". So such a distribution.

This makes sense, but for this case our current config naming is a bit confusing: while the time actually means "longest overall time", it's named `...Timeout` :smile_cat:

YuriRomanowski:
> Skipping HEAD seems to be not justified in this case, because the primary motivation for it was not to simply gain a speed-up, but to avoid loading more data than necessary (AFAIU GET can return much more information) and to not use much of the user's traffic.

Well, I suggested to call `GET` not really because it's `GET`, but because we let greater timeout for it. But we can we just call `HEAD` with greater timeout allowed. In fact, we could analyze the result of the `HEAD` result in more details - if `HEAD` is allowed, but there is not enough time for it to respond, it seems that there's no reason to change the method, just to change the timeout.

Martoon-00:
> This makes sense, but for this case our current config naming is a bit confusing: while the time actually means "longest overall time", it's named `...Timeout` :smile_cat:

Weeell, I think this appears confusing when looking at it from the inner implementation point of view. But for the end user, there are no different `HEAD`/`GET` calls, and for him "timeout" is a fair synonym for "longest overall time".

Martoon-00:
Please note that several discussions above are unresolved since at least partially the points there are not addressed.

Also, we should do this:
> Also, I guess we will have to document somewhere what a timeout can mean in our interpretation

src/Xrefcheck/Verify.hs
YuriRomanowski: > I would now also try to split this canvas of code, maybe by extracting the part not related to progress update to a separate function - if this turns out to be viable.

It seems that this function is already OK:+1: now. If the decision is up to me, let's leave it as is.

YuriRomanowski:
Hm, for some reason one of the unit tests fails on the CI, although everything is OK for me locally. Let's try to increase the allowed timeout in the tests.

Martoon-00:
Hmhm, so increasing timeout didn't help.

Given that this is Windows-only error, maybe `localhost` is not necessarily equal to `127.0.0.1` there :thinking:  And the timeout then means that the target is unaccessible.

src/Xrefcheck/Verify.hs
Martoon-00: Okay, sounds nice :+1:

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
Martoon-00: #251 has been created for this, we will re-consider things there.

src/Xrefcheck/Verify.hs
```diff
@@ -445,94 +469,113 @@ verifyReference
...
+      Maybe (Maybe RetryAfter, RetryCounter -> RetryCounter)
+    decideWhetherToRetry setOfReturned429 rc = \case
+      ExternalHttpTooManyRequests retryAfter _
+        | totalRetriesNotExceeded -> Just (retryAfter, incTotalCounter)
+      ExternalHttpTimeout (Just domain)
+        | totalRetriesNotExceeded && timeoutRetriesNotExceeded ->
+          -- If a given domain ever returned 429 error, we assume that getting timeout from
+          -- the domain can be considered as a 429-like error, and hence we retry.
+          -- If there was no 429 responses from this domain, then getting timeout from
+          -- it probably means that this site is not working at all.
```
Martoon-00, line 566: Very good :+1: 

And I would also mention that always remains a possibility that we just didn't get the response in time, but this issue cannot be mitigated at this level.

YuriRomanowski:
> Given that this is Windows-only error, maybe localhost is not necessarily equal to 127.0.0.1 there thinking And the timeout then means that the target is unaccessible.

Hm, other tests also use `localhost` and they are not failing... I really have no idea what's wrong

tests/Test/Xrefcheck/TimeoutSpec.hs
```diff
@@ -0,0 +1,146 @@
+{- SPDX-FileCopyrightText: 2021 Serokell <https://serokell.io>
```
Martoon-00, line 1: :warning: The year in this license header is outdated, time to update!





</p>

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit f57944feb29b3d04c685659488b22b95ea1a74f7 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 449f065a4366641d0f4d45f5c22c4edc72776066 lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit eba000d6dddbb33e96a441e395a87bbfd05486da lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">In f57944feb29b3d04c685659488b22b95ea1a74f7 message lacks issue id: <code>Perform GET request after HEAD request failed on timeout</code>.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">In 449f065a4366641d0f4d45f5c22c4edc72776066 message lacks issue id: <code>After rebase: reorder tests, move timeout tests to a separate module</code>.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">In eba000d6dddbb33e96a441e395a87bbfd05486da message lacks issue id: <code>Remove redundant $</code>.<br />
See also [Note].</td>
    </tr>
  </tbody>
</table>

</p>

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

src/Xrefcheck/Verify.hs
Martoon-00: Great, thanks :+1:

Martoon-00:
> Hm, other tests also use localhost and they are not failing... I really have no idea what's wrong

Well, that still fits my theory. You have only two tests that use `localhost`, and in the succeeding one you expect timeout - and it happens (but not due to the expected reason).

Mm, I would still at least try playing with `127.0.0.1` ↔ `localhost` replacements and see how that CI reacts (unless you already did, in this case please respond with your findings).

YuriRomanowski:
Thank you, @Martoon-00, CI is passing. However, it turns out that other test that uses both `localhost` and `127.0.0.1` works quite by chance :smile:

Martoon-00:
> Thank you, @Martoon-00, CI is passing. 

Oh great, glad to know that we figured it out.

> However, it turns out that other test that uses both localhost and 127.0.0.1 works quite by chance smile

Heeh, looks so :see_no_evil: 

AFAIU using `127.0.0.1` and replacing `timeout` with `timeoutother` is now the recommended option to test two different links?

Martoon-00:
When squashing commits, don't forget that cf11b96ce9f0726f94d1109fe1bae995ede72030 is not necessary (AFAIU).

aeqz (approved):

src/Xrefcheck/Config/Default.hs
```diff
@@ -56,6 +56,10 @@ networking:
   # a "429 Too Many Requests" response.
   maxRetries: 3
 
+  # How many attempts to retry an external link after disconnect
+  # due to the timeout.
+  maxTimeoutRetries: 1
```
Martoon-00, line 61: I would elaborate in description of this more, currently it sounds like something related solely to timeouts and not to 429s.

tests/Test/Xrefcheck/UtilRequests.hs
```diff
@@ -7,46 +7,69 @@ module Test.Xrefcheck.UtilRequests
...
+  IO () ->
+  Text ->
+  Progress Int ->
+  VerifyResult VerifyError ->
+  IO ()
+checkLinkAndProgressWithServer configModifier setRef mock link progress vrExpectation =
+  E.bracket (forkIO mock) killThread $ \_ -> do
+    checkLinkAndProgress configModifier setRef link progress vrExpectation
+
+checkLinkAndProgress configModifier setRef link progress vrExpectation = do
```
Martoon-00, line 56: Do you lack the function signature here?

Martoon-00 (commented):
Thanks, very nice work :+1: 
Looking forward to seeing it merged.

I have three comments remaining, including [this one](https://github.com/serokell/xrefcheck/pull/234#issuecomment-1361212479) that you previously noted.

After that, I'll approve.

YuriRomanowski:
> AFAIU using `127.0.0.1` and replacing `timeout` with `timeoutother` is now the recommended option to test two different links?

If you mean different domains, we need both `127.0.0.1` and localhost, different paths don't result in different domains if I understand it properly. So that test we can't change

src/Xrefcheck/Config/Default.hs
Martoon-00: And I would mention the motivation for it, which exactly case this serves for

Martoon-00:
> If you mean different domains, we need both 127.0.0.1 and localhost, different paths don't result in different domains if I understand it properly. So that test we can't change

Oh right, I was saying nonsense.

Okay, let's hope that under Linux this test still tests the original thing and that is enough.

tests/Test/Xrefcheck/TimeoutSpec.hs
```diff
@@ -0,0 +1,143 @@
...
+-- Here all the delays are doubled because we call sites
+-- with HEAD first and then GET methods.
+test_tooManyRequests :: TestTree
+test_tooManyRequests = testGroup "Timeout tests"
+  [ testCase "Returns an error on timeout if there were no 429 responses" $
+      timeoutTestCase [Delay, Delay] False
+        (cNetworkingL . ncMaxTimeoutRetriesL .~ 2)
+  , testCase "Returns an error if there were 429 but no timeouts allowed" $
+      timeoutTestCase [Respond429, Delay, Delay] False
+        (cNetworkingL . ncMaxTimeoutRetriesL .~ 0)
```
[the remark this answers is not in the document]
YuriRomanowski: I have added it on the top, is it OK?

src/Xrefcheck/Verify.hs
```diff
@@ -426,94 +455,109 @@ verifyReference
...
+          in case mbRetryAfter of
+             Just retryAfter -> case pTaskTimestamp vrExternal of
+                    Just (TaskTimestamp ttc start)
+                      | retryAfter +:+ now <= ttc +:+ start -> vrExternalAdvanced
+                    _ -> setTaskTimestamp retryAfter now vrExternalAdvanced
+             Nothing -> vrExternalAdvanced, .. }
+        else VerifyProgress{ vrLocal = moveProgress vrLocal, .. }
+      , ()
+      )
+
```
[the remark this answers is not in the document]
YuriRomanowski: In `decideWhetherToRetry` we just want to know whether the current domain is present in the set, I think it's not important _when_ it was added to the set. But in this case we either add a new domain to the set (for 429), or lookup the domain in the set (for timeout), not together. Yeah, it seems that this code is still a bit coupled and we could handle cases of timeout and 429 separately, and it would become obvious that we use `Set` only once in each case, but I think it's not a big problem. So, to sum up: if 429, we add domain to the set, if timeout, lookup for this domain in the set.

tests/Test/Xrefcheck/TimeoutSpec.hs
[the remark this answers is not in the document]
Martoon-00: Ah right, apparently I forgot to resolve this comment.

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
Martoon-00: Yeah, the idea about exploiting `Set.size` was interesting, but sadly it does not come in handy in our case.

src/Xrefcheck/Config/Default.hs
Martoon-00: Good, but this still mentions only what we do and not why, and the user may miss what's the point behind such complexity.

Instead of the last 3 lines I would just add, in order not to repeat mostly the same information:

```
  # Sometimes, after we exceeded rate limits, the site may stop
  # responding to our requests, and this option helps to workaround
  # that.
```

Martoon-00:
Looks like a rebase awaits you :fearful: 

I'd like to take the last look after it is done.

src/Xrefcheck/Config/Default.hs
YuriRomanowski: I tried to rewrite two config fields' descriptions from scratch, unfortunately it's best that I'm able to produce. If that version is not acceptable, could you please write these descriptions yourself?

Martoon-00: > it's best that I'm able to produce. If that version is not acceptable, let's just put your version there

By the way, from the standpoint of an efficient review-edit cycle, very good if you put it clearly; oftentimes I observed how such reviewer's nitpicking turned into infinite dialogs. I saw that you can write good text covering all the necessary points, and here it's rather me who looks at things from a different angle and want accents to be placed differently.

Now coming to your edit itself:

> Total number of allowed 429 errors and timeouts
> when querying an external link.
> On other errors xrefcheck fails immediately, without retrying.

Since the reason for timeouts handling is not obvious at first, I would leave the first variant + extend it:

```
  # How many attempts to retry an external link after getting
  # Timeouts may also be accounted here, see the description
  # of `maxTimeoutRetries` field.
  
  # If a site once responded with 429 error code, subsequent
  # request timeouts will also be treated as hitting the site's 
  # rate limiter and result in retry attempts, unless the
  # maximum retries number has been reached.
  #
  # On other errors xrefcheck fails immediately, without retrying.
```

> Querying a given domain that ever returned 429 before,
> this defines how many timeouts are allowed.
> For such domains, xrefcheck considers timeouts as possible
> 429 errors, and for other domains timeout immediately
> results in error, assuming that this site is not working at all.

I would rephrase it as follows:

```
Querying a given domain that ever returned 429 before,
this defines how many timeouts are allowed during retries.

For such domains, timeouts likely mean hitting the rate limiter, 
and so xrefcheck considers timeouts in the same way as 429 errors.

For other domains, a timeout results in a respective error, no retry
attempts will be performed. Use `externalRefCheckTimeout` option
to increase the time after which timeout is declared.

This option is similar to `maxRetries`, the difference is that
this `maxTimeoutRetries` option limits only the number of retries
caused by timeouts, and `maxRetries` limits the number of retries
caused both by 429s and timeouts.
```

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Commit 10b2eccb87235ef76cdae72416d2aa450ed4fd6e lacks description :unamused:<br />
Commits marked as <code>[Chore]</code> are exempt from this check.<br />
See also [Note].</td>
    </tr>
  </tbody>
</table>

</p>

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

src/Xrefcheck/Config/Default.hs
YuriRomanowski: Thank you very much @Martoon-00

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

src/Xrefcheck/Config/Default.hs
Martoon-00: Thanks for applying the update :+1:

Martoon-00: @aeqz If you have thoughts how to improve this further, please comment here too.

Martoon-00:
@aeqz Could I also ask you to thoroughly check the test part again? AFAIU it has been rebased on top of your work, and likely there were severe merge conflicts there if I'm not mistaken, so double-checking won't harm.

Martoon-00 (approved):
Looks good to me, thank you for your work :+1: :+1: 

Please wait for @aeqz's comment on rebase (or say that I'm wrong there and rebase was quite trivial), and let's merge after that.

tests/Test/Xrefcheck/UtilRequests.hs
```diff
@@ -60,18 +91,48 @@ checkLinkAndProgressWithServer mock link progress vrExpectation =
...
 
-verifyLink :: Text -> IO (VerifyResult VerifyError, Progress Int)
-verifyLink link = do
+checkLinkAndProgressWithServerDefault ::
+  IORef (Set DomainName) ->
+  IO () ->
+  Text ->
+  Progress Int ->
+  VerifyResult VerifyError ->
+  IO ()
```
aeqz, line 100: I think that multi-line function type signatures should be styled like this:

```haskell
heckLinkAndProgressWithServerDefault
  :: IORef (Set DomainName)
  -> IO ()
  -> Text 
  -> Progress Int 
  -> VerifyResult VerifyError
  -> IO ()
```

LICENSES/LicenseRef-MIT-OA.txt
```diff
@@ -1,6 +1,6 @@
 MIT License
-Copyright (c) 2021-2022 Oxhead Alpha
-Copyright (c) 2019-2021 Tocqueville Group
+Copyright (c) 2022-2022 Oxhead Alpha
+Copyright (c) 2019-2022 Tocqueville Group
```
aeqz, line 3: Is this change right? I'm also not sure, but it seemed fine as it was.

src/Xrefcheck/Config/Default.hs
aeqz: I think that it is clear now 👍

aeqz (approved):
The rebase may have not been trivial, but every conflict seems to be properly resolved. I'm leaving a couple of comments.

LICENSES/LicenseRef-MIT-OA.txt
Martoon-00: Yeah, `Copyright (c) 2019-2021 Tocqueville Group` means that this code belonged to Tocqueville Group during 2019-2021. But they don't own the edits after that (because they don't pay for future code maintenance), so that part of license header remains unmodified.

Hmm, if Danger bot catches you here, maybe it would be cool to update the rules respectively in some separate MR (and for now let the warning be).

YuriRomanowski: Yeah, let's fix these years in the next PR (and in the next year :smile: )

PR #236, opened 2022-12-11
Yuri romanowski/#220 avoid resource leaking in scanners

### LINKED ISSUE #220: Avoid resource leaking in scanners

## Problems

This is an issue that I kept in mind for a long time, and since xrefcheck is becoming a serious tool, I think we should have this issue handled.

It is a known minor issue of `readFile` that until the full produced lazy bytestring/text is processed, the file descriptor remains open.

An open file descriptor is a resource, and OSes tend to have a limit on the number of file descriptors that can be opened, and maybe a separate limit per process too. Meaning that we should be cautious to not exceed these limits.

In the markdown scanner, we use `readFile` and seemingly do not enforce the computation (we apply parsing, but its result is not forced), meaning that the file descriptor won't be closed before the result of scanning being requested, and in case of large repositories this may be a severe issue.

---

Another issue similar to that: our markdown parser works on a strict `Text`, meaning that during parsing the entire file's content will be kept in memory, and memory can be treated as a resource too.

How much severe this issue is? I'm not sure, documentation usually does not take much space, probably even if it is auto-generated. Documentation size is fundamentally limited by how much a human being can read, so we do not expect dozen of gigabytes here, let's further throw this concern away. 

## Proposed solution

Conceptually, we want two things:
1. Minimize the time between `readFile` start and full parsing, as this is the time when we hold the resource (first the opened descriptor, then file content in memory).
2. Make parsing of various files as parallel as possible, in theory.

So from a `ScanAction` object we probably expect that it reads the file as quickly as possible, maybe putting its full content in the memory, but is free to return a not fully evaluated result (`(FileInfo, [ScanError])` pair). Forcing it is the caller's responsibility. 

## What should be done in code

First, let's document the expectations from `ScanAction`, add them to haddock of this type.

Next, we should make sure that after `readFile` call finishes, we file is fully read.

Since we do not parallelize file reads (in case of I/O this AFAIU does not make much sense), this will automatically resolve our issue with too many opened file descriptors.

# Acceptance criteria

* Documentation of `ScanAction` is updated to include the mentioned comments about laziness of each inner action (file read and parsing).
* The current implementation of `ScanAction` is updated respectively.

YuriRomanowski:
## Description

[#220] Avoid resource leaking in scanners
    
Problem: Currently we use lazy `readFile`, and xrefcheck opens all
the files immediately for reading, while file descriptors are
resource that can be finite.
    
Solution: Replaced lazy `readFile` with strict version.

To be done: Understand issues with parallelism

## Related issue(s)

Fixes #220 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

#### ✓ Release Checklist

### DIFF
--- a/src/Xrefcheck/Scan.hs
+++ b/src/Xrefcheck/Scan.hs
@@ -2,7 +2,7 @@
  -
  - SPDX-License-Identifier: MPL-2.0
  -}
-
+{-# LANGUAGE DeriveAnyClass #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 -- | Generalised repo scanner and analyser.
@@ -93,7 +93,8 @@ data ScanError = ScanError
   { sePosition    :: Position
   , seFile        :: FilePath
   , seDescription :: ScanErrorDescription
-  } deriving stock (Show, Eq)
+  } deriving stock (Show, Eq, Generic)
+    deriving anyclass NFData
 
 instance Given ColorMode => Buildable ScanError where
   build ScanError{..} = [int||
@@ -118,7 +119,8 @@ data ScanErrorDescription
   | FileErr
   | ParagraphErr Text
   | UnrecognisedErr Text
-  deriving stock (Show, Eq)
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (NFData)
 
 instance Buildable ScanErrorDescription where
   build = \case
--- a/package.yaml
+++ b/package.yaml
@@ -111,6 +111,7 @@ library:
     - reflection
     - nyan-interpolation
     - safe-exceptions
+    - parallel
 
 executables:
   xrefcheck:
--- a/src/Xrefcheck/Scanners/Markdown.hs
+++ b/src/Xrefcheck/Scanners/Markdown.hs
@@ -12,6 +12,7 @@ module Xrefcheck.Scanners.Markdown
 
   , defGithubMdConfig
   , markdownScanner
+  , markdownParallelScanner
   , markdownSupport
   , parseFileInfo
   , makeError
@@ -35,6 +36,7 @@ import Text.Interpolation.Nyan
 import Xrefcheck.Core
 import Xrefcheck.Scan
 import Xrefcheck.Util
+import Control.Parallel.Strategies
 
 data MarkdownConfig = MarkdownConfig
   { mcFlavor :: Flavor
@@ -415,5 +417,10 @@ parseFileInfo config fp input
 markdownScanner :: MarkdownConfig -> ScanAction
 markdownScanner config path = parseFileInfo config path . decodeUtf8 <$> BS.readFile path
 
+markdownParallelScanner :: MarkdownConfig -> ScanAction
+markdownParallelScanner config path = do
+  resThunk <- parseFileInfo config path . decodeUtf8 <$> BS.readFile path
+  resThunk `usingIO` rparWith rdeepseq
+
 markdownSupport :: MarkdownConfig -> ([Extension], ScanAction)
-markdownSupport config = ([".md"], markdownScanner config)
+markdownSupport config = ([".md"], markdownParallelScanner config)
--- a/src/Xrefcheck/Scanners/Markdown.hs
+++ b/src/Xrefcheck/Scanners/Markdown.hs
@@ -24,11 +24,10 @@ import CMarkGFM
 import Control.Lens (_Just, makeLenses, makeLensesFor, (.=))
 import Control.Monad.Trans.Writer.CPS (Writer, runWriter, tell)
 import Data.Aeson (FromJSON (..), genericParseJSON)
-import Data.ByteString.Lazy qualified as BSL
+import Data.ByteString qualified as BS
 import Data.DList qualified as DList
 import Data.Default (def)
 import Data.Text qualified as T
-import Data.Text.Lazy qualified as LT
 import Fmt (Buildable (..), nameF)
 import Text.HTML.TagSoup
 import Text.Interpolation.Nyan
@@ -406,16 +405,15 @@ textToMode ("ignore" : [x])
   | otherwise        = InvalidMode x
 textToMode _         = NotAnAnnotation
 
-parseFileInfo :: MarkdownConfig -> FilePath -> LT.Text -> (FileInfo, [ScanError])
+parseFileInfo :: MarkdownConfig -> FilePath -> T.Text -> (FileInfo, [ScanError])
 parseFileInfo config fp input
   = runWriter
   $ flip runReaderT config
   $ nodeExtractInfo fp
-  $ commonmarkToNode [optFootnotes] [extAutolink]
-  $ toStrict input
+  $ commonmarkToNode [optFootnotes] [extAutolink] input
 
 markdownScanner :: MarkdownConfig -> ScanAction
-markdownScanner config path = parseFileInfo config path . decodeUtf8 <$> BSL.readFile path
+markdownScanner config path = parseFileInfo config path . decodeUtf8 <$> BS.readFile path
 
 markdownSupport :: MarkdownConfig -> ([Extension], ScanAction)
 markdownSupport config = ([".md"], markdownScanner config)


YuriRomanowski:
When I tried to parallelize file parsing, I got some strange issues. I tried to run a program with ~100 md files repository with 8 core available. This is needed time info:
```
start scan repo; time: 0.00s
end scan repo; time: 0.47s
start verify repo; time: 0.47s
end verify repo; time: 1.88s
```
And here is the threadscope profile:
изображение

We see that CPU activity is not very high, although there is work to do. Something prevents threads to work normally... This is what can be found in the eventlog for that time: 
изображение

There are a lot of foreign calls, although it seems in my code there is no anything that can result in such behaviour. I tried to run our program under the `strace` utility, and it showed this: 
изображение
(Sorry for the russian caption, it means "Resource temporary unavailable")

Googling this returns one Stackoverflow entry, but I didn't understand much there :smile: 
[click](https://stackoverflow.com/questions/60670801/application-freezes-on-futex-call)
Any ideas why this can occur? Maybe my code is written in a wrong way?

Martoon-00:
Woow, elaborate investigation :+1: 

However, what do you think exactly is wrong?

I'll tell my thoughts protecting the idea that everything seems fine, and you show where I'm wrong.

In the threadscope report, to me it looks like what I would expect - low CPU consumption during the scanning stage because it is computation interleaved with I/O (and I/O is not shown on the chart).

It may be reasonable that I/O takes more time than parsing, since the former is reading from the disc, and the latter is reading from memory + some logic written in C that `cmark-gfm` uses internally and which thus can be fast. (Btw, you are on SSD disk?)

On what you see in the eventlog - by default, foreign calls cannot be interrupted, at that moment runtime system gives up on trying to manage anything until the foreign call is finished. Maybe that's why we see those messages, but I think they are fine, they may just mean that the control is passed to the C logic, but everything is fine with threads (I'm not sure here though, may be worth googling particularly the meaning of that eventlog message).

On `strace` output, as far as I understand `EAGAIN` in our case likely means "pipe is yet empty, `read` syscall cannot return anything, please try again later", and this matches the explanation above that CPU is ready to read input at a more intense rate than the device can provide.

If my understanding is correct, most of the witnessed things will be still there even if we try parsing with no parallelization as before.

---

If you had some intuition that told you "something is wrong here" or "things are too slow" that pushed you onto this investigation, could you share it too?

YuriRomanowski:
> In the threadscope report, to me it looks like what I would expect - low CPU consumption during the scanning stage because it is computation interleaved with I/O (and I/O is not shown on the chart).

Well, I also tried to write code such that all of the files were read at once, and then only pure computations were performed in parallel, but results were the same. I can later show same graphs for this case.

> If you had some intuition that told you "something is wrong here" or "things are too slow" that pushed you onto this investigation, could you share it too?

In my mind, when we have a big list of pure tasks, they can be parallelized very effectively, using ~100% load for all available CPUs. But here, although there are some CPUs, total activity is not higher (in average) than just one CPU 100% load (I look at the dashed lines on the activity graph), and this is not really satisfying. When I tried to read all the files first and then parse them in parallel, reading was performed with one thread which took 100% of CPU load, but total activity was lesser for further parsing stage, even for 8 CPUs. Also, I don't like a lot of errors in a row for a system calls log, it looks similar to active waiting, the thing that good programs usually (in my mind) don't want to use. But GHC runtime is written with highly-skilled guys, so that can indicate that we are using their creature in a wrong way. So I believe that there's a problem we could solve, although I don't quite understand what it actually is.

Martoon-00:
> Also, I don't like a lot of errors in a row for a system calls log, it looks similar to active waiting

Yeah, that's true. Also not sure why it is done this way.

> Well, I also tried to write code such that all of the files were read at once, and then only pure computations were performed in parallel

Oh that's a very good setup for testing, then we have to investigate what happens there. I'll try to think about it, if you have more info to share - it is welcome.

aeqz:
I'm not sure if this will reveal anything, but could you try the following in order to check if the parallel performance looks more like what you expected?

Derive `NFData` for `ScanResult` and then, in the `scanRepo` function, replace

```haskell
return . ScanResult errs $ RepoInfo
  { riFiles = M.fromList $ processedFiles <> notProcessedFiles
  , riDirectories = M.fromList
      $ map (, TrackedDirectory) trackedDirs
      <> map (, UntrackedDirectory) untrackedDirs
  , riRoot
  }
```

by 

```haskell
let result = ScanResult errs $ RepoInfo
      { riFiles = M.fromList $ processedFiles <> notProcessedFiles
      , riDirectories = M.fromList
                        $ map (, TrackedDirectory) trackedDirs
                        <> map (, UntrackedDirectory) untrackedDirs
      }

evaluateNF result
```

src/Xrefcheck/Scan.hs
```diff
@@ -118,7 +119,8 @@ data ScanErrorDescription
   | FileErr
   | ParagraphErr Text
   | UnrecognisedErr Text
-  deriving stock (Show, Eq)
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (NFData)
```
aeqz, line 123: You can derive `NFData` with standalone deriving if you prefer to follow the same style that in `Xrefcheck.Core` module, which contains a block like the following:

```haskell
instance NFData Position
instance NFData Reference
instance NFData AnchorType
instance NFData Anchor
instance NFData FileInfo
instance NFData RepoInfo

```

YuriRomanowski:
I tried to force evaluation of CMarkGFM part at first, and it looks like this. Here the highlighted area corresponds to parsing using CMarkGFM, and then sparks are created to do the work in pure Haskell. And in the CMarkGFM area there are a lot of foreign calls. It seems that these are just calls to C library :) It's a good question whether we can parallelize foreign calls, in some circumstances we should be able to. Also we can stop here and just run parallel computations interleaved with foreign calls. Anyway, we do get some speedup, but not a big one.
изображение

Martoon-00:
Interesting. I guess foreign calls _should be_ parallelizable...

So from the recent picture we can tell, that if you don't use sparks for `commonmarkToNode` but use them for all the following parsing, then we see the expectedly high usage of CPU cores for that "following parsing".

I'd really like to see the threadscope report from this case:

> Well, I also tried to write code such that all of the files were read at once, and then only pure computations were performed in parallel, but results were the same. I can later show same graphs for this case.

Martoon-00:
Although 0.07 seconds that this unparallelized call to cmark-gfm currently takes is really not that much, I agree we could stop optimizing here if we get stuck at it.

YuriRomanowski:
> Although 0.07 seconds that this unparallelized call to cmark-gfm currently takes is really not that much, I agree we could stop optimizing here if we get stuck at it.

I'm sorry for confusing you, in fact all the time from 0s to 1.55s is for files reading and cmark parsing

YuriRomanowski:
> Although 0.07 seconds that this unparallelized call to cmark-gfm currently takes is really not that much, I agree we could stop optimizing here if we get stuck at it.

I'm sorry for confusing you, but the highlighted period is not the full period of CMark parsing, just an example.

> Well, I also tried to write code such that all of the files were read at once, and then only pure computations were performed in parallel, but results were the same. I can later show same graphs for this case.

Yep, later I'll show more accurate examples for different cases.

Martoon-00:
> but the highlighted period is not the full period of CMark parsing, just an example

Oh I see, okay. Then we should try hard I suppose :smile_cat: 

---

That's just a wild guess, but as one of the measures I would try replacing sparks with `IO` threading, maybe sparks really have problems with running foreign calls in parallel.

Mm, probably to have this organized, I'd also suggest preserving all your attempts in separate branches, maybe call them `yuri/#220-experiment-xxx`. And load your reports there as separate commits too. This way someone else can play with the approaches you tried.

Martoon-00:
On the current state of this PR:
* Yet need to understand whether can we get higher CPU consumption.
* In #220 I'm asking to fully document the behaviour of `ScanAction`. If I write my own scanner, what should I do to make on strict on reading and then initialize parsing in parallel with parsing other files? What should I do to read + parse on the fly?

---

@YuriRomanowski I'm not sure if you manage to work on this PR during the remaining time, but that's fine, we can transfer this PR to someone.

In case you manage to allocate time on this, I suggest asking in Slack in Haskell channel, maybe someone will manage to explain what we observe. If not, let's suggest that we did quite a lot to make it performant and stop on what we have gained.

PR #238, opened 2022-12-14
Yuri romanowski/#64 refactor markdown scanner

### LINKED ISSUE #64: Try to implement copy-paste protection checks

YuriRomanowski:
## Description
Problem: Current implementation of the markdown scanner is hard to extend, so we need to refactor it to add support for new annotations.

Solution: Refactor: improve handling annotations, remove IMSAll state, rename functions, isolate processing annotations for different types of Nodes, sharing common behaviour with factored out functions.

## Related issue(s)

Needed to fix #64 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

#### ✓ Release Checklist

### DIFF
--- a/src/Xrefcheck/Scanners/Markdown.hs
+++ b/src/Xrefcheck/Scanners/Markdown.hs
@@ -19,14 +19,15 @@ module Xrefcheck.Scanners.Markdown
 
 import Universum
 
-import CMarkGFM
-  (Node (..), NodeType (..), PosInfo (..), commonmarkToNode, extAutolink, optFootnotes)
+import CMarkGFM (NodeType (..), PosInfo (..), commonmarkToNode, extAutolink, optFootnotes)
+import CMarkGFM qualified as C
 import Control.Lens (_Just, makeLenses, makeLensesFor, (.=))
 import Control.Monad.Trans.Writer.CPS (Writer, runWriter, tell)
 import Data.Aeson (FromJSON (..), genericParseJSON)
 import Data.ByteString.Lazy qualified as BSL
 import Data.DList qualified as DList
 import Data.Default (def)
+import Data.List (span)
 import Data.Text qualified as T
 import Data.Text.Lazy qualified as LT
 import Fmt (Buildable (..), nameF)
@@ -49,8 +50,8 @@ defGithubMdConfig = MarkdownConfig
   { mcFlavor = GitHub
   }
 
-instance Buildable Node where
-  build (Node _mpos ty mSubs) = nameF (show ty) $
+instance Buildable C.Node where
+  build (C.Node _mpos ty mSubs) = nameF (show ty) $
     maybe "[]" interpolateBlockListF (nonEmpty mSubs)
 
 toPosition :: Maybe PosInfo -> Position
@@ -67,7 +68,7 @@ toPosition = Position . \case
         |]
 
 -- | Extract text from the topmost node.
-nodeExtractText :: Node -> Text
+nodeExtractText :: (C.Node) -> Text
 nodeExtractText = T.strip . mconcat . map extractText . nodeFlatten
   where
     extractText = \case
@@ -75,8 +76,8 @@ nodeExtractText = T.strip . mconcat . map extractText . nodeFlatten
       CODE t -> t
       _ -> ""
 
-    nodeFlatten :: Node -> [NodeType]
-    nodeFlatten (Node _pos ty subs) = ty : concatMap nodeFlatten subs
+    nodeFlatten :: (C.Node) -> [NodeType]
+    nodeFlatten (C.Node _pos ty subs) = ty : concatMap nodeFlatten subs
 
 
 data IgnoreMode
@@ -140,13 +141,13 @@ initialScannerState = ScannerState
 type ScannerM a = StateT ScannerState (Writer [ScanError]) a
 
 -- | A fold over a `Node`.
-cataNode :: (Maybe PosInfo -> NodeType -> [c] -> c) -> Node -> c
-cataNode f (Node pos ty subs) = f pos ty (cataNode f <$> subs)
+cataNode :: (Maybe PosInfo -> NodeType -> [c] -> c) -> C.Node -> c
+cataNode f (C.Node pos ty subs) = f pos ty (cataNode f <$> subs)
 
--- | Sets correct @_ssParentNodeType@ before running scanner on each node
+-- | Sets correct @_ssParentNodeType@ before running scanner on each node.
 cataNodeWithParentNodeInfo
   :: (Maybe PosInfo -> NodeType -> [ScannerM a] -> ScannerM a)
-  -> Node
+  -> C.Node
   -> ScannerM a
 cataNodeWithParentNodeInfo f node = cataNode f' node
   where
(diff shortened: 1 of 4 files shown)


src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -388,23 +459,30 @@ getXrefcheckContent node =
...
 -- and annotation length.
-getPosition :: Node -> Maybe PosInfo
-getPosition node@(Node pos _ _) = do
+getPosition :: C.Node -> Maybe PosInfo
+getPosition node@(C.Node pos _ _) = do
   annLength <- length . T.strip <$> getHTMLText node
   PosInfo sl sc _ _ <- pos
   pure $ PosInfo sl sc sl (sc + annLength - 1)
 
 -- | Extract `IgnoreMode` if current node is xrefcheck annotation.
```
aeqz, line 468: I think that this comment should be updated:

`IgnoreMode` `->` `GetAnnotation`

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -340,24 +404,31 @@ nodeExtractInfo fp input@(Node _ _ nSubs) = do
...
+    isSimpleComment :: C.Node -> Bool
+    isSimpleComment node = do
+      let isComment = isJust $ getCommentContent node
+          isNotXrefcheckAnnotation = isNothing $ getXrefcheckContent node
+      isComment && isNotXrefcheckAnnotation
 
-    isIgnoreFile :: Node -> Bool
-    isIgnoreFile = (ValidMode IMAll ==) . getIgnoreMode
+    isIgnoreFile :: C.Node -> Bool
+    isIgnoreFile = (Just (IgnoreAnnotation IMAll) ==) . getAnnotation
```
aeqz, line 421: Maybe we can have a function like

`isGlobalAnnotation :: GetAnnotation -> Bool`

and use it here. Also, if this function has no wildcard pattern matchings, the compiler will ask us if we want any new annotation that we add in the future to be global or not.

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -142,114 +141,179 @@ initialScannerState = ScannerState
...
+              IMAGE {}  -> handleLink      ign pos ty subs
+              _         -> handleOther     ign pos ty subs
+
+    handleLink ::
+      Maybe Ignore ->
+      Maybe PosInfo ->
+      NodeType ->
+      [ScannerM C.Node] ->
+      ScannerM C.Node
+    handleLink ign pos ty subs = do
```
aeqz, line 188: Now all the `ign` arguments can renamed to something like `ann`. Right?

aeqz: Sorry, they always correspond to `Ignore`.

`<!-- ignore my-previous-comment -->`

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -142,114 +141,179 @@ initialScannerState = ScannerState
...
+    traverseNodeWithLinkExpected ::
+      IgnoreLinkState ->
+      Maybe PosInfo ->
+      Maybe PosInfo ->
+      NodeType ->
+      [ScannerM C.Node] ->
+      ScannerM C.Node
+    traverseNodeWithLinkExpected ignoreLinkState modePos pos ty subs = do
+      when (ignoreLinkState == ExpectingLinkInSubnodes) $
+        ssIgnore . _Just . ignoreMode .=  IMSLink ParentExpectsLink
```
aeqz, line 254: Extra space after `.=`?

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -142,114 +141,179 @@ initialScannerState = ScannerState
...
-          IMParagraph -> pure IMSParagraph
-
-          IMAll -> pure IMSAll
-
-        (ssIgnore .= Just (Ignore ignoreModeState correctPos)) $> defNode
-      InvalidMode msg -> do
+      -> GetAnnotation
+      -> ScannerM C.Node
+    handleAnnotation pos nodeType = \case
+      IgnoreAnnotation mode  -> do
```
aeqz, line 271: Extra space after `mode`?

aeqz (changes requested):
I like this refactor. It will allow us to create new annotations more easily. I have added just a few comments.

Also, maybe there will be the need for another small refactor if at some point we increase the `ScannerState` with stateful annotations that do not correspond to ignore, but it can be done at that moment.

src/Xrefcheck/Scanners/Markdown.hs
YuriRomanowski: Currently there can be only one type of annotation, which is `IgnoreFile`. [Here](https://github.com/serokell/xrefcheck/pull/240) similar function is introduced, called `isHeaderNode`.

aeqz (approved):

aeqz (approved):

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -217,28 +210,47 @@ removeIgnored fp = withIgnoreMode . cataNodeWithParentNodeInfo remove
...
+        let reportIfThereWasAnnotation :: ScannerM ()
+            reportIfThereWasAnnotation = do
+              curIgnore <- use ssIgnore
+              whenJust curIgnore $ \case
+                Ignore IMSParagraph prevPos ->
+                  lift . tell . makeError prevPos fp . ParagraphErr $ prettyType nodeType
+                Ignore (IMSLink _) prevPos ->
+                  lift $ tell $ makeError prevPos fp LinkErr
```
Martoon-00, line 279: Here too, the commit says that you apply a refactoring, but here in fact you add a behaviour that was not present before.

I more than agree that this check is good to add, but let's extract it to a separate commit.

Martoon-00 (changes requested):

src/Xrefcheck/Scanners/Markdown.hs
YuriRomanowski: In fact the behavior hasn't changed here, because in the original version some of these checks were present in other cases on the top level of `remove` function.

tests/golden/check-scan-errors/expected.gold
```diff
@@ -18,7 +18,7 @@
   ➥  In file check-scan-errors.md
      scan error at src:21:1-50:
 
-     Unrecognised option "unrecognised-annotation" perhaps you meant <"ignore link"|"ignore paragraph"|"ignore all">
+     Unrecognised option "ignore unrecognised-annotation" perhaps you meant <"ignore link"|"ignore paragraph"|"ignore all">
```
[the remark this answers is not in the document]
YuriRomanowski: OK, got it

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -140,81 +141,132 @@ initialScannerState = ScannerState
...
-                        _ -> pass
-                    return node'
-
-      when (ty == PARAGRAPH) $ use ssIgnore >>= \case
+          Nothing -> do
+            case ty of
+              PARAGRAPH -> handleParagraph ign pos ty subs
+              LINK {}   -> handleLink      ign pos ty subs
+              IMAGE {}  -> handleLink      ign pos ty subs
+              _         -> handleOther     ign pos ty subs
```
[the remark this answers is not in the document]
YuriRomanowski: For me, it's easier to reason about code behavior when a node type is known, because our types for ignore modes are, IMHO, rather confusing and hard to understand. And, in contrast, node type is something very clear and easy to work with, so I preferred it.

Things only gets worse if there are some different types of annotation (the primary reason of the whole this refactor). We'll have to handle a lot of different cases for all types of annotations and nodes.

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -140,81 +141,132 @@ initialScannerState = ScannerState
...
+    handleLink ::
+      Maybe Ignore ->
+      Maybe PosInfo ->
+      NodeType ->
+      [ScannerM C.Node] ->
+      ScannerM C.Node
+    handleLink ign pos ty subs = do
+      let traverseChildren = C.Node pos ty <$> sequence subs
+      -- It can be checked that it's correct for all the cases
+      ssIgnore .= Nothing
```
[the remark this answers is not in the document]
YuriRomanowski: > Also, AFAIU in the case of ign = IMSParagraph you rely on the behaviour of handleParagraph to strip the entire Paragraph subtree.

It seems that this code is fully made up of such implicit dependencies, because the main reason is how we use annotations. We put an annotation somewhere in the text, and then the further behavior depends on the next nodes.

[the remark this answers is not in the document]
YuriRomanowski: In case of `IMSParagraph`, at least, we could try to look at the next node (not very easy to do) and perform some actions immediately, so, if the next is paragraph, ignore it, and if the next is something else, emit an error. But in case of link annotations we have to handle all this implicit stuff (and that's sad).

YuriRomanowski:
@Martoon-00 There are a lot of troubles in the `Scanners/Markdown.hs` module that are inherited from the past, I didn't fix them fully, just refactor slightly such that adding new annotations ceased to be really impossible. But we can try to further improve it in some extent, sure.

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -167,48 +165,43 @@ removeIgnored fp = withIgnoreMode . cataNodeWithParentNodeInfo remove
...
-                case currentIgnore of
-                  Just (Ignore {_ignoreMode = IMSLink ParentExpectsLink}) -> do
-                    lift $ tell $ makeError modePos fp LinkErr
+        let mbAnnotation = getAnnotation node
+        case mbAnnotation of
+          Just ann -> handleAnnotation pos ty ann
+          Nothing -> case e of
+            Nothing                    -> Node pos ty <$> sequence subs
+            Just (Ignore mode modePos) ->
+                case (mode, ty) of
```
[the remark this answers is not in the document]
YuriRomanowski: :thinking: Where is it? It seems it's not present in the final version

src/Xrefcheck/Scanners/Markdown.hs
```diff
@@ -349,24 +404,31 @@ nodeExtractInfo fp input@(Node _ _ nSubs) = do
...
+      ignoreFile = any isIgnoreFile headerNodes
+  (ignoreFile, contentsNodes)
   where
-    isComment :: Node -> Bool
-    isComment = isJust . getCommentContent
+    isSimpleComment :: C.Node -> Bool
+    isSimpleComment node = do
+      let isComment = isJust $ getCommentContent node
+          isNotXrefcheckAnnotation = isNothing $ getXrefcheckContent node
+      isComment && isNotXrefcheckAnnotation
```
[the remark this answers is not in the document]
YuriRomanowski: Hm, no, I think it's better to stop at incorrect annotations :thinking:

Martoon-00:
Related ticket: https://github.com/serokell/xrefcheck/issues/276.

PR #243, opened 2022-12-15
[#197] Dependent RepoInfo keys example

aeqz:
## Description

Problem: During the implementation of issue #197, we though about how we could customize the way filepaths are searched within the repository data in an efficient way, depending on if filepaths are case-insensitive or not.

Solution: We implemented a GADT with its polymorphic parameter, corresponding to Map keys, hidden under an existentially quantified type. Although we end up not requiring this, we separate the essence of the idea to a commit in a separate branch, so it can be reviewed in the future.

## Related issue(s)

Relates #197 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/src/Xrefcheck/Command.hs
+++ b/src/Xrefcheck/Command.hs
@@ -25,7 +25,7 @@ import Xrefcheck.Progress (allowRewrite)
 import Xrefcheck.Scan
   (FormatsSupport, ScanError (..), ScanResult (..), reportScanErrs, scanRepo,
   specificFormatsSupport)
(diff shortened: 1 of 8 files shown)


Martoon-00:
I believe these changes may be handy if in the future we implement different scanners. As different languages may use different filepath formats when referring to the surrounding files, and these formats may have different comparison rules.

I suggested few changes to this PR [in this thread](https://github.com/serokell/xrefcheck/pull/230#discussion_r1048352850).

We have to thoroughly think about use cases again and decide how much sense all this have.

Issue #25, opened 2020-02-25
Report redirects as broken links


gromakovsky:
# Clarification and motivation

As far as I understand currently `xrefcheck` treats redirects as valid links, but usually they indicate that something is wrong. If you have a link `x` in your md file and it redirects to `y`, it will usually be the case that:
1. You can safely replace `x` with `y` and eliminate unnecessary redirection step.
2. `y` is some page that says "`x` does not exist, but you can enjoy `z` right here" (and does not make `xrefcheck` fail).

# Acceptance criteria

By default 3xx codes should be reported as broken links (I don't know these codes well enough, maybe some specific codes are ok, you are welcome to consider them ok if you find it better).

Optional: make it possible (add a flag or something) to say that 3xx codes are valid.

Note that there is another (private) issue about ignoring specific broken links.

Martoon-00:
I encountered an interesting case with redirects:
3. For example, when accessing https://github.com/serokell/xrefcheck/issues/new/choose page usually I get it opened fine with 200, but if I'm not authorized (and xrefcheck is not), I get redirected to the login page.
Till now I was thinking that all problems with not being authorized could be mitigated just via treating 401 as ok, but now it gets ambiguous.

How do you think, is it the problem on that page and we should use some another page, or xrefcheck should be able to handle such cases? If the latter holds, would it be enough if I allow specifying exclusions list?

Martoon-00:
There is another possible solution to this which may be handy in general, that is, allow local errors suppressions.

For instance, prefixing a link with `<!-- xrefcheck: allow-redirect -->` comment allows this link to produce 302.

kirelagin:
The semantics of 302 and 307 are _temporary_ redirects, so if you get them, it means that, most likely, the URL is valid, it is just altered temporarily, but for future requests you should use this same URL.

304 is only used for caching, so it is equivalent to 2xx for our purposes.

Martoon-00:
What about the case with gitlab, that on access to non-existing file it returns "302 Found" and redirects to the root of repo? 
I guess I should allow adding it an exception written down in config.

kirelagin:
As I said, I am for just accepting 302, 304, and 307 as valid unconditionally.

Martoon-00:
Ah okay, I got your point.

@gromakovsky what do you think?

gromakovsky:
> I guess I should allow adding it an exception written down in config then thinking

[...]

PR #233, opened 2022-12-07
[#218] Change redirects default behaviour

aeqz:
## Description

Problem: Xrefcheck currently always follows redirect links.

Solution: We are changing its default behaviour regarding redirect links to fail and report permanent redirects, and to pass for temporary redirects. Further PRs will allow the user to configure other policies.

## Related issue(s)

Fixes #218
Relates #25

## :white_check_mark: Checklist for your Pull Request

Ideally a PR has all of the checkmarks set.

If something in this list is irrelevant to your PR, you should still set this
checkmark indicating that you are sure it is dealt with (be that by irrelevance).

#### Related changes (conditional)

- Tests
  - [x] If I added new functionality, I added tests covering it.
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [x] I checked whether I should update the docs and did so if necessary:
    - [README](https://github.com/serokell/xrefcheck/tree/master/README.md)
    - Haddock

- Public contracts
  - [x] Any modifications of public contracts comply with the [Evolution
  of Public Contracts](https://www.notion.so/serokell/Evolution-of-Public-Contracts-2a3bf7971abe4806a24f63c84e7076c5) policy.
  - [x] I added an entry to the [changelog](https://github.com/serokell/xrefcheck/tree/master/CHANGES.md) if my changes are visible to the users
        and
  - [x] provided a migration guide for breaking changes if possible

#### Stylistic guide (mandatory)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [x] My code complies with the [style guide](https://github.com/serokell/style/blob/master/haskell.md).

Issue #218, opened 2022-11-07
[BUG] Investigate 503


dcastro:
# Description

In some of our gitlab projects, we have links such as `https://gitlab.com/morley-framework/morley-infra/-/issues/new`.

For some reason, when xrefcheck attempts to verify these links, it gets a 503 Service Temporarily Unavailable, which causes xrefcheck to fail. As a result, our gitlab projects ignore those links with a `xrefcheck: ignore link` annotation.

However, if you manually verify it using curl or using a browser, you get a 302 redirection (which would make xrefcheck pass).

* Let's investigate the reason for this discrepancy.
* If there's anything that can be doe in xrefcheck to fix this, let's do it.
* If we manage to fix it, let's remove the `xrefcheck: ignore annotation` from those gitlab projects. These are the ones I'm aware of:
    - [ ] https://gitlab.com/morley-framework/morley-debugger
    - [ ] https://gitlab.com/morley-framework/morley-infra
    - [ ] https://gitlab.com/morley-framework/indigo/-/issues

# To Reproduce

```
$ echo '[a](https://gitlab.com/morley-framework/morley-infra/-/issues/new)' > a.md
$ git add a.md
$ xrefcheck
```

```
Configuration file not found, using default config for GitHub repositories

=== Invalid references found ===                                                                                                                    

  ➥  In file a.md
     bad reference (external) at src:7:1-66:
       - text: "a"
       - link: https://gitlab.com/morley-framework/morley-infra/-/issues/new
       - anchor: -
     
     ⛂  Resource unavailable (503 Service Temporarily Unavailable)
     
Invalid references dumped, 1 in total.
```

However, if we check the same link via curl, it replies with a 302 redirect.

```
$ curl -I https://gitlab.com/morley-framework/morley-infra/-/issues/new
HTTP/2 302 
...
location: https://gitlab.com/users/sign_in
...
```

Similarly, opening the link in the browser (while not authenticated into gitlab.com) gets us redirected to `https://gitlab.com/users/sign_in`.

# Expected behavior

Expected the link to reply with a 302 and for xrefcheck to pass.

# Environment

* xrefcheck v0.2.2

Martoon-00:
AFAIR if I visit that page from the browser without being authenticated into any account, I will get 503 too. So this is an issue around authentication.

dcastro:
Just double-checked to be sure, it replies with a 302:

[...]

PR #250, opened 2022-12-19
[#25] Redirect links with configuration rules

aeqz:
## Description

Problem: We previously changed the default behaviour of Xrefcheck when following link redirects, but did not provide a way to configure it.

Solution: We are adding a new field in the configuration file to allow writing a list of redirect rules that will be applied to links that match them.

## Related issue(s)

Relates #25 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,8 +162,8 @@ There are several ways to fix this:
         - to: "https?://forbidden.com.*"
           on: 307
           outcome: invalid
-        - from: "^http://.*"
-          to: "^https://.*"
+        - from: "http://.*"
+          to: "https://.*"
           outcome: follow
       ```
 
--- a/README.md
+++ b/README.md
@@ -141,8 +141,36 @@ There are several ways to fix this:
     * This behavior can be disabled by setting `ignoreAuthFailures: false` in the config file.
 
 1. How does `xrefcheck` handle redirects?
-    * Permanent redirects (i.e. 301 and 308) are reported as errors.
-    * Temporary redirects (i.e. 302, 303 and 307) are assumed to be valid.
+    * The rules from the default configuration are as follows:
+      * Permanent redirects (i.e. 301 and 308) are reported as errors.
+      * Temporary redirects (i.e. 302, 303 and 307) are assumed to be valid.
+    * Redirect rules can be specified with the `externalRefRedirects` parameter within `networking`, which accepts an array of
+      rules with keys `from`, `to`, `on` and `outcome`. The rule applied is the first one that matches with
+      the `from`, `to` and `on` fields, if any, where
+      * `from` is a regular expression, as in `ignoreExternalRefsTo`, for the source link in a single redirection step. Its absence means that
+        every link matches.
+      * `to` is a regular expression for the target link in a single redirection step. Its absence also means that every link matches.
+      * `on` accepts `temporary`, `permanent` or a specific redirect HTTP code. Its absence also means that
+        every response code matches.
+      * The `outcome` parameter accepts `valid`, `invalid` or `follow`. The last one follows the redirect by applying the
+        same configuration rules.
+
+      For example, this configuration forbids 307 redirects to a specific domain and makes redirections from HTTP to HTTPS to be followed:
+
+      ```
+      externalRefRedirects:
+        - to: "https?://forbidden.com.*"
+          on: 307
+          outcome: invalid
+        - from: "^http://.*"
+          to: "^https://.*"
+          outcome: follow
+      ```
+
+      The first one applies if both of them match.
+
+    * The number of redirects allowed in a single redirect chain is limited and can be configured with the
+      `maxRedirectFollows` parameter, also within `networking`. A number smaller than 0 disables the limit.
 
 1. How does `xrefcheck` handle localhost links?
     * By default, `xrefcheck` will ignore links to localhost.
--- a/src/Xrefcheck/Config.hs
+++ b/src/Xrefcheck/Config.hs
@@ -7,6 +7,7 @@
 
 module Xrefcheck.Config
   ( module Xrefcheck.Config
+  , module Xrefcheck.Data.Redirect
   , defConfigText
   ) where
 
@@ -15,11 +16,12 @@ import Universum
 import Control.Lens (makeLensesWith)
 import Data.Aeson (genericParseJSON)
 import Data.Yaml (FromJSON (..), decodeEither', prettyPrintParseException, withText)
-
+import Text.Regex.TDFA.Text ()
 import Time (KnownRatName, Second, Time (..), unitsP)
 
 import Xrefcheck.Config.Default
 import Xrefcheck.Core
+import Xrefcheck.Data.Redirect
 import Xrefcheck.Scan
 import Xrefcheck.Scanners.Markdown
 import Xrefcheck.Util (Field, aesonConfigOption, postfixFields)
@@ -78,7 +80,15 @@ data NetworkingConfig' f = NetworkingConfig
     -- this `maxTimeoutRetries` option limits only the number of retries
     -- caused by timeouts, and `maxRetries` limits the number of retries
     -- caused both by 429s and timeouts.
+  , ncMaxRedirectFollows        :: Field f Int
+    -- ^ Maximum number of links that can be followed in a single redirect
+    -- chain.
+  , ncExternalRefRedirects      :: Field f RedirectConfig
+    -- ^  Rules to override the redirect behavior for external references.
   } deriving stock (Generic)
+
+-- | A list of custom redirect rules.
+type RedirectConfig = [RedirectRule]
 
 -- | Type alias for ScannersConfig' with all required fields.
 type ScannersConfig = ScannersConfig' Identity
@@ -118,7 +128,8 @@ overrideConfig config
 
     defScanners = cScanners $ defConfig flavor
     defExclusions = cExclusions $ defConfig flavor
-    defNetworking = cNetworking $ defConfig flavor
+    defNetworking = cNetworking (defConfig flavor)
+      & ncExternalRefRedirectsL .~ []
 
     overrideExclusions exclusionConfig
       = ExclusionConfig
@@ -138,6 +149,8 @@ overrideConfig config
         , ncDefaultRetryAfter         = overrideField ncDefaultRetryAfter
         , ncMaxRetries                = overrideField ncMaxRetries
         , ncMaxTimeoutRetries         = overrideField ncMaxTimeoutRetries
+        , ncMaxRedirectFollows        = overrideField ncMaxRedirectFollows
+        , ncExternalRefRedirects      = overrideField ncExternalRefRedirects
         }
       where
         overrideField :: (forall f. NetworkingConfig' f -> Field f a) -> a
--- a/src/Xrefcheck/Config/Default.hs
+++ b/src/Xrefcheck/Config/Default.hs
@@ -7,6 +7,7 @@ module Xrefcheck.Config.Default where
 
 import Universum
 
+import Fmt (Builder)
 import Text.Interpolation.Nyan
 
 import Xrefcheck.Core
@@ -65,21 +66,44 @@ networking:
   # On other errors xrefcheck fails immediately, without retrying.
   maxRetries: 3
 
-# Querying a given domain that ever returned 429 before,
-# this defines how many timeouts are allowed during retries.
-#
-# For such domains, timeouts likely mean hitting the rate limiter,
-# and so xrefcheck considers timeouts in the same way as 429 errors.
-#
-# For other domains, a timeout results in a respective error, no retry
-# attempts will be performed. Use `externalRefCheckTimeout` option
-# to increase the time after which timeout is declared.
-#
-# This option is similar to `maxRetries`, the difference is that
-# this `maxTimeoutRetries` option limits only the number of retries
-# caused by timeouts, and `maxRetries` limits the number of retries
-# caused both by 429s and timeouts.
+  # Querying a given domain that ever returned 429 before,
+  # this defines how many timeouts are allowed during retries.
+  #
+  # For such domains, timeouts likely mean hitting the rate limiter,
+  # and so xrefcheck considers timeouts in the same way as 429 errors.
+  #
+  # For other domains, a timeout results in a respective error, no retry
+  # attempts will be performed. Use `externalRefCheckTimeout` option
+  # to increase the time after which timeout is declared.
+  #
+  # This option is similar to `maxRetries`, the difference is that
+  # this `maxTimeoutRetries` option limits only the number of retries
+  # caused by timeouts, and `maxRetries` limits the number of retries
+  # caused both by 429s and timeouts.
   maxTimeoutRetries: 1
+
+  # Maximum number of links that can be followed in a single redirect
+  # chain.
+  #
+  # The link is considered as invalid if the limit is exceeded.
+  maxRedirectFollows: 10
+
+  # Rules to override the redirect behavior for external references that
+  # match, where
+  #   - 'from' is a regular expression for the source link in a single
+  #     redirection step. Its absence means that every link matches.
+  #   - 'to' is a regular expression for the target link in a single
+  #     redirection step. Its absence also means that every link matches.
+  #   - 'on' accepts 'temporary', 'permanent' or a specific redirect HTTP code.
+  #     Its absence also means that every response code matches.
+  #   - 'outcome' accepts 'valid', 'invalid' or 'follow'. The last one follows
+  #      the redirect by applying the same configuration rules so, for instance,
+  #      exclusion rules would also apply to the following links.
+  #
+  # The first one that matches is applied, and the link is considered
+  # as valid if none of them does match.
+  externalRefRedirects:
+#{interpolateIndentF 4 externalRefRedirects}
 
 # Parameters of scanners for various file types.
 scanners:
@@ -95,7 +119,7 @@ scanners:
 |]
   where
     ignoreLocalRefsFrom :: NonEmpty Text
-    ignoreLocalRefsFrom = fromList $  case flavor of
+    ignoreLocalRefsFrom = fromList $ case flavor of
       GitHub ->
         [ ".github/pull_request_template.md"
         , ".github/issue_template.md"
@@ -108,7 +132,7 @@ scanners:
         ]
 
     ignoreLocalRefsTo :: NonEmpty Text
-    ignoreLocalRefsTo = fromList $  case flavor of
+    ignoreLocalRefsTo = fromList $ case flavor of
       GitHub ->
         [ "../../../issues"
         , "../../../issues/*"
@@ -121,3 +145,20 @@ scanners:
         , "../../merge_requests"
         , "../../merge_requests/*"
         ]
+
+    externalRefRedirects :: Builder
+    externalRefRedirects = case flavor of
+      GitHub ->
+        [int||
+        - on: permanent
+          outcome: invalid|]
+      GitLab ->
+        [int||
+        - on: permanent
+          outcome: invalid
+        # GitLab redirects non-existing files to the repository's main page
+        # with a 302 code instead of answering with a 404 response.
+        - from: https?://gitlab.com/.*/-/blob/.*
+          to: https?://gitlab.com/.*
+          on: 302
+          outcome: invalid|]
--- a/src/Xrefcheck/Config/Default.hs
+++ b/src/Xrefcheck/Config/Default.hs
@@ -37,7 +37,7 @@ exclusions:
   # List of POSIX extended regular expressions.
   ignoreExternalRefsTo:
     # Ignore localhost links by default
-    - ^(https?|ftps?)://(localhost|127\\.0\\.0\\.1).*
+    - (https?|ftps?)://(localhost|127\\.0\\.0\\.1).*
 
 # Networking parameters.
 networking:
--- a/src/Xrefcheck/Data/Redirect.hs
+++ b/src/Xrefcheck/Data/Redirect.hs
@@ -1,0 +1,170 @@
+{- SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io>
+ -
+ - SPDX-License-Identifier: MPL-2.0
+ -}
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+module Xrefcheck.Data.Redirect
+  ( RedirectChain
+  , RedirectChainLink (..)
+  , emptyChain
+  , pushRequest
+  , hasRequest
+  , totalFollowed
+
+  , RedirectRule (..)
+  , RedirectRuleOn (..)
+  , RedirectRuleOutcome (..)
+  , redirectRule
+
+  , isPermanentRedirectCode
+  , isRedirectCode
+  , isTemporaryRedirectCode
+  ) where
+
+import Universum
+
+import Data.Aeson (genericParseJSON)
+import Data.Yaml (FromJSON (..), withText)
+import Fmt (Buildable (..))
+import Text.Regex.TDFA.Text (Regex)
+
+import Data.Sequence ((|>))
+import Xrefcheck.Scan ()
+import Xrefcheck.Util
+
+-- | A custom redirect rule.
+data RedirectRule = RedirectRule
+  { rrFrom :: Maybe Regex
+    -- ^ Redirect source links that match to apply the rule.
+    --
+    -- 'Nothing' matches any link.
+  , rrTo :: Maybe Regex
+    -- ^ Redirect target links that match to apply the rule.
+    --
+    -- 'Nothing' matches any link.
+  , rrOn :: Maybe RedirectRuleOn
+    -- ^ HTTP code selector to apply the rule.
+    --
+    -- 'Nothing' matches any code.
+  , rrOutcome :: RedirectRuleOutcome
+    -- ^ What to do when an HTTP response matches the rule.
+  } deriving stock (Generic)
+
+-- | Rule selector depending on the response HTTP code.
+data RedirectRuleOn
+    = RROCode Int
+      -- ^ An exact HTTP code
+    | RROPermanent
+      -- ^ Any HTTP code considered as permanent according to 'isPermanentRedirectCode'
+    | RROTemporary
+      -- ^ Any HTTP code considered as permanent according to 'isTemporaryRedirectCode'
+    deriving stock (Show, Eq)
+
+-- | What to do when receiving a redirect HTTP response.
+data RedirectRuleOutcome
+    = RROValid
+      -- ^ Consider it as valid
+    | RROInvalid
+      -- ^ Consider it as invalid
+    | RROFollow
+      -- ^ Try again by following the redirect
+    deriving stock (Show, Eq)
+
+-- | Links in a redirection chain.
+newtype RedirectChain = RedirectChain
+  { unRedirectChain :: Seq RedirectChainLink
+  } deriving newtype (Show, Eq)
+
+-- | A single link in a redirection chain.
+newtype RedirectChainLink = RedirectChainLink
+  { unRedirectChainLink :: Text
+  } deriving newtype (Show, Eq)
+
+instance FromList RedirectChain where
+  type ListElement RedirectChain = Text
+  fromList = RedirectChain . fromList . fmap RedirectChainLink
+
+emptyChain :: RedirectChain
+emptyChain = RedirectChain mempty
+
+pushRequest :: RedirectChain -> RedirectChainLink -> RedirectChain
+pushRequest (RedirectChain chain) = RedirectChain . (chain |>)
+
+hasRequest :: RedirectChain -> RedirectChainLink -> Bool
+hasRequest (RedirectChain chain) = (`elem` chain)
+
+totalFollowed :: RedirectChain -> Int
+totalFollowed = length . unRedirectChain
+
+instance Buildable RedirectChain where
+  build (RedirectChain linksStack) = build chainText
+    where
+      link (True, RedirectChainLink l) = "-| " <> l
+      link (False, RedirectChainLink l) = "-> " <> l
+
+      chainText = mconcat
+        $ intersperse "\n"
+        $ fmap link
+        $ zip (True : repeat False)
+        $ toList linksStack
+
+-- | Redirect rule to apply to a link when it has been responded with a given
+-- HTTP code.
+redirectRule :: Text -> Text -> Int -> [RedirectRule] -> Maybe RedirectRule
+redirectRule source target code rules =
+  find (matchRule source target code) rules
+
+-- | Check if a 'RedirectRule' matches a given link and HTTP code.
+matchRule :: Text -> Text -> Int -> RedirectRule -> Bool
+matchRule source target code RedirectRule{..} = and
+  [ matchCode
+  , matchLink source rrFrom
+  , matchLink target rrTo
+  ]
+  where
+    matchCode = case rrOn of
+      Nothing -> True
+      Just RROPermanent -> isPermanentRedirectCode code
+      Just RROTemporary -> isTemporaryRedirectCode code
+      Just (RROCode other) -> code == other
+
+    matchLink link = \case
+      Nothing -> True
+      Just regex -> doesMatchAnyRegex link [regex]
+
+isRedirectCode :: Int -> Bool
+isRedirectCode code = code >= 300 && code < 400
+
+isTemporaryRedirectCode :: Int -> Bool
+isTemporaryRedirectCode = flip elem [302, 303, 307]
+
+isPermanentRedirectCode :: Int -> Bool
+isPermanentRedirectCode = flip elem [301, 308]
+
+instance FromJSON (RedirectRule) where
+  parseJSON = genericParseJSON aesonConfigOption
+
+instance FromJSON (RedirectRuleOutcome) where
+  parseJSON = withText "Redirect rule outcome" $
+    \case
+      "valid" -> pure RROValid
+      "invalid" -> pure RROInvalid
+      "follow" -> pure RROFollow
+      _ -> fail "expected (valid|invalid|follow)"
+
+instance FromJSON (RedirectRuleOn) where
+  parseJSON v = code v
+    <|> text v
+    <|> fail "expected a redirect (3XX) HTTP code or (permanent|temporary)"
+    where
+      code cv = do
+        i <- parseJSON cv
+        guard $ isRedirectCode i
+        pure $ RROCode i
+      text = withText "Redirect rule on" $
+        \case
+          "permanent" -> pure RROPermanent
+          "temporary" -> pure RROTemporary
+          _ -> mzero
--- a/src/Xrefcheck/Util.hs
+++ b/src/Xrefcheck/Util.hs
@@ -10,6 +10,7 @@ module Xrefcheck.Util
   , (-:)
   , aesonConfigOption
   , composeFuncList
+  , doesMatchAnyRegex
   , posixTimeToTimeSecond
   , utcTimeToTimeSecond
   , unlessFunc
@@ -29,6 +30,7 @@ import Data.Time (UTCTime)
 import Data.Time.Clock (nominalDiffTimeToSeconds)
 import Data.Time.Clock.POSIX (POSIXTime, utcTimeToPOSIXSeconds)
 import Fmt (Builder)
+import Text.Regex.TDFA.Text (Regex, regexec)
 import Time (Second, Time (..), sec)
 
 import Xrefcheck.Util.Colorize
@@ -72,3 +74,12 @@ whenFunc False _ = id
 
 unlessFunc :: Bool -> (a -> a) -> (a -> a)
 unlessFunc = whenFunc . not
+
+doesMatchAnyRegex :: Text -> ([Regex] -> Bool)
+doesMatchAnyRegex src = any $ \regex ->
+  case regexec regex src of
+    Right res -> case res of
+      Just (before, match, after, _) ->
+        null before && null after && not (null match)
+      Nothing -> False
+    Left _ -> False
--- a/src/Xrefcheck/Verify.hs
+++ b/src/Xrefcheck/Verify.hs
@@ -49,7 +49,7 @@ import Data.Text.Metrics (damerauLevenshteinNorm)
 import Data.Time (UTCTime, defaultTimeLocale, formatTime, readPTime, rfc822DateFormat)
 import Data.Time.Clock.POSIX (getPOSIXTime)
 import Data.Traversable (for)
-import Fmt (Buildable (..), fmt, maybeF, nameF)
+import Fmt (Buildable (..), Builder, fmt, maybeF, nameF)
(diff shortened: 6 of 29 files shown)


aeqz:
I have created this draft PR to discuss some things. This work is the continuation of #233 for allowing to define redirect rules in the configuration file.

src/Xrefcheck/Verify.hs
```diff
@@ -255,6 +261,24 @@ instance Given ColorMode => Buildable VerifyError where
         #{redirectedUrl}
       |]
 
+    RedirectCycle _ ->
+      [int||
+      Cycle found in the following redirect chain:
+      TODO print chain
```
aeqz, line 267: I'm thinking about how to show a redirect chain of links. How about this?

```
Cycle found in the following redirect chain:
  -| http://github.com/some
  -> https://github.com/some
  -> https://github.com/other
```

Martoon-00: Wow, cool. I like it.

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Description of f1526d8261337fa729059f9f6c591a9e7c1fa633 does not follow the template.<br />
Try <code>Problem:</code>/<code>Solution:</code> structure.<br />
If you really have to, you can add <code>And yes, I don't care about templates</code> to the commit message body.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Description of 6f1bc0085fdaacf8208536c22710edd2c4d9d99f does not follow the template.<br />
Try <code>Problem:</code>/<code>Solution:</code> structure.<br />
If you really have to, you can add <code>And yes, I don't care about templates</code> to the commit message body.<br />
See also [Note].</td>
    </tr>
  </tbody>
</table>

</p>

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

src/Xrefcheck/Config.hs
```diff
@@ -62,8 +61,86 @@ data NetworkingConfig' f = NetworkingConfig
...
+doesMatchAnyRegex src = any $ \regex ->
+  case regexec regex src of
+    Right res -> case res of
+      Just (before, match, after, _) ->
+        null before && null after && not (null match)
+      Nothing -> False
+    Left _ -> False
+
+-- | A list of custom redirect rules.
+type RedirectConfig = [RedirectRule]
```
aeqz, line 78: Should I move all this new redirect-related code to a dedicated module, or it is fine to let it here?

README.md
```diff
@@ -141,10 +141,20 @@ There are several ways to fix this:
...
 1. How does `xrefcheck` handle redirects?
-    * Permanent redirects (i.e. 301 and 308) are reported as errors.
-    * Temporary redirects (i.e. 302, 303 and 307) are assumed to be valid.
-
-1. How does `xrefcheck` handle localhost links?
+    * The default rules are as follows:
+      * Permanent redirects (i.e. 301 and 308) are reported as errors.
+      * Temporary redirects (i.e. 302, 303 and 307) are assumed to be valid.
+    * The redirect rules can be specified with the `externalRefRedirects` within `networking`, which accepts an array of
+      rules like `{to: "^http://.*", on: temporary, outcome: follow }`. The rule applied is the first one that matches with
```
aeqz, line 148: It remains to decide if the rule that applies is the first or the last that matches.

src/Xrefcheck/Config/Default.hs
```diff
@@ -56,6 +56,16 @@ networking:
   # a "429 Too Many Requests" response.
   maxRetries: 3
 
+  # Rules to override the redirect behavior for external references that
+  # match.
+  externalRefRedirects:
```
aeqz, line 61: I have moved the default behaviour from code to default rules.

src/Xrefcheck/Verify.hs
```diff
@@ -723,15 +751,18 @@ checkExternalResource Config{..} link
 
       let maxTime = Time @Second $ unTime ncExternalRefCheckTimeout * timeoutFrac
 
-      mres <- liftIO (timeout maxTime $ void reqLink) `catch`
-        (either throwError (\() -> return (Just ())) . interpretErrors)
-      maybe (throwError $ ExternalResourceSomeError "Response timeout") pure mres
-
-    isTemporaryRedirectCode :: Int -> Bool
-    isTemporaryRedirectCode = flip elem [302, 303, 307]
+      reqRes <- catch (liftIO (timeout maxTime $ reqLink $> RRDone)) $ \httpErr ->
```
aeqz, line 754: Now, instead of `Maybe ()`, the result is `Maybe ResponseResult`, where `Nothing` means timeout and `Just ResponseResult` tells "Done" or "Follow this redirect".

src/Xrefcheck/Config.hs
```diff
@@ -54,8 +54,86 @@ data NetworkingConfig' f = NetworkingConfig
   , ncMaxRetries                :: Field f Int
     -- ^ How many attempts to retry an external link after getting
     -- a "429 Too Many Requests" response.
+  , ncExternalRefRedirects      :: RedirectConfig
+    -- ^  Rules to override the redirect behavior for external references.
```
Martoon-00, line 58: I would actually make it optional and treat it as if the user said: I don't care about redirects, handle them somehow. And I would default this to `[]`.

And later when we see a redirect that matched no rule, we would treat that as OK, since we have to be optimistic.

What do you think?

Martoon-00 (changes requested):
Thanks, that's very decent work :+1: 

Most progress on the redirects is already behind.

I left numerous comments, and one more thing I wanted to suggest: please create a ticket for displaying redirects in the progress bar.

src/Xrefcheck/Verify.hs
```diff
@@ -725,18 +757,26 @@ checkExternalResource Config{..} link
...
         HttpExceptionRequest _ exc -> case exc of
           StatusCodeException resp _
-            | isPermanentRedirectCode code -> Left
-              . PermanentRedirectError link
-              . fmap decodeUtf8
-              . lookup "Location"
-              $ responseHeaders resp
-            | isTemporaryRedirectCode code -> Right ()
-            | isAllowedErrorCode code -> Right ()
+            | Just RedirectRule{..} <- redirectRule link code ncExternalRefRedirects ->
```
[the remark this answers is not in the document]
aeqz: Right! I will add that check.

> Also, what if we see a redirect but no rule matched?

Before this PR, I explicitly wrote code for fail on permanent redirects and succeed on temporary ones. Now I turned that into the default configuration rules, so in case that no rule matched I think that it will produce an `ExternalHttpResourceUnavailable` verification error.

I'm changing it to pass in that case, but we can continue discussing it (I think that there are other comments related with that).

src/Xrefcheck/Verify.hs
```diff
@@ -241,19 +246,51 @@ instance Given ColorMode => Buildable VerifyError where
...
+          Just RROPermanent -> "Permanent redirect"
+          Just RROTemporary -> "Temporary redirect"
+          Just (RROCode code) -> show code <> "redirect"
+
+buildRedirectChain :: NonEmpty Text -> [Text] -> Builder
+buildRedirectChain (l :| ls) attached
+  = build chainText <> build attachedText
+  where
+    link (True, c) = "-| " <> c
+    link (False, c) = "-> " <> c
```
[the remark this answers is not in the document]
aeqz: A link in the redirect chain, I will give it a proper name.

src/Xrefcheck/Verify.hs
```diff
@@ -241,19 +246,51 @@ instance Given ColorMode => Buildable VerifyError where
...
+      [int||
+      Cycle found in the following redirect chain:
+      #{interpolateIndentF 2 $ buildRedirectChain chain ["^-- here"]}
+      |]
+
+    RedirectChainBroken chain ->
       [int||
-      Permanent redirect found:
-      #{url}
+      Missin location header in the following redirect chain:
```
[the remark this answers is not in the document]
aeqz: I'm not sure if it is necessary because there is a `^-- here` indication pointing to the last link in the chain, but I can modify it if you prefer.

Actually, there is no example with that output in the Golden tests. I will try to find an example to include it.

src/Xrefcheck/Config.hs
```diff
@@ -145,6 +228,32 @@ instance FromJSON (NetworkingConfig' Maybe) where
...
+      _ -> fail "expected (valid|invalid|follow)"
+
+instance FromJSON (RedirectRuleOn) where
+  parseJSON v = code v
+    <|> text v
+    <|> fail "expected a redirect (3XX) HTTP code or (permanent|temporary)"
+    where
+      code cv = do
+        i <- parseJSON @Int cv
+        unless (isRedirectCode i) $ fail mempty
```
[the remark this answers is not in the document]
aeqz: I set the error message as empty in both auxiliary parsers (`code` and `text`) because they are used as follows:

```
code <|> text <|> fail "A common error message"
```

Is there a better way to combine both error messages?

README.md
```diff
@@ -141,10 +141,20 @@ There are several ways to fix this:
...
+      the `to` and `on` parameters, if any, where
+      * `to` is a regular expression as in `ignoreExternalRefsTo` and its absence means that every link
+        matches.
+      * `on` accepts `temporary`, `permanent` or a specific redirect HTTP code. Its absence also means that
+        every response code matches.
+      * The `outcome` parameter accepts `valid`, `invalid` or `follow`. The last one follows the redirect by applying the
+        same configuration rules.
+
+2. How does `xrefcheck` handle localhost links?
```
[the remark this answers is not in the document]
aeqz: Sorry! My IDE automatically changed it 🙃

src/Xrefcheck/Verify.hs
```diff
@@ -444,7 +481,7 @@ verifyReference
             let shownFilepath = dropWhile isPathSeparator (toString rLink)
             canonicalPath <- riRoot </ shownFilepath
             checkRef rAnchor riRoot canonicalPath shownFilepath
-          RIExternal -> checkExternalResource config rLink
+          RIExternal -> checkExternalResource mempty config rLink
```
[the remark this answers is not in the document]
aeqz: Completely agree. In fact, previously, when xrefcheck automatically followed redirects, it was also applying the retrying logic to the entire chain in a hidden way.

Now that this redirect following will be explicit in our code, maybe we can handle it in a different way.

src/Xrefcheck/Verify.hs
```diff
@@ -255,6 +261,24 @@ instance Given ColorMode => Buildable VerifyError where
         #{redirectedUrl}
       |]
 
+    RedirectCycle _ ->
+      [int||
+      Cycle found in the following redirect chain:
+      TODO print chain
+      |]
```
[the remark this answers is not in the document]
aeqz: Right! We can add a configuration parameter for this that defaults to 10 (this is the same default value as in the HTTP client library that we are using, but other suggestions are also accepted).

README.md
```diff
@@ -141,10 +141,20 @@ There are several ways to fix this:
...
+    * The redirect rules can be specified with the `externalRefRedirects` within `networking`, which accepts an array of
+      rules like `{to: "^http://.*", on: temporary, outcome: follow }`. The rule applied is the first one that matches with
+      the `to` and `on` parameters, if any, where
+      * `to` is a regular expression as in `ignoreExternalRefsTo` and its absence means that every link
+        matches.
+      * `on` accepts `temporary`, `permanent` or a specific redirect HTTP code. Its absence also means that
+        every response code matches.
+      * The `outcome` parameter accepts `valid`, `invalid` or `follow`. The last one follows the redirect by applying the
+        same configuration rules.
+
```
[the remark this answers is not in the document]
aeqz: This comment made me realize that I implemented `to` as if it meant `from`. I am changing it, although it implies that I have also to change some previously implemented tests: I was delaying the check of having a location header in the response after the rules check, and now the location has to be obtained beforehand.

src/Xrefcheck/Config.hs
aeqz: But then we would be changing the default behavior that we previously implemented. Right? It is now provided with these two default rules:

```
- to: .*
  on: permanent
  outcome: invalid
- to: .*
  on: temporary
  outcome: valid
```

src/Xrefcheck/Verify.hs
```diff
@@ -682,8 +689,11 @@ parseUri link = do
 checkExternalResource :: [Text] -> Config -> Text -> IO (VerifyResult VerifyError)
 checkExternalResource followed config@Config{..} link
   | isIgnored = return mempty
+  | link `elem` followed = pure $
+      VerifyResult [RedirectChainCycle (link :| followed)]
+  | ncMaxRedirectFollows >= 0 && length followed > ncMaxRedirectFollows = pure $
```
Martoon-00: Which purpose this `ncMaxRedirectFollows >= 0`  serves?

src/Xrefcheck/Config/Default.hs
```diff
@@ -56,6 +56,10 @@ networking:
   # a "429 Too Many Requests" response.
   maxRetries: 3
 
+  # Maximum number of links that can be followed in a single redirect
+  # chain.
+  maxRedirectFollows: 10
```
Martoon-00, line 61: This is a bit subjective concern: for me as a user, an immediate question would be: what happens if this number is exceeded? Do we skip the check, or something else? I would put a very brief clarification.

src/Xrefcheck/Util.hs
```diff
@@ -59,3 +61,12 @@ posixTimeToTimeSecond posixTime =
...
 utcTimeToTimeSecond = posixTimeToTimeSecond . utcTimeToPOSIXSeconds
+
+doesMatchAnyRegex :: Text -> ([Regex] -> Bool)
+doesMatchAnyRegex src = any $ \regex ->
+  case regexec regex src of
+    Right res -> case res of
+      Just (before, match, after, _) ->
+        null before && null after && not (null match)
+      Nothing -> False
+    Left _ -> False
```
Martoon-00, line 72: Good :+1:

src/Xrefcheck/Verify.hs
```diff
@@ -71,6 +71,7 @@ import URI.ByteString qualified as URIBS
 
 import Xrefcheck.Config
 import Xrefcheck.Core
+import Xrefcheck.Data.Redirect
```
Martoon-00, line 74: Minor suggestion: since this is used only as part of config, I would re-export this in `Xrefcheck.Config`.

You can add an entire module re-export by placing e.g.

```
  ( -- export list
  , module Xrefcheck.Data.Redirect  -- this module should be imported too 
  , ... 
```

src/Xrefcheck/Data/Redirect.hs
```diff
@@ -58,6 +65,39 @@ data RedirectRuleOutcome
...
+  fromList = RedirectChain . reverse
+
+pushRequest :: RedirectChain -> Text -> RedirectChain
+pushRequest (RedirectChain chain) = RedirectChain . (: chain)
+
+hasRequest :: RedirectChain -> Text -> Bool
+hasRequest (RedirectChain chain) = (`elem` chain)
+
+totalFollowed :: RedirectChain -> Int
+totalFollowed = length . unRedirectChain
```
Martoon-00, line 87: Nice :+1:

tests/Test/Xrefcheck/ConfigSpec.hs
```diff
@@ -35,7 +35,7 @@ test_config =
         | flavor <- allFlavors]
   , testGroup "Filled default config matches the expected format"
     -- The config we match against can be regenerated with
-    -- stack exec xrefcheck -- dump-config -t GitHub -o tests/configs/github-config.yaml
+    -- stack exec xrefcheck -- dump-config -t GitHub -o tests/configs/github-config.yaml --force
```
Martoon-00, line 38: By the way, I think this should not be a fixup commit. It's fine to include in this PR, but I would form it as a standalone commit saying that it is a follow-up change after another PR (which is perfectly fine, it happens).

src/Xrefcheck/Verify.hs
```diff
@@ -137,9 +136,15 @@ data VerifyError
   | ExternalFtpException FTPException
   | FtpEntryDoesNotExist FilePath
   | ExternalResourceSomeError Text
-  | PermanentRedirectError Text (Maybe Text)
+  | RedirectChainCycle (NonEmpty Text)
+  | RedirectChainBroken (NonEmpty Text)
+  | RedirectRuleError (Maybe RedirectRuleOn) (NonEmpty Text) (Maybe Text)
```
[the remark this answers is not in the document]
Martoon-00: I still think that at least the last field is worth clarification.

src/Xrefcheck/Config.hs
Martoon-00: Right. I think we really want two things:
1. By default, suggest to the user those defaults you implemented in the previous PR: temporary=ok, permanent=failure.
2. Give the user a simple way to say "don't bother me with redirects".

And I think, making `ncExternalRefRedirects` optional and defaulting it to `[]` that would then be interpreted as "always ok" is the best way to handle the second point.

And IMO insisting on the old behaviour is not so important, different users may find it weird depending on the use cases they face in real life. It is easy to obtain our recommended defaults for config, and I think that's enough.

A made a little detour with my reasoning for this, but this is how I see it now. There are other options to approach all this, but to me `[]` default yet seems to be the winning one.

README.md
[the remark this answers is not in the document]
Martoon-00: Ah, I see.

If you find that it would be easy to remain both `to` and `from` (given that you have both implementations in the git history), please go for it.

README.md
```diff
@@ -141,8 +141,36 @@ There are several ways to fix this:
...
+      * `from` is a regular expression, as in `ignoreExternalRefsTo`, for the source link in the redirection. Its absence means that
+        every link matches.
+      * `to` is a regular expression for the target redirect link. Its absence also means that every link matches.
+      * `on` accepts `temporary`, `permanent` or a specific redirect HTTP code. Its absence also means that
+        every response code matches.
+      * The `outcome` parameter accepts `valid`, `invalid` or `follow`. The last one follows the redirect by applying the
+        same configuration rules.
+
+      For example, this configuration forbids 307 redirects to a specific domain and makes redirections from HTTP to HTTPS to be followed:
+
```
Martoon-00, line 159: I have found some trailing whitespaces here:
```suggestion:-0+0

```




</p>

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Trailing whitespaces and/or incorrect end-of-file newlines detected.</td>
    </tr>
  </tbody>
</table>

</p>

tests/Test/Xrefcheck/ConfigSpec.hs
aeqz: Right! Actually, this change should not be here. I have already changed it in the appropriate PR.

Martoon-00: Ah yeah, I actually saw it there after reviewing this PR.

src/Xrefcheck/Verify.hs
aeqz: It is because I allowed to disable the limit when given a negative value. Do you see it useful? For example:

- Limit set to 2: follow 2 redirects at most.
- Limit set to 0: no redirect following allowed.
- Limit set to -1: no limit in the redirect following.

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
aeqz: I have created #260 for it 👍

src/Xrefcheck/Verify.hs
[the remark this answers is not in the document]
aeqz: It corresponds to the last step in the redirect chain. In fact, I think that it would be better if it is included in the same redirect chain as in the other error constructors.

src/Xrefcheck/Config.hs
YuriRomanowski: We can use [] for the defaults, but add our previous behaviour as an example in the default config comments.

src/Xrefcheck/Config/Default.hs
```diff
@@ -56,6 +56,35 @@ networking:
...
+  # Rules to override the redirect behavior for external references that
+  # match, where
+  #   - 'from' is a regular expression for the source link in the redirection.
+  #     Its absence means that every link matches.
+  #   - 'to' is a regular expression for the target link in the redirection.
+  #     Its absence also means that every link matches.
+  #   - 'on' accepts 'temporary', 'permanent' or a specific redirect HTTP code.
+  #     Its absence also means that every response code matches.
+  #   - 'outcome' accepts 'valid', 'invalid' or 'follow'. The last one follows
+  #      the redirect by applying the same configuration rules.
```
YuriRomanowski, line 74: Hm, what is meant by "applying the same configuration rules"? That, when a rule with `follow` matches, we process the next link with same configured list of rules? But we apply them always, why should we mention it separately?

README.md
```diff
@@ -141,8 +141,36 @@ There are several ways to fix this:
...
+      ```
+      externalRefRedirects:
+        - to: "https?://forbidden.com.*"
+          on: 307
+          outcome: invalid
+        - from: "^http://.*"
+          to: "^https://.*"
+          outcome: follow
+      ```
```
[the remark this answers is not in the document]
aeqz: I would agree with simplifying every regex in our configuration and examples, but the function that was being used for regex match checking in `ignoreExternalRefsTo` is the following:

```haskell
doesMatchAnyRegex :: Text -> ([Regex] -> Bool)
doesMatchAnyRegex src = any $ \regex ->
  case regexec regex src of
    Right res -> case res of
      Just (before, match, after, _) ->
        -- Requires a whole match: nothing before and nothing after
        null before && null after && not (null match)
      Nothing -> False
    Left _ -> False
```

Do you know if that was decided for some reason? It seems that you expected this to match only part of the string, which we could easily achieve with the following change:

```haskell
doesMatchAnyRegex :: Text -> ([Regex] -> Bool)
doesMatchAnyRegex src = any $ \regex ->
  case regexec regex src of
    Right res -> case res of
      Just (before, match, after, _) -> not (null match)
      Nothing -> False
    Left _ -> False
```

src/Xrefcheck/Data/Redirect.hs
```diff
@@ -0,0 +1,162 @@
...
+newtype RedirectChain = RedirectChain
+  { unRedirectChain :: [Text]
+  } deriving newtype (Show, Eq)
+
+instance FromList RedirectChain where
+  type ListElement RedirectChain = Text
+  fromList = RedirectChain . reverse
+
+pushRequest :: RedirectChain -> Text -> RedirectChain
+pushRequest (RedirectChain chain) = RedirectChain . (: chain)
```
YuriRomanowski, line 85: In all these functions, let's replace `Text` with something more meaningful (either `type` or `newtype`)

src/Xrefcheck/Data/Redirect.hs
```diff
@@ -0,0 +1,162 @@
...
+      -- ^ Try again by following the redirect
+    deriving stock (Show, Eq)
+
+-- | Links in a redirection chain.
+--
+-- They are stored in reverse order, so it is quick to
+-- accumulate new ones while following a redirection chain.
+newtype RedirectChain = RedirectChain
+  { unRedirectChain :: [Text]
+  } deriving newtype (Show, Eq)
```
YuriRomanowski, line 78: How about using `Data.Seq`? We won't need to remember what is reversed and what is not.

src/Xrefcheck/Config.hs
aeqz: Maybe I'm missing something. If we default `ncExternalRefRedirects` to `[]` (in the default config I guess) then, how do we suggest the previously implemented defaults? With a section in the README file?

YuriRomanowski: Hm, maybe we could write it as a comment in the default config, as our "recommendation" :)

src/Xrefcheck/Verify.hs
```diff
@@ -241,20 +246,43 @@ instance Given ColorMode => Buildable VerifyError where
...
+    RedirectChainCycle chain ->
       [int||
-      Permanent redirect found:
-      #{url}
+      Cycle found in the following redirect chain:
+      #{interpolateIndentF 2 $ attachToRedirectChain chain "here"}
       |]
 
-    PermanentRedirectError url (Just redirectedUrl) ->
+    RedirectChainBroken chain ->
```
YuriRomanowski, line 255: Maybe `RedirectMissingLocation`? `RedirectChainBroken` sounds too general, not saying what is broken.

src/Xrefcheck/Verify.hs
```diff
@@ -241,20 +246,43 @@ instance Given ColorMode => Buildable VerifyError where
...
+      #{redirect} found:
+      #{interpolateIndentF 2 $ attachToRedirectChain chain "stopped before this one"}
+      |]
+      where
+        redirect :: Text
+        redirect = case mOn of
+          Nothing -> "Redirect"
+          Just RROPermanent -> "Permanent redirect"
+          Just RROTemporary -> "Temporary redirect"
+          Just (RROCode code) -> show code <> "redirect"
```
YuriRomanowski, line 278: Don't we need a space between the code and "redirect"? `show code <> " redirect"`

src/Xrefcheck/Config.hs
```diff
@@ -113,11 +195,17 @@ overrideConfig config
...
+        , ncMaxRedirectFollows        = overrideField ncMaxRedirectFollows
+        , ncExternalRefRedirects      = externalRefRedirects
         }
       where
         overrideField :: (forall f. NetworkingConfig' f -> Field f a) -> a
         overrideField field = fromMaybe (field defNetworking) $ field networkingConfig
 
+        externalRefRedirects :: RedirectConfig
+        externalRefRedirects = maybe defRedirectConfig ncExternalRefRedirects $
+          cNetworking config
```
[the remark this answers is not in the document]
aeqz: In #218 we detected that GitLab also does a 302 redirect when trying to access a private resource while being not authenticated and, in this case, we want to treat it as valid.

Is there any way to distinguish between these two cases with our rules?

src/Xrefcheck/Verify.hs
```diff
@@ -444,7 +472,7 @@ verifyReference
             let shownFilepath = dropWhile isPathSeparator (toString rLink)
             canonicalPath <- riRoot </ shownFilepath
             checkRef rAnchor riRoot canonicalPath shownFilepath
-          RIExternal -> checkExternalResource config rLink
+          RIExternal -> checkExternalResource (fromList mempty) config rLink
```
YuriRomanowski, line 475: How about defining `emptyChain :: RedirectChain`?

src/Xrefcheck/Verify.hs
aeqz: True! I have no Golden tests with these messages because they would require to also have a mock server to replicate very specific redirect cases.

src/Xrefcheck/Verify.hs
```diff
@@ -699,15 +722,18 @@ checkExternalResource Config{..} link
...
 
-    isPermanentRedirectCode :: Int -> Bool
-    isPermanentRedirectCode = flip elem [301, 308]
+      case reqRes of
+        Nothing -> throwError $ ExternalResourceSomeError "Response timeout"
+        Just RRDone -> pure ()
+        Just (RRFollow nextLink) -> do
+          VerifyResult errs <- liftIO $
+            checkExternalResource (followed `pushRequest` link) config nextLink
+          mapM_ throwError errs
```
YuriRomanowski, line 736: :trollface: What the hell is going on here? Don't we just throw the first error and ignore all others?

tests/Test/Xrefcheck/RedirectChainSpec.hs
```diff
@@ -0,0 +1,103 @@
...
+    , testCase "Cycle" $
+        checkLinkAndProgressWithServer
+        (config 5)
+        mockRedirect
+        (link "/cycle1")
+        progress
+        (VerifyResult [RedirectChainCycle $ chain ["/cycle1", "/cycle2", "/cycle3", "/cycle4", "/cycle2"]])
+    , testGroup "Limit"
+      [ testCase "Takes effect" $
+        checkLinkAndProgressWithServer
```
YuriRomanowski, line 41: Please add indentation for this test case

YuriRomanowski:
Thank you @aeqz, that's cool. Suggested just minor style and naming changes.

src/Xrefcheck/Data/Redirect.hs
aeqz: Cool! I will give it a go.

src/Xrefcheck/Verify.hs
aeqz: Now that the underlying type will be just `Data.Seq`, I will derive `Monoid` instead. I did't derive it because of the current reversed list representation.

src/Xrefcheck/Verify.hs
aeqz: That is right, but `checkExternalResource` returns `VerifyResult VerifyError` (i.e. a list of `VerifyError`) and this is `ExceptT VerifyError IO ()`. I also saw that `checkExternalResource` does not produce ever more than one error.

In the future, if `checkExternalResource` produces more than one error, some of them may be missed when following redirects. Maybe we can just change `checkExternalResource` to return `Maybe VerifyError`, or this `ExceptT VerifyError IO ()` to be `ExceptT (VerifyResult VerifyError) IO ()` instead.

What do you think? If I'm not wrong, you were also refactoring these error types and monads in some of your PRs.

src/Xrefcheck/Verify.hs
YuriRomanowski: Hm, I don't like this idea, because we only want to append links, not join two link chains, so maybe we don't need the monoid instance.

src/Xrefcheck/Verify.hs
aeqz: Also, I think that this would be surely changed in #260, an issue that has been created due to a related discussion in this same PR.

YuriRomanowski: I think for now it's OK to just throw the first error, in [234](https://github.com/serokell/xrefcheck/pull/234) all the types are changed such that only one error was possible, and that pull is ready to merge.

src/Xrefcheck/Config.hs
Martoon-00: Under "using `[]` as default" I understood - if `ncExternalRefRedirects` is missing in the config, parse it as if it was `[]` (it really required clarification). And our recommendation for this field will be in the default config.

src/Xrefcheck/Verify.hs
aeqz: Let's go for `emptyChain` then!

src/Xrefcheck/Config/Default.hs
aeqz: Previously, redirects were automatically followed by the HTTP client, so none of our behaviours were applied to each step, but to the entire chain. For example:
- Timeout was applied to the entire chain.
- Retry logic was applied to the entire chain.
- Ignore check was only applied to the initial link.

This new feature allows to explicitly follow redirects, and we have decided to apply the same configuration rules at each step of the chain. But we are still applying everything else to the entire chain, and we plan to review it in other issues like #260.

Martoon-00: Yeah. It may sound a bit confusing (due to how obvious it seems), but the behaviour should be clarified.

Probably if we add something like:

> so for instance, exclusion rules would apply to the following links

it would sound more benign, more like a hint about a possible trick rather than an obvious statement, and would still make the exact behaviour clear.

README.md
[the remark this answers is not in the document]
Martoon-00: Ah okay, so we actually match the full string. My conclusion was made just by looking at our regex examples, but looking at this function I think the original intention was to match the full string.

I'd personally leave it as is, and treat those `^` as either mistakes or artifacts of parallel work - something that should be fixed. Does it sound good?

Feel free to do the update in this PR or in a separate ticket/PR.

src/Xrefcheck/Config.hs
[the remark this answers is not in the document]
Martoon-00: Hmhm, right.

Perhaps we should handle 302 redirects not from any GitLab page, but from pages that refer to file, like this:

https://gitlab.com/morley-framework/morley-upgradeable/-/blob/master/.hlint-universum.yam

They have `blob` in their name.

We should check which other pages we should handle. But for instance, GitLab issues do not require special actions, we still get 404 there:

https://gitlab.com/morley-framework/morley-upgradeable/-/issues/9999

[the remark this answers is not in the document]
Martoon-00: This way `/issues/new` page does not fall under our rule.

src/Xrefcheck/Config/Default.hs
```diff
@@ -56,6 +56,35 @@ networking:
...
+  # Maximum number of links that can be followed in a single redirect
+  # chain.
+  #
+  # The link is considered as invalid if the limit is exceeded.
+  maxRedirectFollows: 10
+
+  # Rules to override the redirect behavior for external references that
+  # match, where
+  #   - 'from' is a regular expression for the source link in the redirection.
+  #     Its absence means that every link matches.
```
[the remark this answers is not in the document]
Martoon-00: Fair question. 

Hmm, I thought I had a couple of examples in mind, but thinking about them more I see they are not much justified :thinking: 

Probably let's not add any tickets then unless an actual use case for such filter arises.

README.md
```diff
@@ -141,10 +141,38 @@ There are several ways to fix this:
...
+
+      The first one applies if both of them match.
+
+    * The number of redirects allowed in a single redirect chain is limited and can be configured with the
+      `maxRedirectFollows` parameter, also within `networking`. A number smaller than 0 disables the limit.
+
+2. How does `xrefcheck` handle localhost links?
```
Martoon-00, line 175: Yeah, that's an unfair fight :smile:

README.md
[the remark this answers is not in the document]
aeqz: Sure! I will apply this simplification with a commit in this same PR.

src/Xrefcheck/Config.hs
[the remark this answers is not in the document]
aeqz: I have added a default rule for the GitLab flavour and tried it with the link from your example 👍 

If we notice other GitLab redirect details that worth a redirect rule, we can add it.

src/Xrefcheck/Config/Default.hs
```diff
@@ -96,3 +123,20 @@ scanners:
...
+      GitLab ->
+        [int||
+        - on: permanent
+          outcome: invalid
+        # GitLab redirects non-existing files to the repository's main page
+        # with a 302 code instead of answering with a 404 response.
+        - from: https?://gitlab.com/.*/-/blob/.*
+          to: https?://gitlab.com/.*
+          on: 302
+          outcome: invalid|]
```
Martoon-00, line 142: Very good :+1: 

The fact that those regexes are internally glob patterns works for us: somehow that `.*` in between does not match things that contain slashes, and this is exactly what we what.

README.md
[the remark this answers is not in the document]
Martoon-00: Looks very good now, awesome.

src/Xrefcheck/Config.hs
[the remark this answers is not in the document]
Martoon-00: Sounds good :+1:

src/Xrefcheck/Config.hs
```diff
@@ -95,6 +105,7 @@ overrideConfig config
     defScanners = cScanners $ defConfig flavor
     defExclusions = cExclusions $ defConfig flavor
     defNetworking = cNetworking $ defConfig flavor
+    defRedirectConfig = []
```
Martoon-00, line 108: Oh, I now see why my proposal was not so clear, till this moment we defaulted fields to the values from the default config.

tests/configs/github-config.yaml
```diff
@@ -45,6 +45,33 @@ networking:
...
+  #     Its absence also means that every response code matches.
+  #   - 'outcome' accepts 'valid', 'invalid' or 'follow'. The last one follows
+  #      the redirect by applying the same configuration rules so, for instance,
+  #      exclusion rules would also apply to the following links.
+  #
+  # The first one that matches is applied, and the link is considered
+  # as valid if none of them does match.
+  #
+  # If a value is provided for 'networking' but not for 'externalRefRedirects',
+  # then it will default to an empty list of rules and every redirect will pass.
```
Martoon-00, line 70: This seems to be error-prone (if I got it correctly), this field should default to `[]` even if `networking` is not provided.

Maybe there is a simple way to make it work so?

YuriRomanowski (approved):

src/Xrefcheck/Config.hs
aeqz: Yeah! I though that it should still being that way, but it is clear now.

tests/configs/github-config.yaml
aeqz: Done!

aeqz:
I have squashed fixup and temporary commits for the ease of rebasing this PR branch.

src/Xrefcheck/Config.hs
```diff
@@ -118,7 +128,8 @@ overrideConfig config
 
     defScanners = cScanners $ defConfig flavor
     defExclusions = cExclusions $ defConfig flavor
-    defNetworking = cNetworking $ defConfig flavor
+    defNetworking = cNetworking (defConfig flavor)
+      & ncExternalRefRedirectsL .~ []
```
Martoon-00, line 132: Nice :+1:

src/Xrefcheck/Config/Default.hs
```diff
@@ -65,22 +66,45 @@ networking:
...
+  # and so xrefcheck considers timeouts in the same way as 429 errors.
+  #
+  # For other domains, a timeout results in a respective error, no retry
+  # attempts will be performed. Use `externalRefCheckTimeout` option
+  # to increase the time after which timeout is declared.
+  #
+  # This option is similar to `maxRetries`, the difference is that
+  # this `maxTimeoutRetries` option limits only the number of retries
+  # caused by timeouts, and `maxRetries` limits the number of retries
+  # caused both by 429s and timeouts.
```
Martoon-00, line 82: :+1:

Martoon-00 (approved):
LGTM, thanks for your work :+1: 

The last comment I have is that the second commit has some changes that seem to be unrelated to it, this may be worth a cleanup. But other than that, good, feel free to merge after fixing that.

Issue #219, opened 2022-11-17
Optimize performance in case of local links


Martoon-00:
# Clarification and motivation

One use case that I wanted to see very fast - running xrefcheck on local links only. So that as a user, I could have a very responsive "Verified >> Edited" loop.

However, currently it seems to take quite long.

If I run the current `xrefcheck` on Morley, then it takes 0.75s in average.  0.13s of those turned out to be repo scanning. So most part of the time is taken in verification

# Acceptance criteria

* We investigated what are the bottlenecks (probably via profiling), and tried to optimize them.
* We use concurrency where appropriate.

Martoon-00:
Note: as far as I understand it correctly, making IO concurrent does not make much sense. But if pure logic takes a decent amount of time, it should be parallelized.

Martoon-00:
I tried `xrefcheck` on Morley, and it takes 0.75s to check the local links.

0.13s turned out to be consumed for files traversal, 0.08s on scanning, so looks like the most portion of time is consumed by verification. And I'm not sure how can this be, it is mostly pure code if we exclude external links checks :thinking:

Sorokin-Anton:
IMHO, this proportion of time may be result of laziness (so it reported about finishing scan before e.g. all anchors were parsed)

Martoon-00:
Ah yes, I first measured it incorrectly but then remembered the need in running some `evaluateNF` after the scan. Otherwise, that result could be completely spoiled, that's true.

Martoon-00:
I tried profiling this in a free time, and if I interpret the results correctly, we really have something to optimize in the verification logic. I have run on Morley repository with `-m local-only`.


![morley-run](https://user-images.githubusercontent.com/5394217/202824998-b60cc532-6329-4680-b49f-0d89ff73bbb4.svg)

(I also had `evaluateNF` applied at the end of the scan logic to exclude parsing from being computed as part of verification).

I get approximately the same flamegraph if focusing on memory, not time.

I believe, something like this is fully expected, given that we brought proper handling of filepaths comparison just recently.

---

Looks like in both fat places, `equalFilePath` takes all the time. 

`equalFilePath` on all systems is approximately `(==) 'on' normalize`, and I believe that if we extract normalization out and maybe keep those in `Set`s, this will result in a dramatic time save.

YuriRomanowski:
Also we can try to use `Text` instead of `String` if it eventually boils down to simple `(==)`

Martoon-00:
[...]

PR #262, opened 2022-12-30
[#244] Symlink scanner

aeqz:
## Description

Problem: as GitHub and GitLab do not render symlinks as the file they point to, and we have disabled the markdown scanner to process them in #261, we are also considering to implement a new scanner for symlinks that verifies them up to some extent.

Solution: in this proposal, a scanner that validates the reference from a symlink has been implemented in the same style as the markdown scanner.

## Related issue(s)

Fixes #244 

## :white_check_mark: Checklist for your Pull Request

Ideally a PR has all of the checkmarks set.

If something in this list is irrelevant to your PR, you should still set this
checkmark indicating that you are sure it is dealt with (be that by irrelevance).

#### Related changes (conditional)

- Tests
  - [x] If I added new functionality, I added tests covering it.
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [x] I checked whether I should update the docs and did so if necessary:
    - [README](https://github.com/serokell/xrefcheck/tree/master/README.md)
    - Haddock

- Public contracts
  - [x] Any modifications of public contracts comply with the [Evolution
  of Public Contracts](https://www.notion.so/serokell/Evolution-of-Public-Contracts-2a3bf7971abe4806a24f63c84e7076c5) policy.
  - [x] I added an entry to the [changelog](https://github.com/serokell/xrefcheck/tree/master/CHANGES.md) if my changes are visible to the users
        and
  - [x] provided a migration guide for breaking changes if possible

#### Stylistic guide (mandatory)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [x] My code complies with the [style guide](https://github.com/serokell/style/blob/master/haskell.md).

Issue #235, opened 2022-12-10
Refactor the interface to progress bar updates


Martoon-00:
# Clarification and motivation

For providing a progress bar interface we provide functions: `incProgress`, `incProgressUnfixableErrors`, `decProgressFixableErrors`, but they constitute quite a low-level interface. If I use `incProgressFixableErrors`, it should be accompanied by `incProgress`, forgetting it is likely an error.

`fixableToUnfixable` is better in this regard.

Let's discuss in comments how to better improve this situation.

# Acceptance criteria

* We provide a convenient high-level interface to changing progress bar state that is not error-prone to a reasonable extent.

Martoon-00:
<details>
  <summary>My proposal</summary>
  
I think what humans expect here is (just illustrating idea, details are irrelevant):

```hs
data ProgressItemStatus = SUnchecked | SOk | SFixable | SFailed

(->) :: ProgressItem -> ProgressItem -> Progress a -> Progress a
```

```hs
-- example 
let progress' = progress & SUnchecked -> SOk
```

This way, intention behind some of the updates may be expressed more clearly:

```hs
progressUpdateOnRetry =
  (if retryCounter == 0 then SUnchecked else SFixable)
  ->
  (if retryCounter < maxRetries then SFixable else SFailed)
``` 

</details>

Martoon-00:
<details>
  <summary>Another, more complex but better proposal</summary>

With both the current solution and the improvement I suggested above there is an issue - we have to figure out the old status correctly for things to work. And this is pretty error-prone, not saying that it may complicate future features (e.g. if we decide to track redirects too).

So instead we could internally keep `Map Link ProgressStatus` and each link will overwrite its own status.
Maybe instead of using `Link` as key (I'm not sure how we currently reflect duplicated links in progress bar) there should be a unique token instead.

</details>

PR #123, opened 2022-07-22
[Chore] Add missing changelog entries

dcastro:
## Description

Some changelog entries were missing from recent PRs. This PR adds those missing entries.

## Related issue(s)

Fixed none

## :white_check_mark: Checklist for your Pull Request

Ideally a PR has all of the checkmarks set.

If something in this list is irrelevant to your PR, you should still set this
checkmark indicating that you are sure it is dealt with (be that by irrelevance).

#### Related changes (conditional)

- Tests
  - [x] If I added new functionality, I added tests covering it.
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [x] I checked whether I should update the docs and did so if necessary:
    - [README](/README.md)
    - Haddock

- Public contracts
  - [x] Any modifications of public contracts comply with the [Evolution
  of Public Contracts](https://www.notion.so/serokell/Evolution-of-Public-Contracts-2a3bf7971abe4806a24f63c84e7076c5) policy.
  - [x] I added an entry to the [changelog](/CHANGES.md) if my changes are visible to the users
        and
  - [x] provided a migration guide for breaking changes if possible

#### Stylistic guide (mandatory)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [x] My code complies with the [style guide](/docs/code-style.md).

PR #263, opened 2023-01-09
[#239][#249] Further filepath refactor

### LINKED ISSUE #239: Make canonical paths relative to the repository root

After the #230 refactor, we will be using a custom `FilePath` canonicalization that is weaker than `System.Directory.canonicalizePath`. It currently applies a `FilePath.makeAbsolute` transformation and we can try to avoid it in order to reduce some filesystem state dependency.

# Acceptance criteria

- If possible, replace the `FilePath.makeAbsolute` transformation in our `canonicalizePath` function by another transformation that allows us to hold paths relative to the repository root instead of absolute ones.

### LINKED ISSUE #249: Try to exploit newtypes over FilePath more to better separate semantics

After #230 we still have some ambiguities around `FilePath`, e.g. it may be always in Posix format or platform dependent.

I first abandoned this idea, but it starts getting more desired to me just as a measure to make the code more readable. Namely: let's introduce some `MarkdownFilePath` or `PosixFilePath` that would be used in markdown links. And leave `FilePath` for platform-dependent version of path.

This way one can easily understand the expected format at any place in the code just by looking at the type.

We may end up using posix-formatted paths in at least two places after #230 - in `FileInfo` and in error type, so having a separate type starts to make sense for me.

On implementation notes:
* We will likely need to start separating local links and remote links in `FileInfo`, let's probably do so.

# Acceptance criteria

* We try to introduce the mentioned newtype.
* Looking at the result, we assess how much sense it makes.

aeqz:
## Description

Problem: after refactoring the `FilePath` usages in the codebase to have a canonical representation of them, we noticed that further improvements could be applied, such as clarifying whether the path is system dependent and avoiding absolute file system paths.

Solution: we now use POSIX relative paths during the analysis, and system dependant ones for reading file contents in the scan phase.

## Related issue(s)

Fixes #239
Fixes #249
Relates #219

#### Related changes

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

#### Stylistic guide


src/Xrefcheck/System.hs
```diff
@@ -50,121 +41,99 @@ askWithinCI = lookupEnv "CI" <&> \case
...
+-- considered as relative to a given 'FilePath'.
+filePathFromRoot :: FilePath -> RelPosixLink -> FilePath
+filePathFromRoot rootPath = (rootPath FP.</>)
+  . withPathSeparator FP.pathSeparator
+  . toString
+  . unRelPosixLink
+
+-- | Join two 'RelPosixLink's.
+(</>) :: RelPosixLink -> RelPosixLink -> RelPosixLink
+UnsafeRelPosixLink a </> UnsafeRelPosixLink b = mkRelPosixLink
```
Martoon-00, line 80: I have found some trailing whitespaces here:
```suggestion:-0+0
UnsafeRelPosixLink a </> UnsafeRelPosixLink b = mkRelPosixLink
```




</p>

src/Xrefcheck/System.hs
```diff
@@ -50,121 +41,99 @@ askWithinCI = lookupEnv "CI" <&> \case
...
+takeDirectory :: RelPosixLink -> RelPosixLink
+takeDirectory = UnsafeRelPosixLink
+  . fromString
+  . FPP.takeDirectory
+  . toString
+  . unRelPosixLink
+
+-- | 'FilePath.takeExtension' version for 'RelPosixLink'.
+takeExtension :: RelPosixLink -> String
+takeExtension = FPP.takeExtension
```
Martoon-00, line 93: I have found some trailing whitespaces here:
```suggestion:-0+0
takeExtension = FPP.takeExtension
```




</p>

src/Xrefcheck/System.hs
```diff
@@ -50,121 +41,99 @@ askWithinCI = lookupEnv "CI" <&> \case
...
+takeDirectory = UnsafeRelPosixLink
+  . fromString
+  . FPP.takeDirectory
+  . toString
+  . unRelPosixLink
+
+-- | 'FilePath.takeExtension' version for 'RelPosixLink'.
+takeExtension :: RelPosixLink -> String
+takeExtension = FPP.takeExtension 
+  . toString
```
Martoon-00, line 94: I have found some trailing whitespaces here:
```suggestion:-0+0
  . toString
```




</p>

src/Xrefcheck/System.hs
```diff
@@ -50,121 +41,99 @@ askWithinCI = lookupEnv "CI" <&> \case
...
+takeExtension :: RelPosixLink -> String
+takeExtension = FPP.takeExtension 
+  . toString 
+  . unRelPosixLink
+
+-- | Get the list of directories between a 'RelPosixLink' and its
+-- relative root.
+getIntermediateDirs :: RelPosixLink -> [RelPosixLink]
+getIntermediateDirs link = fmap UnsafeRelPosixLink $
+  case T.splitOn "/" $ unRelPosixLink $ takeDirectory link of
```
Martoon-00, line 101: I have found some trailing whitespaces here:
```suggestion:-0+0
  case T.splitOn "/" $ unRelPosixLink $ takeDirectory link of
```




</p>

src/Xrefcheck/System.hs
```diff
@@ -50,121 +41,99 @@ askWithinCI = lookupEnv "CI" <&> \case
...
+  case T.splitOn "/" $ unRelPosixLink $ takeDirectory link of 
+    [] -> []
+    ["."] -> [""]
+    [".."] -> ["", ".."]
+    d : ds -> scanl (\a b -> a <> "/" <> b) d ds
+
+-- | Check if a 'RelPosixLink' passes through its relative root when
+-- expanding indirections.
+hasUnexpanededParentIndirections :: RelPosixLink -> Bool
+hasUnexpanededParentIndirections (UnsafeRelPosixLink link) =
```
Martoon-00, line 110: I have found some trailing whitespaces here:
```suggestion:-0+0
hasUnexpanededParentIndirections (UnsafeRelPosixLink link) =
```




</p>

tests/Test/Xrefcheck/RelPosixLinkSpec.hs
```diff
@@ -0,0 +1,49 @@
+{- SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io>
```
Martoon-00, line 1: :warning: The year in this license header is outdated, time to update!





</p>

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Trailing whitespaces and/or incorrect end-of-file newlines detected.</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          4 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">In 65653be7e2b1cdd9f5fa1d13c7b656e5257e1bdf message lacks issue id: <code>[#239-#249] Further filepath refactor</code>.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">In 65653be7e2b1cdd9f5fa1d13c7b656e5257e1bdf subject does not begin with an uppercase letter: <code>[#239-#249] Further filepath refactor</code>.<br />
See also [Note].</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Inappropriate title for PR.<br />
Should start from issue ID (e.g. <code>[#123]</code>), <code>[Chore]</code> or <code>[Style]</code> tag.<br />
Note: please use <code>[Chore]</code> also for tickets tracked internally on YouTrack.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please, only use letters, digits and dashes in the branch name.<br />
        Found: ["#"]</td>
    </tr>
  </tbody>
</table>

</p>

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">WIP commits are still there.</td>
    </tr>
  </tbody>
</table>

</p>

src/Xrefcheck/Core.hs
```diff
@@ -71,66 +70,58 @@ instance Given ColorMode => Buildable Position where
...
+    -- ^ Reference to a file at outer site, e.g @[d](http://www.google.com/doodles)@.
+  | ELOther Text
+    -- ^ Entry not to be processed, e.g. @mailto:e-mail@.
+  deriving stock (Show, Generic)
+
+data FileLink
+  = FLAbsolute Text
+    -- ^ Reference to a file relative to the repository root.
+  | FLRelative Text
+    -- ^ Reference to a file relative to the given one.
```
aeqz, line 105: These two `Text` types could be changed also to `RelPosixLink`, but then the `Reference` `Buildable`  instance would show the link as normalised instead of as it is written in the Markdown file.

Would that be fine? Or it is preferable to keep the raw link here as it is.

src/Xrefcheck/Core.hs
```diff
@@ -198,51 +189,73 @@ data DirectoryStatus
...
+        reference #{paren $ colorIfNeeded Red "external"} #{rPos}:
+          - text: #s{rName}
+          - link: #{if null url then "-" else url}
+        |]
+      RIExternal (ELOther url) ->
+        [int||
+        reference (other) #{rPos}:
+          - text: #s{rName}
+          - link: #{if null url then "-" else url}
+        |]
```
aeqz, line 259: Here I have removed unnecessary fields from reference types. For example:
- Link in local references.
- Anchor in external references. It is now part of the URL.

Do you agree with this output change?

aeqz:
How should I properly write the PR and commit issue id referring to two different issues?

I have tackled both of them at the same time because, when trying to refine when POSIX links are used and when platform dependent ones, I realised that we just need relative POSIX links during the verification phase (either relative to the repository root or to other repository file) and system dependent ones during the scan phase, which do not require further processing.

Martoon-00:
> I have tackled both of them at the same time because

Yeah, sounds fair.

Usually we write e.g. `[#123][#456] Text` in PR title or commit subject if it concerns multiple tickets.

src/Xrefcheck/Core.hs
Martoon-00: That's really interesting.

Strictly speaking, external links also can have anchors, but we do not account for them at all, so not displaying them seems reasonable. I would say, it is better than before.

And for link in local references - I also agree.

Nice that parsing the link into `ReferenceInfo` let us for such fine-tuned pretty-printing and stuff.

src/Xrefcheck/Core.hs
```diff
@@ -71,66 +70,58 @@ instance Given ColorMode => Buildable Position where
...
+data ExternalLink
+  = ELUrl Text
+    -- ^ Reference to a file at outer site, e.g @[d](http://www.google.com/doodles)@.
+  | ELOther Text
+    -- ^ Entry not to be processed, e.g. @mailto:e-mail@.
+  deriving stock (Show, Generic)
+
+data FileLink
+  = FLAbsolute Text
+    -- ^ Reference to a file relative to the repository root.
```
Martoon-00, line 103: Note that links to directories are also valid, so this naming may be misleading.

I cannot imagine a better naming myself, but at least a haddock description seems mandatory here.

src/Xrefcheck/System.hs
```diff
@@ -50,121 +41,99 @@ askWithinCI = lookupEnv "CI" <&> \case
...
-    expand acc (".." : xs) = expand (acc + 1) xs
-    expand acc ("." : xs) = expand acc xs
-    expand 0 (x : xs) = x : expand 0 xs
-    expand acc (_ : xs) = expand (acc - 1) xs
-    expand acc [] = replicate acc ".."
+-- It should be created from a system dependent 'FilePath' via
+-- 'mkRelPosixLink'.
+newtype RelPosixLink = UnsafeRelPosixLink
+  { unRelPosixLink :: Text
+  } deriving newtype (Show, Eq, Ord, NFData, Buildable, Pretty)
```
Martoon-00, line 50: Btw, I would put it clearly somewhere in the code, that `FilePath` is used for platform-dependent filepath.

Maybe in haddock to `RelPosixPath`, for instance as "This type exists in contrast to `FilePath` which is used ..."

src/Xrefcheck/Core.hs
```diff
@@ -363,11 +376,15 @@ data VerifyProgress = VerifyProgress
...
+  { vrLocal = initProgress . length . filter isNothing $ extLinks
+  , vrExternal = initProgress . length . ordNub . catMaybes $ extLinks
   }
   where
-    (extRefs, localRefs) = L.partition (isExternal . rInfo) references
+    extLinks = flip fmap references $ \ref ->
+      case rInfo ref of
+        RIExternal (ELUrl url) ->
+          Just url
+        _ -> Nothing
```
Martoon-00, line 387: A challenge: will you be able to write the same using prisms from `lens` package?

You can autogenerate prisms using `makePrisms`, and then use those lens to filter differently.

Also, I suppose `ELOther` should not be included into "local" part (in all fairness, previously it was included into "external" part which seems to be also invalid). Wonder if #235 it would be possible (and viable) to force the initialization of progress bar to be also not so error-prone.

tests/Test/Xrefcheck/IgnoreAnnotationsSpec.hs
```diff
@@ -38,28 +39,30 @@ test_ignoreAnnotations =
...
       ]
   , testGroup "\"ignore paragraph\" mode"
       [ testCase "Check \"ignore paragraph\" performance" $ do
-         (fi, errs) <- parse GitHub "tests/markdowns/with-annotations/ignore_paragraph.md"
-         getRefs fi @?= ["blog", "contacts"]
-         errs @?= []
+          let file = mkRelPosixLink "tests/markdowns/with-annotations/ignore_paragraph.md"
+          (fi, errs) <- parse GitHub "" file
+          getRefs fi @?= ["blog", "contacts"]
+          errs @?= []
```
Martoon-00, line 53: Neat to see this better aligned :+1:

Martoon-00 (commented):
I reviewed superficially for now, and quite like how it looks like.

And otherall to me this looks simpler and more robust now.

@YuriRomanowski I would also appreciate your review, if you find time; meanwhile I'll try to spend more time on your PRs.

src/Xrefcheck/Core.hs
aeqz: It is also a win if we go on with the symlink scanner (#262). With this changes, it would be easier to model and display them.

src/Xrefcheck/Core.hs
aeqz: Cool, I will try to do so!

> previously it was included into "external" part

I think that previously it was also included in the local part. `isExternal` `RIOtherProtocol` was defined to be `False`.

src/Xrefcheck/Core.hs
```diff
@@ -363,11 +379,11 @@ data VerifyProgress = VerifyProgress
 
 initVerifyProgress :: [Reference] -> VerifyProgress
 initVerifyProgress references = VerifyProgress
-  { vrLocal = initProgress (length localRefs)
-  , vrExternal = initProgress (length (ordNub $ map rLink extRefs))
+  { vrLocal = initProgress . length $
```
Martoon-00, line 382: I have found some trailing whitespaces here:
```suggestion:-0+0
  { vrLocal = initProgress . length $
```




</p>

src/Xrefcheck/Core.hs
aeqz: I have modified this to not include `ELOther` in the progress bar, which also required to modify `modifyProgressRef` because it was also updating progress corresponding to `ELOther` references in the local progress bar.

src/Xrefcheck/Core.hs
aeqz: The original one, for displaying, would also be with POSIX path separators although not normalised? Otherwise, as the original one could be platform-dependant, the program output would also be platform-dependant.

aeqz: I found some trouble when trying to keep both the canonical and the original path because the resulting code was not so neat. I tried then a different approach: keep the original path and implement "normalised" `Eq` and `Ord` instances but reached a similar situation.

Finally, I have tried to define a `RelPosixLink` that is just a relative POSIX link, with no normalisation, and a `CanonicalRelPosixLink` which is a `RelPosixLink` with normalisation applied. I will leave some comments here regarding the result.

src/Xrefcheck/System.hs
```diff
@@ -174,27 +171,53 @@ mkGlobPattern path = do
...
+      . expandPosixIndirections
+      . fromString
+      . FPP.normalise
+      . FPP.dropTrailingPathSeparator
+
+-- Checks if a 'CanonicalRelPosixLink' matches some of the given
+-- 'CanonicalRelGlobPattern's.
+--
+-- They are considered as relative to the same root.
+matchesGlobPatterns :: [CanonicalRelGlobPattern] -> CanonicalRelPosixLink -> Bool
```
aeqz, line 185: Glob patterns are canonicalised and matched against canonicalised links.

I tried to not canonicalise glob patterns or to not canonicalise the links that we match for, but we have test cases that require either one or the other thing.

src/Xrefcheck/System.hs
Martoon-00: Looks very good now, thanks :+1:

src/Xrefcheck/Core.hs
```diff
@@ -379,12 +379,11 @@ data VerifyProgress = VerifyProgress
 
 initVerifyProgress :: [Reference] -> VerifyProgress
 initVerifyProgress references = VerifyProgress
-  { vrLocal = initProgress $ length references - length mExternalUrls
-  , vrExternal = initProgress . length . ordNub $ mExternalUrls
+  { vrLocal = initProgress . length $ 
+      references ^.. folded.to rInfo._RIFile
+  , vrExternal = initProgress . length . ordNub $ 
+      references ^.. folded.to rInfo._RIExternal._ELUrl
```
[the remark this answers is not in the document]
aeqz: Yeah! The [lens package readme](https://github.com/ekmett/lens) uses that OOP-like syntax, for the examples, so I thought that it was a common style when using lenses. I will turn back to FP then.

aeqz:
> Okay, I guess it is now better than before. Although if I got from the beginning that everything is consistent in this regard, I would probably abort my suggestion to change the existing code.

I will undo that change. As you say, it was already consistent.

Now I wonder if counting "other" references as local progress could be better initialised by using lenses:

```haskell
vrLocal = initProgress $
      (length $ references ^.. folded . to rInfo . _RIFile) +
      (length $ references ^.. folded . to rInfo . _RIExternal . _ELOther)
```

This is not valid, but something similar would be cool instead:

```haskell
vrLocal = initProgress $ length $ 
  references ^.. folded . to rInfo . (_RIFile <|> _RIExternal . _ELOther)
```

src/Xrefcheck/Core.hs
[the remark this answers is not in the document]
Martoon-00: Oh I see, that's kinda surprising to me :thinking: 

Yeah, let's use spaces, at least to be consistent with other code in Serokell.

Martoon-00:
> I will undo that change. As you say, it was already consistent.

Thanks :pray: 

> This is not valid, but something similar would be cool instead:

Oh, I cannot miss this call for perfection!

AFAIR internally `Traversal`s are about lists, not `Maybe`s, so you should try `<>` instead of `<|>`.

Plus `_RIFile` and `_RIExternal . _ELOther` may return different types, so to lead everything to one type I would use `united` lens.

This works for me:

```hs
> let vals = [ Left "a", Right (Left 0), Right (Right 1) ] :: [Either String (Either Word Int)]
> vals ^.. L.folded . (L._Left . L.united <> L._Right . L._Right . L.united)
[(),()]
```

src/Xrefcheck/System.hs
```diff
@@ -41,50 +44,42 @@ askWithinCI = lookupEnv "CI" <&> \case
...
---  * FilePath.normalise, which is platform-dependent.
---
---  * Expand '.' and '..' indirections syntactically.
---
-mkRelPosixLink :: FilePath -> RelPosixLink
+-- | Create a POSIX file path from a platform-dependent one.
+mkRelPosixLink :: Text -> RelPosixLink
 mkRelPosixLink = UnsafeRelPosixLink
-  . fromString
   . withPathSeparator FPP.pathSeparator
```
Martoon-00, line 62: Btw, do we check somewhere that markdown links should not use backslash as path separator?

src/Xrefcheck/Core.hs
Martoon-00: Okay, after looking at your code I suppose I now understand where my initial intention (including both fields into `RelPosixLink` and use that as a key) would cause problems.

I guess what you got is really the only sensible way.

src/Xrefcheck/System.hs
Martoon-00: Could you elaborate what problems we get if glob is not normalized?

src/Xrefcheck/System.hs
aeqz: We are not currently checking it but, just in case, here `\` is replaced by `/`. Should we check it and consider backslash as path separator usages invalid?

aeqz:
> AFAIR internally `Traversals` are about lists, not `Maybes`, so you should try `<>` instead of `<|>`

Nice! I was also trying to unify types, but to `Either` instead of `()`, and also thinking that this sum type getters are "partial", so that's why my intuition was going more like "try with this getter that goes to `Maybe (Left a)` or, if it fails, this other one that goes to `Maybe (Right b)`. But, as you pointed out, `^..` is about lists, not maybes.

Your proposal is simpler and worked perfectly 👌

src/Xrefcheck/System.hs
aeqz: Yeah! I have found the test cases that fail.

### If we do not normalize Glob patterns

In `check-local-refs/Checking behavior when there are virtual files, root = "dir1"`, the following links are supposed to be ignored by config:

A) `dir2/../DIR2` (canonicalized is DIR2)
B) `b/../../b.md` (canonicalized is ../b.md)

Config indicates if with the following patterns:

`DIR2/` (this will match A if we remove trailing whitespaces in the pattern)
`b/../../*` (this will match B if we expand indirections in the pattern)

### If we do not normalize neither Glob patterns nor the paths that we try to match

Given the previous case, we can think about trying to match paths as they are, but then other test case fails

In `check-ignoreLocalRefsTo/IgnoreLocalRefsTo: all references should be valid`, the following link is supposed to be ignored by config:

C) `one/../two/b.md`

Config indicates it with the following pattern:

`./two/*` (this will match C if indirections are expanded both in the pattern and in C itself)

So, our test cases suggest that we expect glob matching to be normalized in both sides: patterns and links.

src/Xrefcheck/Core.hs
aeqz: > Wonder if https://github.com/serokell/xrefcheck/issues/235 it would be possible (and viable) to force the initialization of progress bar to be also not so error-prone.

Regarding this, I have added an initialization function that accepts a list of "progress unit witnesses", so it is currently a list of `()` for the local part and a list of `Text` for the external one.

src/Xrefcheck/System.hs
Martoon-00: I think yes, since GitHub and GitLab think so.

So we need two versions of `mkRelPosixLink` - one that accepts `FilePath` and does the replacement, and another one that accepts `Text` and fails on `\` encounters.

---

One other but related thing I wonder is - under Windows, will we get normal filepaths or they will be lower-cased? :joy:  Hope the tests will show us.

src/Xrefcheck/Verify.hs
```diff
@@ -506,17 +506,17 @@ verifyReference
...
-      case res of
+      case (rInfo, res) of
         -- Success
-        Right () -> do
-          modifyProgressRef Nothing $ reportSuccess rLink
+        (RIExternal (ELUrl url), Right ()) -> do
+          modifyExternalProgress Nothing $ reportSuccess url
           pure res
-        Left err -> do
+        (RIExternal (ELUrl url), Left err) -> do
```
Martoon-00, line 514: Hm, such complex pattern match that checks `2x2` cases + one of the cases is a nested thing - seems dangerous to me. If one day we need to treat another constructor as external link, bugs or code bloat will likely be introduced.

I would handle the difference between `()` and `Text` witnesses in some other way.

Maybe leave `modifyProgressRef`, but make it accept `(w -> Progress Int w -> Progress Int w)` function? As before, let in pattern match on `rInfo` inside but also make it supply `()` / `url` argument, not expect its caller to do so?

Martoon-00 (commented):
One more comment

src/Xrefcheck/Verify.hs
aeqz: Good point. It can be kept as before by using a `(forall w. w -> Progress Int w -> Progress Int w)` function.

src/Xrefcheck/System.hs
aeqz: I have changed the following:

- The `RelPosixLink` constructor is now exposed and not unsafe to use.
- `mkRelPosixLink` is now for building a `RelPosixLink ` from a `FilePath`.

After trying some approaches to this, I think that we should not forbid `\` characters in `GitHub` and `GitLab` links. Both of them seem to allow files named like `a\a.md`, and references to them work as expected. However, I tried to create a Golden test case for this and encountered some problems:

- When a file name contains some character like `\`, its corresponding line in the `git-ls files` command appears surrounded by double quotes and with that character escaped, and Xrefcheck considers that surrounding double quotes as being part of the file name, so it is invalid and cannot be read. Maybe we could use `String`'s `read` when the line starts with `"`.
- If, with that change, we managed to make Xrefcheck work with these file names, I think that we would have a problem in Windows:

File `a.md`
```
[Ref to a\a](a\a.md)
```

File `a\a.md`
```
[Ref to a.md](a.md)
```

This would be fine in GitHub, GitLab, Linux and macOS, but in Windows that last `a.md` reference would fail because it should be `../a.md` instead. The first one to `a\a.md` would work though.

Do you think that this deserves creating an issue?

src/Xrefcheck/Verify.hs
Martoon-00: While maybe functions like `getTaskTimestamp` can really not accept a witness, I don't see what will break in this case.

src/Xrefcheck/System.hs
Martoon-00: > After trying some approaches to this, I think that we should not forbid \ characters in GitHub and GitLab links. Both of them seem to allow files named like a\a.md, and references to them work as expected

Oh, that true. Allowing `\` is also an option.

Hmm, but still most of the `\` usages are likely mistakes introduced by Windows users. Perhaps to handle this, it would be worth extending the "file not found" error message in case it contains backshashes, say that likely a mistake takes place?

> When a file name contains some character like \, its corresponding line in the git ls-files command appears surrounded by double quotes

Interesting. I believe that this happens not only for `\` and so should generally be fixed. Could you do that in a separate commit? (and feel free to put that commit at the beginning of your PR)

AFAIS [from docs](https://git-scm.com/docs/git-ls-files#_output) we can also pass `-z` option, that would simplify parsing.

> If, with that change, we managed to make Xrefcheck work with these file names, I think that we would have a problem in Windows:

I wanted to say that Markdown's `a\a.md` should then correspond to Window's `a\\a.md` (or something like that) and vice versa, but then found out that `\` in filenames under Windows is not allowed, and including such file under Linux in git then can cause severe problems for Windows users :thinking:  I.e. `a\a.md` file will simplify get rejected by OS on `git clone`.

So maybe we don't even need to consider that case under Windows.

src/Xrefcheck/System.hs
```diff
@@ -51,125 +44,122 @@ askWithinCI = lookupEnv "CI" <&> \case
...
   _ -> False
 
--- | A FilePath that has been canonicalized.
+-- | Relative file path with POSIX path separators.
 --
--- It should be created via 'canonicalizePath'.
+-- This type exist in contrast to 'FilePath' which, in this project,
+-- is used for platform-dependent file paths and related filesystem
+-- IO operations.
+newtype RelPosixLink = RelPosixLink
```
Martoon-00, line 52: If `RelPosixLink` is safe constructor, I would then mention explicitly that non-posix separators can also appear here and that they will serve as mere part of filename. And that this is the logic that GitHub/GitLab follow.

src/Xrefcheck/Verify.hs
aeqz: I made `getTaskTimestamp` and `setTaskTimestamp` accept a witness in #265 because, from the interface point of view, they seem like operations that require a witness, but the current implementation does not really need a witness.

Do you prefer to make them not require a witness? I'm also not completely sure about what this progress bar timestamp means.

src/Xrefcheck/System.hs
aeqz: Done!

> So maybe we don't even need to consider that case under Windows

It sounds good then, but it also means that I cannot add a test case with a `a\a.md` named file. Right?

> Perhaps to handle this, it would be worth extending the "file not found" error message

I think that it is a good idea and a simple thing to add 👍

src/Xrefcheck/Verify.hs
Martoon-00: That time is displayed when rate limiteters asked us to wait.

I'd say that following the interface sanity is really important, however the description of `TaskTimestamp` and `pTaskTimestamp` say that those are timestamps of an anonymous timer task, so actually the semantics also assumes here that the witness is irrelevant, as far as I see it.

src/Xrefcheck/System.hs
Martoon-00: > It sounds good then, but it also means that I cannot add a test case with a a\a.md named file. Right?

Ugh, fair.

Mmm, this seems like a justified case to start using OS-dependent bats tests :thinking:  Like, set up some env variable in CI config, and rely on it in the test?

src/Xrefcheck/Verify.hs
```diff
@@ -154,6 +152,13 @@ data ResponseResult
 
 instance Given ColorMode => Buildable VerifyError where
   build = \case
+    LocalFileDoesNotExist file | '\\' `elem` unRelPosixLink file ->
+      [int||
+      File does not exist:
+        #{file}
+        Its reference contains a backslash. Maybe it uses the wrong path separator.
+      |]
+
```
Martoon-00, line 161: Cannot miss adding a couple of nitpicks )

* As to me, this causes a significant amount of code duplication, I would try to add it differently.
* Also, let's have a dedicated function for `'\\' 'elem' unRelPosixLink file` in `RelPosixLink` utilities, I think this would make certain sense interface-wise.

src/Xrefcheck/Verify.hs
aeqz: Oh, I missed the "anonymous" part. I agree then with the irrelevance of the witness in these functions.

Maybe this is overthinking: from the interface point of view, `getTaskTimestamp` surely does not require a witness because it retrieves a `TaskTimestamp`, which is anonymous, but `setTaskTimestamp` constructs a `TaskTimestamp`, so maybe it can be seen as a point in where a task passes to be anonymous, and then it would be justified to require a witness there.

aeqz: Actually, by following the code, I think that this timestamp is just informative and has nothing to do with progress units, right? I will change both functions then.

src/Xrefcheck/Scan.hs
```diff
@@ -193,83 +192,92 @@ readDirectoryWith mode config scanner root = do
...
-      (shell "git ls-files"){cwd = Just $ unCanonicalPath root} ""
+      (shell "git ls-files"){cwd = Just root} ""
     getUntrackedFiles = readCreateProcess
-      (shell "git ls-files --others --exclude-standard"){cwd = Just $ unCanonicalPath root} ""
+      (shell "git ls-files --others --exclude-standard"){cwd = Just root} ""
 
-    scanFile :: CanonicalPath -> IO (CanonicalPath, a)
+    filePath :: String -> FilePath
+    filePath = \case
+      quoted@('\"' : _) -> fromMaybe quoted $ readMaybe quoted
```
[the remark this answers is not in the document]
aeqz: Yeah, it was just to satisfy the type system in a weird situation.

I could change it for throwing an error instead, but I'm going to try with the `-z` option that you proposed and let's decide if we like the result.

src/Xrefcheck/Scan.hs
```diff
@@ -193,83 +192,101 @@ readDirectoryWith mode config scanner root = do
...
     getTrackedFiles = readCreateProcess
-      (shell "git ls-files"){cwd = Just $ unCanonicalPath root} ""
+      (shell "git ls-files -z"){cwd = Just root} ""
     getUntrackedFiles = readCreateProcess
-      (shell "git ls-files --others --exclude-standard"){cwd = Just $ unCanonicalPath root} ""
+      (shell "git ls-files -z --others --exclude-standard"){cwd = Just root} ""
 
-    scanFile :: CanonicalPath -> IO (CanonicalPath, a)
+    fileLines :: String -> [String]
+    fileLines = go DL.empty
```
aeqz, line 200: This is `lines` but with null line delimiters. Do we have already a function available to split a list in this way? We could also use the [split](https://hackage.haskell.org/package/split) package.

src/Xrefcheck/Verify.hs
aeqz: > Mmm, this seems like a justified case to start using OS-dependent bats tests 🤔 Like, set up some env variable in CI config, and rely on it in the test?

I would be worried about some points anyways:
- We should set the corresponding OS env variable for running bats tests in our device.
- Perhaps some Windows user or developer could have some trouble when cloning the repository if it contains such test files. I currently don't have a Windows device for trying it.

aeqz:
> Mmm, this seems like a justified case to start using OS-dependent bats tests 🤔 Like, set up some env variable in CI config, and rely on it in the test?

I would be worried about some points anyways:
- We should set the corresponding OS env variable for running bats tests in our device.
- Perhaps some Windows user or developer could have some trouble when cloning the repository if it contains such test files. I currently don't have a Windows device for trying it.

src/Xrefcheck/Verify.hs
```diff
@@ -157,7 +155,11 @@ instance Given ColorMode => Buildable VerifyError where
     LocalFileDoesNotExist file ->
       [int||
       File does not exist:
-        #{file}
+        #{file}#{build @Text
```
Martoon-00, line 158: I have found some trailing whitespaces here:
```suggestion:-0+0
        #{file}#{build @Text
```




</p>

src/Xrefcheck/Verify.hs
```diff
@@ -157,7 +155,11 @@ instance Given ColorMode => Buildable VerifyError where
     LocalFileDoesNotExist file ->
       [int||
       File does not exist:
-        #{file}
+        #{file}#{build @Text 
+          if hasBackslash file
```
Martoon-00, line 159: I have found some trailing whitespaces here:
```suggestion:-0+0
          if hasBackslash file
```




</p>

src/Xrefcheck/Verify.hs
```diff
@@ -157,7 +155,11 @@ instance Given ColorMode => Buildable VerifyError where
     LocalFileDoesNotExist file ->
       [int||
       File does not exist:
-        #{file}
+        #{file}#{build @Text 
+          if hasBackslash file 
+          then "\\n  Its reference contains a backslash. Maybe it uses the wrong path separator."
```
Martoon-00, line 160: I have found some trailing whitespaces here:
```suggestion:-0+0
          then "\\n  Its reference contains a backslash. Maybe it uses the wrong path separator."
```




</p>

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">WIP commits are still there.</td>
    </tr>
  </tbody>
</table>

</p>

src/Xrefcheck/Verify.hs
Martoon-00: > so maybe it can be seen as a point in where a task passes to be anonymous (such a task should exist in order to perform the action), and then it would be justified to require a witness there.

Hmm, actually were it me who is doing this task, and if I wanted to make it clear and neat, I guess I would think in the same manner. And that timestamp is semantically tightly coupled with fixable failures, only they can contribute to that timer.

What made me dislike the need in witness was primarily the fact that sometimes we pass dummy `""` as a witness to those functions. But given your reasoning, and given this `""` occurs only in tests, that seems justified.

So I got convinced. Fell free to revert that commit; probably it would also make sense to provide your reasoning in haddock, otherwise I guess someone else will just remove that witness argument in the future.

src/Xrefcheck/Scan.hs
Martoon-00: I also don't know about such a function, that's sad that it seems to be absent in base.

Yeah, `split` library is an option, but I think let's not add extra dependencies for just one use case. I believe via `Data.List.break` this can be implemented relatively short.

src/Xrefcheck/Verify.hs
```diff
@@ -157,7 +155,11 @@ instance Given ColorMode => Buildable VerifyError where
     LocalFileDoesNotExist file ->
       [int||
       File does not exist:
-        #{file}
+        #{file}#{build @Text
```
Martoon-00, line 158: Minor: can use `#l{ ... }` for textual literals, to avoid `build @Text`.

Martoon-00:
> > Mmm, this seems like a justified case to start using OS-dependent bats tests thinking Like, set up some env variable in CI config, and rely on it in the test?
> 
> I would be worried about some points anyways:
> 
>     * We should set the corresponding OS env variable for running bats tests in our device.
> 
>     * Perhaps some Windows user or developer could have some trouble when cloning the repository if it contains such test files. I currently don't have a Windows device for trying it.

That's true, I forgot that tests are run not only by CIs.

Hmm, if I remember correctly, we already [had tests](https://github.com/serokell/xrefcheck/blob/master/tests/golden/check-ignoreExternalRefsTo/check-ignoreExternalRefsTo.bats#L25) that imply different output under Linux and Windows.

That test is written as `assert_diff ... || assert_diff ...` which is a bit sad, but if in our case at Windows we expect quite custom "File does not exist", I think that should work. What would you say?

src/Xrefcheck/System.hs
```diff
@@ -51,125 +45,133 @@ askWithinCI = lookupEnv "CI" <&> \case
...
-      | FP.isPathSeparator c = FPP.pathSeparator
-      | otherwise = c
-
-    dropLeadingSepAndEmptyCase :: FilePath -> FilePath
-    dropLeadingSepAndEmptyCase path = case dropWhile FP.isPathSeparator path of
-      "" -> "."
-      other -> other
+-- It should be created from a 'RelPosixLink' via
+-- 'canonicalizeRelPosixLink'.
+newtype CanonicalRelPosixLink = UnsafeCanonicalRelPosixLink
```
Martoon-00, line 121: Btw on the naming, aren't "canonical" and "posix" mutually constradicting properties? One is about system-dependent path-separators, and the other is about posix path-separators; or it's more than that?

src/Xrefcheck/System.hs
Martoon-00: Thanks, this clarifies to me, sounds like a fair change :+1:

aeqz:
> That test is written as assert_diff ... || assert_diff ... which is a bit sad, but if in our case at Windows we expect quite custom "File does not exist", I think that should work. What would you say?

I tried to add my example and the Windows CI failed: the repo cannot be cloned if it contains the `a\a.md` file. Then I tried to create this file on the fly during the test and it cannot be created on Windows, so I made the test to pass if this file cannot be created.

src/Xrefcheck/System.hs
aeqz: Hmm currently a canonical link is a valid relative POSIX link (i.e. `CanonicalRelPosixLink` subset of `RelPosixLink`). 

Maybe the name is an implementation detail leak, is that what you mean? I can change it to just `CanonicalLink` if it sounds more appropriate.

Martoon-00: Oh sorry, I see I was saying nonsense, please revert that commit.

Martoon-00:
The last request: do we have tests on paths case-sensitivity? That `aBc.md` file would not be considered as referred by `abc.md` link?

I think that would be aa finishing touch for this refactoring.

aeqz:
> The last request: do we have tests on paths case-sensitivity?

We have tests for anchor case-sensitivity, but not for path case-sensitivity. I will add them 👍

tests/golden/check-case-sensitivity-path/check-case-sensitivity-path.bats
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bats
+
+# SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io>
```
Martoon-00, line 3: :warning: The year in this license header is outdated, time to update!





</p>

Martoon-00 (approved):
Great, thank you for this work!

Please squash the fixup commits and let's merge.

PR #264, opened 2023-01-10
[#25] Handle 304 redirects

### LINKED ISSUE #25: Report redirects as broken links

aeqz:
## Description

Problem: after recent work on Xrefcheck redirect behavior, it remained to discuss how to handle 304 redirects.

Solution: with the current default configuration, 304 redirects are considered as valid, which seems appropriate taking into account that 304 responses usually mean that you previously received a successful response for that same request and there is no need to retransmit the resource again. We add a test case for this default config and update the FAQ section.

It is also related to conditional  (e.g. `If-None-Match` and `If-Modified-Since`) and cache request headers. As Xrefcheck does not send them when performing HTTP requests, usually it should not receive 304 redirects.

## Related issue(s)

Fixes #25 

#### Related changes

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

#### Stylistic guide


tests/Test/Xrefcheck/RedirectDefaultSpec.hs
```diff
@@ -23,7 +23,7 @@ import Xrefcheck.Verify
 
 test_redirectRequests :: TestTree
 test_redirectRequests = testGroup "Redirect response defaults"
-  [ testGroup "Temporary" $ temporaryRedirectTests <$> [302, 303, 307]
+  [ testGroup "Temporary and 304" $ allowedRedirectTests <$> [302, 303, 304, 307]
```
Martoon-00, line 26: I would put it to a separate group. I think, separation into groups should rather be by semantics, not by the behaviour, and semantics of 304 is quite custom, its behaviour just happened to fall into the same group as temporary redirects.

Martoon-00:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">WIP commits are still there.</td>
    </tr>
  </tbody>
</table>

</p>

Martoon-00 (approved):
LGTM, thanks.

PR #265, opened 2023-01-12
[#235] Refactor progress bar interface

aeqz:
## Description

Problem: the current progress bar interface is quite low-level and error-prone.

Solution: the proposed solution here modifies both the progress bar internal model and interface by representing progress units by witnesses. It does not require great changes, but the resulting interface is clearer and less error-prone.

## Related issue(s)

Relates #235

#### Related changes

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

#### Stylistic guide

### DIFF
--- a/src/Xrefcheck/Progress.hs
+++ b/src/Xrefcheck/Progress.hs
@@ -9,16 +9,16 @@ module Xrefcheck.Progress
     TaskTimestamp (..)
 
     -- * Progress
-  , Progress (..)
+  , Progress
   , initProgress
-  , incProgress
-  , incProgressUnfixableErrors
-  , incProgressFixableErrors
-  , decProgressFixableErrors
-  , fixableToUnfixable
+  , reportSuccess
+  , reportError
+  , reportRetry
+  , getTaskTimestamp
   , setTaskTimestamp
   , removeTaskTimestamp
   , checkTaskTimestamp
+  , sameProgress
   , showProgress
 
     -- * Printing
@@ -31,6 +31,7 @@ import Universum
 
 import Data.Ratio ((%))
 import Data.Reflection (Given)
+import Data.Set qualified as S
 import Time (Second, Time, sec, unTime, (-:-))
 
 import Xrefcheck.Util
@@ -44,7 +45,7 @@ import Xrefcheck.Util
 data TaskTimestamp = TaskTimestamp
   { ttTimeToCompletion :: Time Second
     -- ^ The amount of time required for the task to be completed.
-  , ttStart            :: Time Second
+  , ttStart :: Time Second
     -- ^ The timestamp of when the task had started, represented by the number of seconds
     -- since the Unix epoch.
   } deriving stock (Show)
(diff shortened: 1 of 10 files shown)


src/Xrefcheck/Util.hs
```diff
@@ -65,16 +63,6 @@ posixTimeToTimeSecond posixTime =
...
-composeFuncList :: [a -> a] -> a -> a
-composeFuncList = foldr (.) id
-
-whenFunc :: Bool -> (a -> a) -> (a -> a)
-whenFunc True f = f
-whenFunc False _ = id
-
-unlessFunc :: Bool -> (a -> a) -> (a -> a)
-unlessFunc = whenFunc . not
-
```
aeqz, line 77: I think that @YuriRomanowski added these helper functions recently, which look useful, but I would remove them if this PR proposal is accepted because they would become unused.

I can undo this change if anyone prefers keeping them here.

YuriRomanowski: Yep, I think this is OK

YuriRomanowski (approved):

YuriRomanowski:
I like how you changed the Progress fields and I like your trick with `Set` :) I think this is cool, approved.

src/Xrefcheck/Progress.hs
```diff
@@ -53,72 +54,73 @@ data TaskTimestamp = TaskTimestamp
...
-    -- ^ How much of the completed work finished with an unfixable error.
-  , pErrorsFixable   :: !a
-    -- ^ How much of the completed work finished with an error that can be
-    -- eliminated upon further verification.
-  , pTaskTimestamp   :: Maybe TaskTimestamp
+  , pSuccess :: !a
+    -- ^ How much has been completed with success.
+  , pError :: !a
+    -- ^ How much has been completed with error.
+  , pRetrying :: !(S.Set w)
```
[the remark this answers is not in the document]
aeqz: Yeah, the purpose of having a set is for avoiding to externally keep track of links being retried, which can be observed in the `Verify` module usage of this interface. Previously, it was required to use different functions depending on whether the resource was being retried or not (e.g. when success, `decProgressFixableErrors` vs `incProgress`). Now, in this case, it is only required to use `reportSuccess` with the corresponding 'witness'.

This is an example of how the progress state evolves by having this `Set`:

```
initProgress 2
-- 2 Pending
reportRetry "a"
-- 1 Pending, 1 Retrying
reportSuccess "b"
-- 1 Ok, 1 Retrying
reportError "a"
-- 1 Ok, 1 Error
```

If we had a number of retrying units instead:

```
initProgress 2
-- 2 Pending
reportRetry "a"
-- 1 Pending, 1 Retrying
reportSuccess "b"
-- At this point, how do we know if the state should be
--     1 Ok, 1 Retrying (i.e. increment ok)
-- or
--     1 Pending, 1 Ok (i.e. decrement retrying and increment ok)
-- ?
```

Martoon-00 (approved):
Looks good.

I like how you use witnesses after all, this really fits the best.

And most of the refactorings make sense for me (I'm not actually sure about changes in `modifyProgressRef` though, but since @YuriRomanowski approved this PR, let's suggest those changes are for the better).

PR #269, opened 2023-01-23
[#222] Prevent progress bar rendering from abort

### LINKED ISSUE #222: Make sure that even exception does not abort progress bar rendering on the middle

In #163 we have fixed the progress bar rendering thread to never abort in the middle of printing disregard when the links verification has finished. Now the terminal is not supposed to periodically appear in some weird state after `xrefcheck` completes.

However in case of an async exception (it may be a `UserInterrupt` which we now handle), AFAIS we still let the progress bar to be rendered partially, and this seems to be as bad as having a not cleaned up resource.

# Acceptance criteria

* Even in case of exceptions we never render the terminal partially.
* Double `^C` should still abort the application immediately.

aeqz:
## Description

Problem: It seems that the current progress bar rendering action is not completely protected against exceptions.

Solution: As it is an action that runs for a short amount of time and we really want it to not be interrputed, we can wrap it with `uninterruptibleMask`. The resulting rendering loop also gets simpler with this change.

## Related issue(s)

Fixes #222 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

### DIFF
--- a/src/Xrefcheck/Verify.hs
+++ b/src/Xrefcheck/Verify.hs
@@ -27,9 +27,9 @@ module Xrefcheck.Verify
 
 import Universum
 
-import Control.Concurrent.Async (Async, async, cancel, poll, wait, withAsync)
+import Control.Concurrent.Async (Async, cancel, poll, wait, withAsync)
 import Control.Exception (AsyncException (..), throwIO)
-import Control.Exception.Safe (handleAsync)
+import Control.Exception.Safe (handleAsync, uninterruptibleMask_)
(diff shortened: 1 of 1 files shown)


Martoon-00 (approved):
That's what I meant, thank you :+1: 

However let's also wait for @dcastro's reply on [this message](https://github.com/serokell/xrefcheck/pull/163#issuecomment-1321021135), he is now back from vacation and I asked him to take a look when he has time.

PR #267, opened 2023-01-16
Replace buildkite ci with github-actions

PhilTaken:
## Related issue(s)

Fixes #

## :white_check_mark: Checklist for your Pull Request

Ideally a PR has all of the checkmarks set.

If something in this list is irrelevant to your PR, you should still set this
checkmark indicating that you are sure it is dealt with (be that by irrelevance).

#### Related changes (conditional)

- Tests
  - [ ] If I added new functionality, I added tests covering it.
  - [ ] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [ ] I checked whether I should update the docs and did so if necessary:
    - [README](https://github.com/serokell/xrefcheck/tree/master/README.md)
    - Haddock

- Public contracts
  - [ ] Any modifications of public contracts comply with the [Evolution
  of Public Contracts](https://www.notion.so/serokell/Evolution-of-Public-Contracts-2a3bf7971abe4806a24f63c84e7076c5) policy.
  - [ ] I added an entry to the [changelog](https://github.com/serokell/xrefcheck/tree/master/CHANGES.md) if my changes are visible to the users
        and
  - [ ] provided a migration guide for breaking changes if possible

#### Stylistic guide (mandatory)

- [ ] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [ ] My code complies with the [style guide](https://github.com/serokell/style/blob/master/haskell.md).

#### ✓ Release Checklist

- [ ] I updated the version number in `package.yaml`.
- [ ] I updated the [changelog](https://github.com/serokell/xrefcheck/tree/master/CHANGES.md) and moved everything
      under the "Unreleased" section to a new section for this release version.
- [ ] (After merging) I edited the [auto-release](https://github.com/serokell/xrefcheck/releases/tag/auto-release).
    * Change the tag and title using the format `vX.Y.Z`.
    * Write a summary of all user-facing changes.
    * Deselect the "This is a pre-release" checkbox at the bottom.
- [ ] (After merging) I updated [`xrefcheck-action`](https://github.com/serokell/xrefcheck-action#updating-supported-versions).
- [ ] (After merging) I uploaded the package to [hackage](https://hackage.haskell.org/package/xrefcheck).

PR #284, opened 2023-06-10
Compatibility with `optparse-applicative-0.18`

andreasabel:
`optparse-applicative-0.18` switched to `prettyprinter`, so some reexports are missing now.
- https://github.com/pcapriotti/optparse-applicative/issues/481

```
Building library for xrefcheck-0.2.2..
[17 of 19] Compiling Xrefcheck.CLI    ( src/Xrefcheck/CLI.hs, /Users/abel/bin/src/xrefcheck/dist-newstyle/build/x86_64-osx/ghc-9.2.8/xrefcheck-0.2.2/build/Xrefcheck/CLI.o, /Users/abel/bin/src/xrefcheck/dist-newstyle/build/x86_64-osx/ghc-9.2.8/xrefcheck-0.2.2/build/Xrefcheck/CLI.dyn_o )

src/Xrefcheck/CLI.hs:34:46: error:
    Module ‘Options.Applicative.Help.Pretty’ does not export ‘displayS’
   |
34 | import Options.Applicative.Help.Pretty (Doc, displayS, fill, fillSep, indent, renderPretty, text)
   |                                              ^^^^^^^^

src/Xrefcheck/CLI.hs:34:79: error:
    Module
    ‘Options.Applicative.Help.Pretty’
    does not export
    ‘renderPretty’
   |
34 | import Options.Applicative.Help.Pretty (Doc, displayS, fill, fillSep, indent, renderPretty, text)
   |                                                                               ^^^^^^^^^^^^

src/Xrefcheck/CLI.hs:34:93: error:
    Module ‘Options.Applicative.Help.Pretty’ does not export ‘text’
   |
34 | import Options.Applicative.Help.Pretty (Doc, displayS, fill, fillSep, indent, renderPretty, text)
   |                                                                                             ^^^^
Error: cabal: Failed to build xrefcheck-0.2.2
```

This PR restores compatibility with the latest `optparse-applicative`.

I simplified the code to display the help text, but the rendering still looks fine:
```
To ignore a link in your markdown, include "<!-- xrefcheck: ignore <mode> -->"
comment with one of these modes:
  "link"                   Ignore the link right after the comment.
  "paragraph"              Ignore the whole paragraph after the comment.
  "file"                   This mode can only be used at the top of markdown or
                           right after comments at the top.
```
So maybe the code was unnecessary complicated to begin with.

On hackage I added upper bound `optparse-applicative < 0.18`, e.g. https://hackage.haskell.org/package/xrefcheck-0.2.2/revisions/.

### DIFF
--- a/src/Xrefcheck/Verify.hs
+++ b/src/Xrefcheck/Verify.hs
@@ -768,7 +768,7 @@ checkExternalResource followed config@Config{..} link
           other -> throwError $ ExternalResourceSomeError $ show other
       where
         retryAfterInfo :: Response a -> Maybe RetryAfter
-        retryAfterInfo = readMaybe . decodeUtf8 <=< L.lookup hRetryAfter . responseHeaders
+        retryAfterInfo = readMaybe . decodeUtf8 @Text <=< L.lookup hRetryAfter . responseHeaders
 
     checkFtp :: URI -> Bool -> ExceptT VerifyError IO ()
     checkFtp uri secure = do
--- a/src/Xrefcheck/CLI.hs
+++ b/src/Xrefcheck/CLI.hs
@@ -31,7 +31,7 @@ import Options.Applicative
   (Mod, OptionFields, Parser, ReadM, auto, command, eitherReader, execParser, flag, flag',
   footerDoc, fullDesc, help, helpDoc, helper, hsubparser, info, infoOption, long, metavar, option,
   progDesc, short, strOption, switch, value)
-import Options.Applicative.Help.Pretty (Doc, displayS, fill, fillSep, indent, renderPretty, text)
+import Options.Applicative.Help.Pretty (Doc, fill, fillSep, indent, pretty)
 import Options.Applicative.Help.Pretty qualified as Pretty
 import Text.Interpolation.Nyan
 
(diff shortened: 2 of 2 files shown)


gromakovsky:
@andreasabel thank you for this PR. We are currently switching from Buildkite to GitHub Actions in #267. I hope we'll finish soon and then CI might start working in this PR after rebasing on the latest `master` branch.

andreasabel:
Thanks for the heads up, @gromakovsky !  I leave further handling of this PR to you, please use/modify it in any way you want.

src/Xrefcheck/Verify.hs
```diff
@@ -768,7 +768,7 @@ checkExternalResource followed config@Config{..} link
           other -> throwError $ ExternalResourceSomeError $ show other
       where
         retryAfterInfo :: Response a -> Maybe RetryAfter
-        retryAfterInfo = readMaybe . decodeUtf8 <=< L.lookup hRetryAfter . responseHeaders
+        retryAfterInfo = readMaybe . decodeUtf8 @Text <=< L.lookup hRetryAfter . responseHeaders
```
aeqz, line 771: In this case, `readMaybe` comes from `Text.Read` and takes a `String` parameter, so it does not compile for me unless this type parameter application is removed.

andreasabel: I was surprised that I had to add it, because it is orthogonal to the concern of this PR, namely `optparse-applicative`. 
Let me investigate some more...

andreasabel: So indeed `stack build` fails now (which uses LTS 19 which is GHC 9.0), but `cabal build -w ghc-9.2.8` succeeds and needs this change.  
I am gonna remove this patch, it is something to be sorted out when migrating from GHC 9.0 to 9.2 and up.
(Note that GHC 9.2.8 is the currently recommended version of GHC.)

aeqz: Yes, it should be updated at some point. I will create an issue referencing this comments thread. Thank you!

andreasabel: NB: `cabal build -w ghc-9.0.2` fails because of:
- https://github.com/vincenthz/hs-connection/issues/54

aeqz:
> but it does not happen when built with GHC 9.2.8 on cabal

So all the points that I mentioned seem to be related to the GHC version difference. We will go back to this PR when upgrading the stack setup, probably after finishing the ongoing CI work!

PR #301, opened 2024-09-11
[Chore] Update development setup

int-index:
* flake.nix: inherit devShells, so that `nix develop` works
* tests: configurable mock server port to avoid conflicts

With these changes I've been able to build the project and run its tests as follows

```
nix shell nixpkgs#haskellPackages.hpack nixpkgs#cabal-install
nix develop -c $SHELL
hpack
vsftpd \
   -orun_as_launching_user=yes \
   -olisten_port=2221 \
   -olisten=yes \
   -oftp_username=$(whoami) \
   -oanon_root=./ftp-tests/ftp_root \
   -opasv_min_port=2222 \
   -ohide_file='{.*}' \
   -odeny_file='{.*}' \
   -oseccomp_sandbox=no \
   -olog_ftp_protocol=yes \
   -oxferlog_enable=yes \
   -ovsftpd_log_file=./ftp.log &
cabal test ftp-tests --test-options="--ftp-host ftp://127.0.0.1:2221"
cabal test xrefcheck-tests --test-options="--mock-server-port 3001"
```

### DIFF
--- a/tests/Test/Xrefcheck/ConfigSpec.hs
+++ b/tests/Test/Xrefcheck/ConfigSpec.hs
@@ -13,7 +13,7 @@ import Control.Exception qualified as E
 import Data.List (isInfixOf)
 import Data.Yaml (ParseException (..), decodeEither')
 import Network.HTTP.Types (Status (..))
-import Test.Tasty (TestTree, testGroup)
+import Test.Tasty (TestTree, askOption, testGroup)
 import Test.Tasty.HUnit (assertFailure, testCase, (@?=))
 import Test.Tasty.QuickCheck (ioProperty, testProperty)
 
@@ -22,7 +22,7 @@ import Xrefcheck.Core (Flavor (GitHub), allFlavors)
 import Xrefcheck.Scan (ecIgnoreExternalRefsToL)
 import Xrefcheck.Verify (VerifyError (..), checkExternalResource)
 
-import Test.Xrefcheck.Util (mockServer)
+import Test.Xrefcheck.Util (mockServer, mockServerUrl)
 
 test_config :: [TestTree]
 test_config =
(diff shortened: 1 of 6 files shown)


tests/Test/Xrefcheck/ConfigSpec.hs
```diff
@@ -80,7 +81,8 @@ test_config =
   ]
 
   where
-    checkLinkWithServer config link expectation =
-      E.bracket (forkIO mockServer) killThread $ \_ -> do
-        result <- runExceptT $ checkExternalResource emptyChain config link
+    checkLinkWithServer mockServerPort config link expectation =
+      E.bracket (forkIO (mockServer mockServerPort)) killThread $ \_ -> do
+        let url = (mockServerUrl mockServerPort link)
```
gromakovsky, line 86: Nitpick: brackets look redundant here.

tests/Test/Xrefcheck/Util.hs
```diff
@@ -18,7 +21,30 @@ parse :: Flavor -> ScanAction
...
+newtype MockServerPort = MockServerPort Int
+  deriving stock (Show, Eq)
+
+instance IsOption MockServerPort where
+  defaultValue = MockServerPort 3000
+  optionName = "mock-server-port"
+  optionHelp = "[Test.Xrefcheck.Util] Mock server port"
+  parseValue v = MockServerPort <$> safeRead v
+  optionCLParser = MockServerPort <$> option auto
+    (  long (untag (optionName :: Tagged MockServerPort String))
```
gromakovsky, line 48: Can't you just use `optionName @MockServerPort` here instead of `Tagged`?

int-index: I bet I can, but I followed the pattern established in `ftp-tests/Test/Xrefcheck/FtpLinks.hs`.

Will remove `Tagged` in both places then.

tests/Test/Xrefcheck/ConfigSpec.hs
int-index: Fixed

tests/Test/Xrefcheck/Util.hs
int-index: Fixed

int-index:
@gromakovsky How is it looking now?

tests/Test/Xrefcheck/Util.hs
gromakovsky: I asked because I noticed a new dependency being added which seems unnecessary (I mean `tagged`). Turns out it's not really new, it's new for that test suite, but not for the whole package. However, I don't quite understand the new code:
```
  optionCLParser = FtpHostOpt <$> strOption
    (  long (untag @FtpHostOpt optionName)
    <> help (untag @FtpHostOpt optionHelp)
    )
```

Can't it be just
```
  optionCLParser = FtpHostOpt <$> strOption
    (  long optionName
    <> help optionHelp
    )
```
?
We have 
> long :: [HasName](https://hackage.haskell.org/package/optparse-applicative-0.18.1.0/docs/Options-Applicative-Builder.html#t:HasName) f => [String](https://hackage.haskell.org/package/base-4.16.3.0/docs/Data-String.html#t:String) -> [Mod](https://hackage.haskell.org/package/optparse-applicative-0.18.1.0/docs/Options-Applicative-Builder.html#t:Mod) f a

and 

> strOption :: [IsString](https://hackage.haskell.org/package/base-4.16.3.0/docs/Data-String.html#t:IsString) s => [Mod](https://hackage.haskell.org/package/optparse-applicative-0.18.1.0/docs/Options-Applicative-Builder.html#t:Mod) [OptionFields](https://hackage.haskell.org/package/optparse-applicative-0.18.1.0/docs/Options-Applicative-Builder.html#t:OptionFields) s -> [Parser](https://hackage.haskell.org/package/optparse-applicative-0.18.1.0/docs/Options-Applicative-Types.html#t:Parser) s

and we want the result of `strOption (…)` to be `Parser Text`. I think my short version should just work, because the argument of the `FtpHostOpt` constructor is `Text`, but if it doesn't, I think it can be solved by explicit type applications for `long` or `help`.

int-index: The problem is that `optionName :: Tagged v String` and `optionHelp :: Tagged v String` and we need to unwrap them to get the strings. See the definition of [IsOption](https://hackage.haskell.org/package/tasty-1.5/docs/Test-Tasty-Options.html#t:IsOption) in `Test.Tasty.Options`:

```
class Typeable v => IsOption v where
  defaultValue :: v
  parseValue :: String -> Maybe v
  optionName :: Tagged v String
  optionHelp :: Tagged v String
  showDefaultValue :: v -> Maybe String
  optionCLParser :: Parser v
```

So we can't fully get rid of `Tagged` there. It's used to avoid ambiguous type variables.

Regarding a new dependency, it's not entirely new: I had to make it explicit to use `untag`, but we already depended on `tagged` transitively via `tasty`.

gromakovsky: Oh, ok, I see now. For some reason I thought that `optionName` was a local variable, not a type class member. Okay, then it all makes sense to me.

gromakovsky (approved):

PR #307, opened 2024-11-11
[#226] Change error message formatting

### LINKED ISSUE #226: Revise styling of the output

Look at the [readme](https://github.com/serokell/xrefcheck#readme), at how the output looks like.

This is all subjective, but there are things that feel like a bad design of the style:
* Names of bad files are given in gray, and they should on contrary stand out.
* Those `-` before `text`, `link`, and `anchor` look redundant, especially after #202.

Let's discuss all this and apply some fixes before the release.

# Acceptance criteria
* Some feedback on the styling is gathered.
* Concrens are discussed.
* On concerns we agreed upon - adjustements in the code are made.

int-index:
## Description

**Problem:** multiple issues with the CLI output

1. In the `--help` output, metavars `GLOB_PATTERN` and `REPOSITORY_TYPE` used spaces instead of underscores. Variables typically don't contain spaces in their names, so it looked confusing.
2. Incorrect or unintuitive coloring of the output. For example, filenames were dim even though they're important and should stand out, the report of success wasn't green, error messages weren't red, etc.
3. Excessive indentation and empty lines that made it difficult to visually parse the error report. Also the use of the exotic `➥` character that looks bad in many font configurations.
4. Filenames and line numbers were far apart, making it impossible to Ctrl+Click to jump to the source of the error.
5. In case of connection failure, the output was too verbose and platform-dependent.

**Solution:**
1. Include the filename in positions returned by the scanner.
2. Adjust the pretty-printing functions and update test output.
3. Extend `VerifyError` with a constructor for connection failures.

## Related issue(s)

Fixes #226.

- Tests
        silently reappearing again.

- Documentation
    - Haddock

- Public contracts
        and

#### ✓ Release Checklist


Martoon-00:
Looks good, I see why these changes are reasonable and agree.

One concern: after merging error message for every single reference into one block, discrepancies in indentation became more apparent:

* The difference in leading `-` being present at one line and being absent at another. Maybe we can remove this `-` in lists after all. (Clearly the last line is not a list, but the difference still stands out)

```
check-images.md:20:1-34: bad reference:
  The reference to "bad image ref 3" failed verification.
  - link (relative): ./3.png      <------
  File does not exist:
    ./3.png                               <------
```

* Indentation size here:

```
The reference to "Reference lowered" failed verification.
  - anchor (file-local): uppercase-name
  Anchor 'uppercase-name' is not present, did you mean:
      - UPPERCASE-NAME (handmade) 14:3-27
```

Maybe these are worth changing to look uniform. WDYT?

int-index:
Agreed, I'll try to change the phrasing further to avoid this issue.

int-index:
I've pushed a change, @Martoon-00 would you be so kind to take another look?

This time indentation should be consistent. Also I've modified the scanner so that all positions include the filename.

Martoon-00:
My apologies for coming late.

Looks much better as to me, thanks for addressing my concern.

Martoon-00:
I have another minor issue:

```
check-ignoreExternalRefsTo.md:7:10-53: bad reference:
  The reference to "web-site" failed verification.
  Connection failure
  when processing an external link: https://localhost:20000/web-site
```

Commonly used links are longer, and likely the last line won't fit into the default terminal size of 80 characters width. If most of the links get wrapped, that may look ugly. For this reason, previously I tried to keep links in their own line; would it make sense here?

Ofc some links will wrap in either case, but, arguably, they'd still look prettier when start from their own line.

int-index:
> Commonly used links are longer, and likely the last line won't fit into the default terminal size of 80 characters width. If most of the links get wrapped, that may look ugly. For this reason, previously I tried to keep links in their own line; would it make sense here?

Good point, I moved the link to its own line.

PR #271, opened 2023-01-26
[#270] Handle relative redirects

aeqz:
## Description

Problem: Currently, Xrefcheck can follow redirects with an absolute location link, but it cannot handle relative ones.

Solution: After parsing the location link, obtain the corresponding absolute link by using the original request one.

## Related issue(s)

Fixes #270 

#### Related changes (conditional)

- Tests
  - [x] If I added new functionality, I added tests covering it.
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [x] I checked whether I should update the docs and did so if necessary:
    - [README](https://github.com/serokell/xrefcheck/tree/master/README.md)
    - Haddock

- Public contracts
  - [x] Any modifications of public contracts comply with the [Evolution
  of Public Contracts](https://www.notion.so/serokell/Evolution-of-Public-Contracts-2a3bf7971abe4806a24f63c84e7076c5) policy.
  - [x] I added an entry to the [changelog](https://github.com/serokell/xrefcheck/tree/master/CHANGES.md) if my changes are visible to the users
        and
  - [x] provided a migration guide for breaking changes if possible

#### Stylistic guide (mandatory)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [x] My code complies with the [style guide](https://github.com/serokell/style/blob/master/haskell.md).

PR #268, opened 2023-01-19
[#237] Modify coloring options

aeqz:
## Description

Problem: We noticed that output was not being colorized in GitLab CI due to the current implemented guesses for whether showing colors or not.

Solution: On the one hand, we extend the current guesses to also enable coloring by default when the CI env var is set to true. On the other hand, we also add a new flag, color, which avoids these guesses and enables colors.

I have tried in a GitLab CI project and the output is colored now.

## Related issue(s)

Fixes #237 

#### Related changes (conditional)

- Tests
  - [x] If I added new functionality, I added tests covering it.
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [x] I checked whether I should update the docs and did so if necessary:
    - [README](https://github.com/serokell/xrefcheck/tree/master/README.md)
    - Haddock

- Public contracts
  - [x] Any modifications of public contracts comply with the [Evolution
  of Public Contracts](https://www.notion.so/serokell/Evolution-of-Public-Contracts-2a3bf7971abe4806a24f63c84e7076c5) policy.
  - [x] I added an entry to the [changelog](https://github.com/serokell/xrefcheck/tree/master/CHANGES.md) if my changes are visible to the users
        and
  - [x] provided a migration guide for breaking changes if possible

#### Stylistic guide (mandatory)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [x] My code complies with the [style guide](https://github.com/serokell/style/blob/master/haskell.md).

PR #261, opened 2022-12-30
[#242] No scan symlinks as md files

aeqz:
Problem: when the repository contains a symlink to a markdown file, it is processed by xrefcheck as if it was the same markdown file but in the symlink's location. This leads to broken references and can be avoided because neither GitHub nor GitLab try to render symlinks as the file they point to.

Solution: consider symlinks as no scannable files. In the future, we will consider to include a new dedicated scanner for symlinks if it works.

## Related issue(s)

Fixes #242
Relates #244

#### Related changes (conditional)

- Tests
  - [x] If I added new functionality, I added tests covering it.
  - [x] If I fixed a bug, I added a regression test to prevent the bug from
        silently reappearing again.

- Documentation
  - [x] I checked whether I should update the docs and did so if necessary:
    - [README](https://github.com/serokell/xrefcheck/tree/master/README.md)
    - Haddock

- Public contracts
  - [x] Any modifications of public contracts comply with the [Evolution
  of Public Contracts](https://www.notion.so/serokell/Evolution-of-Public-Contracts-2a3bf7971abe4806a24f63c84e7076c5) policy.
  - [x] I added an entry to the [changelog](https://github.com/serokell/xrefcheck/tree/master/CHANGES.md) if my changes are visible to the users
        and
  - [x] provided a migration guide for breaking changes if possible

#### Stylistic guide (mandatory)

- [x] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [x] My code complies with the [style guide](https://github.com/serokell/style/blob/master/haskell.md).

Issue #254, opened 2022-12-21
Revise behaviour of `dump-config` command


Martoon-00:
# Clarification and motivation

Currently this command immediately writes to a file, while an intuition would say that it should just print config to the console. (and actually, it was me who implemented it this way :sweat:). But okay, this has some advantages.

But this command can also overwrite the previously created config. I.e. I could create config and tune it up, then accidentally call `xrefcheck dump-config` again (e.g. by getting this command from history) and all my changes will be lost. This is too bad.

# Acceptance criteria

* `dump-config` does not overwrite the existing file, unless this is told explicitly (with some `--force` flag).
* (optional) How this command outputs the config is revised.
* (optional) It is possible to print the config to stdout.

PR #310, opened 2024-12-27
v0.3.0 release

int-index:
## Description

* #176
  + Enabled `autolink` extension for `cmark-gfm`, so now we're finding strings like `www.google.com` or `https://google.com`, treating them as links and checking.
* #175
  + Reorganize top-level config keys.
* #178
  + Rename exclusion-related config options.
* #183
  + Add support for image links.
* #199
  + Now annotation `<!-- xrefcheck: ignore all -->` instead of `<!-- xrefcheck: ignore file -->` should be used to disable checking for links in file, so it's clearer that file itself is not ignored (and links can target it).
* #215
  + Now we notify user when there are scannable files that were not added to Git yet. Also added CLI option `--include-untracked` to scan such files and treat as existing.
* #191
  + Now we consider slash `/` (and only it) as path separator in local links for all OS, so xrefcheck's report is OS-independent
  + Use utf-8 compatible codepage on Windows
* #224
  + Now the program output does not contain unicode characters that are not widely supported.
* #229
  + Now we call references to anchors in current file (e.g. `[a](#b)`) as `file-local` references instead of calling them `current file` (which was ambiguous).
* #233
  + Now xrefxcheck does not follow redirect links by default. It fails for permanent
    redirect responses (i.e. 301 and 308) and passes for temporary ones (i.e. 302, 303, 307).
* #231
  + Anchor analysis takes now into account the appropriate case-sensitivity depending on
    the configured Markdown flavour.
* #254
  + Now the `dump-config` command does not overwrite a file unless explicitly told with a `--force` flag. Also, a `--stdout` flag allows to print the config to stdout instead.
* #250
  + Now the redirect behavior for external references can be modified via rules in the configuration file with the `externalRefRedirects` parameter.
* #261
  + Symlinks are now not processed by the scanner.
* #268
  + Added CLI option `--color` that enables ANSI colors in output.
  + Changed the output coloring defaults to show colors when `CI` env variable is `true`.
* #271
  + Now Xrefcheck is able to follow relative redirects.
* #262
  + Now Xrefcheck includes a scanner that verifies the repository symlinks.
* #307
  + The output of Xrefcheck is now more legible.
  + Add the `--print-unix-paths` (`-u`) flag to print paths in Unix style (with forward slashes) on all platforms.

## ✓ Release Checklist

- [x] I updated the version number in `package.yaml`.
- [x] I updated the [changelog](https://github.com/serokell/xrefcheck/tree/master/CHANGES.md) and moved everything

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,9 @@
 Unreleased
 ==========
 
+0.3.0
+==========
+
 * [#176](https://github.com/serokell/xrefcheck/pull/176)
   + Enabled `autolink` extension for `cmark-gfm`, so now we're finding strings
     like `www.google.com` or `https://google.com`, treating them as links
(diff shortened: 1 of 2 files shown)


Martoon-00 (approved):
:+1:

